PR #78, opened 2019-06-06
Redirects from coops to paybc and back.   bcgov/entity#600

saravankumarpa:
*Issue #:*
https://github.com/bcgov/entity/issues/600

*Description of changes:*
- Handled redirection from coops [any client who is using the pay system]  to paybc and vice versa


pwei1018:
Can we make the header config as a global variable?

pksumesh:
Some labels are in the vue file instead of locale file. Looks good apart from that.

saravankumarpa:
> Can we make the header config as a global variable?

Added a new interceptor.

PR #92, opened 2019-07-11
Changes to remove Process.env and deploy app to TEST

saravankumarpa:
https://github.com/bcgov/entity/issues/720

Below changes are done

- Removed process.env

- Read openshift configs and saved to sessionstorage for later use

- Axios urls are relative.Added sample Caddy configs

- sessionstorage-jest-mock is used to mock things


auth-web/openshift/templates/auth-web-deploy.test.json
```diff
@@ -0,0 +1,150 @@
...
+                                        "name": "web-ui-configuration",
+                                        "readOnly": true,
+                                        "mountPath": "/var/www/html/config"
+                                    }
+                                ],
+                                "envFrom": [],
+                                "env": [
+                                    {
+                                        "name": "AUTH_API_URL",
+                                        "value": "https://auth-api-dev.pathfinder.gov.bc.ca/api/v1"
```
pwei1018: Should this value be the test api url?

auth-web/openshift/templates/auth-web-deploy.test.json
```diff
@@ -0,0 +1,150 @@
...
+                                    },
+                                    {
+                                        "name": "PAY_API_URL",
+                                        "value": "https://pay-api-dev.pathfinder.gov.bc.ca/api/v1"
```
pwei1018: Should this value be the test api url?

pwei1018:
I left comments in auth-web/openshift/templates/auth-web-deploy.test.json. Please check it out.

auth-web/openshift/templates/auth-web-deploy.test.json
saravankumarpa: Changed

auth-web/devtotestpipeline
```diff
@@ -0,0 +1,11 @@
+kind: "BuildConfig"
+apiVersion: "v1"
+metadata:
+  name: "deploy-ui-to-test-pipeline"
+spec:
+  source:
+    git:
+      uri: "https://github.com/saravankumarpa/sbc-auth"
```
sutherlanda: Is this file just for your own testing? Do we want to commit it?

auth-web/src/util/config-helper.ts
```diff
@@ -0,0 +1,38 @@
...
+        throw err
+      }
+      )
+  },
+
+  /**
+ * this will run everytime when vue is being loaded..so do the call only when session storage doesnt have the values
+ */
+  saveConfigToSessionStorage () {
+    console.log('sessionStorage.getItem(\'API_CONFIG\')'+sessionStorage.getItem('API_CONFIG'))
```
sutherlanda: Do we want to leave `console.log` statements in? Maybe we could come up with a convention for including these but having something in place for disabling them.  Otherwise it will be quite a bit of work to go back and remove them all once we move into production.  Maybe something like `DEBUG && console.log`? Not sure, and we should probably discuss as a team before deciding anything.

auth-web/vue.config.js
```diff
@@ -3,6 +3,19 @@ module.exports = {
...
   devServer: {
-    proxy: 'https://auth-api-dev.pathfinder.gov.bc.ca/'
+    proxy: {
+      '/auth/api/*': {
+        target: 'https://auth-api-dev.pathfinder.gov.bc.ca/api/v1', // if your local server is running , use that here
+        pathRewrite: {
+          '/auth/api/': ''
+        }
+      },
+      '/pay/api/*': { // TODO I havent tested this..but hopefully it will work..
```
sutherlanda: Is the TODO still valid? :)

auth-web/devtotestpipeline
saravankumarpa: Removed it. It was a helper file.Need not to be part of code base. May be we can create a gist for it.

auth-web/src/util/config-helper.ts
saravankumarpa: Removed all console logs.Also made eslint strict about console.log.No more console logs :)

auth-web/vue.config.js
saravankumarpa: kind of :-) 
This is only for local development and i havent tested payment in local. Hopefully it will work or first person to work in payment on local will have to figure it out...

PR #107, opened 2019-07-29
Business Contact UI (#904) [WIP]

sutherlanda:
* Updated Vuex store to longer store passcode and store business object instead of business number
* Implementation of business contact UI and input validation

*Issue #:* 904

*Description of changes:* UI for Business Contact entry

*API Code Review Checklist:*
 - #### Common
 - #### REST 
 - #### Code structure
 - #### Test Cases
 - #### Code Quality
    - [ ] Linting
- #### DB Migrations

*UI Code Review Checklist:*


auth-web/src/components/auth/BusinessContactForm.vue
```diff
@@ -0,0 +1,163 @@
...
+    </div>
+    <div class="business-contact-form_row">
+      <v-layout wrap>
+        <v-flex xs6 class="mr-5">
+          <v-text-field
+            box
+            label="Phone e.g. (555)-555-5555"
+            persistent-hint
+            type="tel"
+            v-mask="['(###)-###-####']"
```
saravankumarpa: do you wanna consider moving this to a common place? a constant class for masks..low priority

auth-web/src/components/auth/BusinessContactForm.vue
```diff
@@ -0,0 +1,163 @@
...
+    if (this.isFormValid()) {
+      // TODO: Change this to transition to entity dashboard once complete
+      setTimeout(() => {
+        window.location.href = this.VUE_APP_COPS_REDIRECT_URL
+      }, 500)
+    }
+  }
+
+  skip () {
+    // Go directly to co-op UI without saving
```
pksumesh: May be we need to set a session storage here, to handle the below scneario;
1) User logs in
2) clicks 'Skip'
3) Goes to co-ops, then accidentally by back button/or any action lands to login page.
4) I believe at this point user needs to be sent to Co-Ops again without asking the credentials and not showing the Business Contact Info UI - Since the user opted to not to fill the business contact info for that session ?

auth-web/src/components/auth/BusinessContactForm.vue
mengdong19: Phone number is not necessary to be this format if they are out of Canada.

auth-web/src/components/auth/PasscodeForm.vue
```diff
@@ -81,79 +81,79 @@
...
-          })
-      }
+          )
+          iframeServices.emit(this.getIFrameContent(), msg)
+          sessionStorage.KEYCLOAK_TOKEN = response.data.access_token
+          sessionStorage.KEYCLOAK_REFRESH_TOKEN = response.data.refresh_token
+          sessionStorage.REGISTRIES_TRACE_ID = response.data['registries-trace-id']
+
+          // attempt to load business
+          this.businessStore.loadBusiness(this.businessNumber)
```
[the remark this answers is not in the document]
sutherlanda: Yeah, for MVP, we'll create a business record on log in (via passcode) if we don't already have a record.  We'll always redirect to contact UI if there is no contact info present.

Post MVP will likely be a different since we won't be logging in with passcode at that point

auth-web/src/components/auth/BusinessContactForm.vue
```diff
@@ -0,0 +1,163 @@
...
+  private emailAddress = ''
+  private confirmedEmailAddress = ''
+  private phoneNumber = ''
+  private extension = ''
+  private formError = ''
+
+  private emailRules = [
+    v => !!v || 'Email address is required',
+    v => {
+      const pattern = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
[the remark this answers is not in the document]
sutherlanda: Good task for Katelyn :)

auth-web/src/components/auth/BusinessContactForm.vue
sutherlanda: @saravankumarpa Yeah will look at moving to common components.

@mengdong19 That's true... I did try to include masking options for different regions but it was very difficult.  Will talk to the UX guys about strategies for this.

auth-web/src/store/modules/business/index.ts
```diff
@@ -0,0 +1,56 @@
...
+              if ((createResponse.status === 200 || createResponse.status === 201) && createResponse.data) {
+                this.context.commit('setCurrentBusiness', response.data)
+              }
+            })
+        }
+      })
+  }
+
+  @Action({ rawError: true })
+  public async updateContact (contact: Contact) {
```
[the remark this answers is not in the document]
sutherlanda: Confirmed working in IE

auth-web/src/components/auth/BusinessContactForm.vue
sutherlanda: After discussion with @pksumesh, will add a variable in session storage to indicate whether entering contact info has been skipped in that session or not.

saravankumarpa: > After discussion with @pksumesh, will add a variable in session storage to indicate whether entering contact info has been skipped in that session or not.

or simply , we can always redirect to coops application if there is a token and if the page is not accessed directly.whichever is easier.

pksumesh:
Few suggestions regarding the API:
- Keeping the contact references in entity table : This way we won't be able to extend the contacts dynamically in future without changing the entity model. Also if there is a requirement to remove a contact and add another one, the logic would be hard, and it can cause to lose the history of contacts as well. If the contact table has reference to entity that may solve these issues I guess.
- Contact coupled with entity : contact is not treated as a resource, so there is no option to add/update contact alone and will need to update entity for these use cases which is not semantically correct. If we treat contact as a child resource for entity; /entities/<id>/contacts that might be better for maintenance.

sutherlanda:
@pksumesh these are good suggestions.  We have time, so I'll make those changes.

PR #126, opened 2019-08-09
Changes for adding system alerts in case of paybc is down

saravankumarpa:
*Issue #:*

*Description of changes:*

- Changes for showing paybc downtime banner

_tests are currently broken,I will fix them in the next PR_

*UI Code Review Checklist:*

### DIFF
--- a/auth-web/src/store/modules/payment/index.ts
+++ b/auth-web/src/store/modules/payment/index.ts
@@ -1,0 +1,27 @@
+import { Action, Module, Mutation, VuexModule } from 'vuex-module-decorators'
+import paymentService from '@/services/payment.services'
+import { PaySystemStatus } from '@/models/PaySystemStatus'
+import { AxiosResponse } from 'axios'
+
+@Module({
+  name: 'paymentmodule'
+})
+export default class PaymentModule extends VuexModule {
+    paySystemStatus: PaySystemStatus = {
+      current_status: true,
+      next_schedule_date: null,
+      next_schedule_time: null
+
+    }
+
+    @Mutation
+    public setPaySystemStatus (paySystemStatus: PaySystemStatus) {
+      this.paySystemStatus = paySystemStatus
+    }
+
+    @Action({ commit: 'setPaySystemStatus' })
+    public async fetchPaySystemStatus () {
+      const response: AxiosResponse<PaySystemStatus> = await paymentService.getPaySystemStatus()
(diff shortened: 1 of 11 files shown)


auth-web/src/views/BusinessProfile.vue
```diff
@@ -2,7 +2,7 @@
   <div class="view-container">
     <article>
       <h1>Update Business Profile</h1>
-      <p class="intro-text">{{ $t('businessContactCopy', { businessType })}}</p>
+      <p class="intro-text">It looks like we are missing some contact information for your {{businessType}}. You will need to supply us with a few additional details before you can get started...</p>
```
sutherlanda: Are we going to forego localization for now? I'm fine with it if that's the case, just making sure.

auth-web/src/store/modules/payment/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Mutation
+    public setPaySystemStatus (paySystemStatus: PaySystemStatus) {
+      this.paySystemStatus = paySystemStatus
+    }
+
+    @Action({ commit: 'setPaySystemStatus' })
+    public async fetchPaySystemStatus () {
+      const response: AxiosResponse<PaySystemStatus> = await paymentService.getPaySystemStatus()
```
sutherlanda: You don't have to explicitly state the type of the `response` here.  It will be inferred automatically since you have declared the return type for `getPaySystemStatus` in `payment.services`.  A minor typescript nitpick.

auth-web/src/store/modules/payment/index.ts
```diff
@@ -0,0 +1,27 @@
...
+      return response.data
```
sutherlanda: The `payment.service` should probably extract the data from the response and return that, rather than the action.

auth-web/src/views/BusinessProfile.vue
saravankumarpa: Nope..We will use it..  It has to do something with the Vuetify added in the common-common-components.Weird behavior. I am working on it.

auth-web/src/store/modules/payment/index.ts
saravankumarpa: Hm..makes sense..Just a different pattern , though its better ,we havent used it.So not changing this now.

PR #127, opened 2019-08-12
User Profile API

sutherlanda:
*Issue #:* #956

*Description of changes:*
* Add API endpints for User Profile and Contact management
* Changes to Entity endpoints to accommodate new data model for Contacts
* Extensive unit tests additions/updates

* NOTE: This PR includes a database migration for Contact data model changes.

*API Code Review Checklist:*
 - #### Common
 - #### REST 
 - #### Code structure
 - #### Test Cases
 - #### Code Quality
    - [ ] Linting
- #### DB Migrations

*UI Code Review Checklist:*


sutherlanda:
Screen Shot 2019-08-11 at 7 18 20 PM
Screen Shot 2019-08-11 at 7 18 51 PM

Coverage is less than 90% due to some of the new schemas not being implemented yet.

pksumesh:
Forgot to check this; I think we would need to add the user to a KeyCloak group (to get the roles) when we create the user profile. Please share your thoughts on this. @sutherlanda

sutherlanda:
> Forgot to check this; I think we would need to add the user to a KeyCloak group (to get the roles) when we create the user profile. Please share your thoughts on this. @sutherlanda

Sounds good.  Will create separate task to address that.

auth-api/src/auth_api/services/user.py
```diff
@@ -30,114 +35,137 @@ class User:  # pylint: disable=too-many-instance-attributes
...
         if not token:
             return None
-        user_dao = UserModel.create_from_jwt_token(token)
 
-        if not user_dao:
+        existing_user = UserModel.find_by_jwt_token(token)
+        if existing_user is None:
+            user_model = UserModel.create_from_jwt_token(token)
+        else:
+            user_model = UserModel.update_from_jwt_token(token)
```
saravankumarpa: i am not sure if its possible at all , but can u pass the existing_user to update_from_jwt_token and avoid the second call within the update_from_jwt_token

auth-api/src/auth_api/models/user.py
```diff
@@ -42,16 +43,18 @@ class User(db.Model):
...
     roles = Column('roles', String(1000))
 
+    contacts = relationship('ContactLink', back_populates='user')
+
     @classmethod
-    def find_by_keycloak_guid(cls, keycloak_guid):
+    def find_by_username(cls, username):
         """Return the first user with the provided KeyCloak GUID."""
-        return cls.query.filter_by(keycloak_guid=keycloak_guid).first()
+        return cls.query.filter_by(username=username).first()
```
saravankumarpa: is it username or GUID ? the comment says GUID..or username is the GUID

auth-api/src/auth_api/models/user.py
```diff
@@ -42,16 +43,18 @@ class User(db.Model):
...
 
     @classmethod
     def find_by_jwt_token(cls, token: dict):
-        """Return if they exist and match the provided JWT."""
+        """Find an existing user by the keycloak GUID in the provided token."""
         return cls.query.filter_by(
-            username=token.get('preferred_username', None)
+            keycloak_guid=token.get('sub', None)
```
saravankumarpa: just to bring the abstraction , may be pass only the value of token.get('sub', None) instead of whole token,... i am not sure which one makes sense

auth-api/src/auth_api/models/user.py
```diff
@@ -70,30 +73,38 @@ def create_from_jwt_token(cls, token: dict):
...
+    def update_from_jwt_token(cls, token: dict):
+        """Update a User from the provided JWT."""
+        if token:
+            user = User.find_by_jwt_token(token)
+            if user:
+                user.username = token.get('preferred_username', user.username)
+                user.firstname = token.get('firstname', user.firstname)
+                user.lastname = token.get('lastname', user.lastname)
+                user.email = token.get('email', user.email)
+                user.modified = datetime.datetime.now()
```
saravankumarpa: is there a way in python to set modified to be auto update on every updates?something which wil go in the model and we dont have to write it in every update method

sutherlanda: Probably, but I don't know what that is.  Would be a better way to do it though, so I will look into it.

auth-api/src/auth_api/models/user.py
sutherlanda: Yes, you're right.

auth-api/src/auth_api/models/user.py
sutherlanda: Yep, missed updating the docstring

auth-api/src/auth_api/models/user.py
pksumesh: `@validates` provides option to validate any field before persisting.

sutherlanda: Thanks @pksumesh ! Will check that out

auth-api/src/auth_api/models/user.py
sutherlanda: Actually, I want to keep this one the way it is.  The reason is that if later on we decide to change the field in the token that we use to fetch the user, we'll only have to change it in one place in the model (`find_by_jwt_token`).  If we were to change it to `find_by_keycloak_guid`, we'd have to change it in several places in the service.

auth-api/src/auth_api/models/user.py
sutherlanda: This change will require another database migration, so going to include it in the work being done for Org API.

PR #133, opened 2019-08-14
Post mvp passcode validation (#1023) and Database Schema document update.

mengdong19:
*Issue #:*
https://github.com/bcgov/entity/issues/1023

*Description of changes:*

*API Code Review Checklist:*
 - #### Common
 - #### REST 
 - #### Code structure
 - #### Test Cases
 - #### Code Quality
    - [ ] Linting
- #### DB Migrations

*UI Code Review Checklist:*


auth-api/tests/unit/services/test_entity.py
```diff
@@ -168,3 +168,28 @@ def test_get_contact_by_business_identifier_no_contact(session):  # pylint:disab
...
+    return EntityModel(business_identifier=business_identifier, pass_code=pass_code)
+
+
+def test_validate_pass_code(app, session):  # pylint:disable=unused-argument
+    """Assert that a contact for an entity can be updated."""
+    entity_info = {
+        'businessIdentifier': 'CP12345678',
+        'pass_code': '12345678'
+    }
+    with app.app_context():
```
sutherlanda: Not a big deal, but you don't actually need the `with app.app_context()` here.  Including `session` as a parameter for the test is sufficient.

auth-api/src/auth_api/services/entity.py
```diff
@@ -133,3 +133,13 @@ def get_contact_for_business(business_identifier):
         if contact_link is None:
             return None
         return contact_link.contact
+
+    @staticmethod
+    def validate_pass_code(business_identifier, pass_code):
+        """Get the contact for a business identified by the given id."""
+        entity = EntityModel.find_by_business_identifier(business_identifier)
+        if entity is None:
+            return None
```
sutherlanda: Should this case return `False` instead of `None`?

auth-api/src/auth_api/models/entity.py
```diff
@@ -29,7 +29,7 @@ class Entity(BaseModel):  # pylint: disable=too-few-public-methods
 
     id = Column(Integer, primary_key=True)
     business_identifier = Column('business_identifier', String(75), unique=True, nullable=False)
-
+    pass_code = Column('pass_code', String(75), unique=False, nullable=True)
```
sutherlanda: Should we add a constraint here to make sure pass_code's are unique? @pksumesh ?

auth-api/tests/unit/services/test_entity.py
mengdong19: removed.

auth-api/src/auth_api/models/entity.py
mengdong19: It's not necessary to be unique, as it's generated by some code. We don't know how the code works.

auth-api/src/auth_api/services/entity.py
mengdong19: I thought None will kind of give more information, can identify if it's pass_code error or entity number error. The program call this service have chance to deal with this again to give more detailed error message.

sutherlanda: Understood, but that creates a lot of coupling between the caller and your code.  The caller has to know what a return value of None means in this case.  It's also not a good idea to have an ambiguous return type for your functions (in this case it's a hybrid Boolean/None).

auth-api/src/auth_api/models/entity.py
sutherlanda: That's true.  We'll probably be encrypting passcodes eventually anyway, so it probably doesn't matter.

auth-api/src/auth_api/services/entity.py
pksumesh: > Understood, but that creates a lot of coupling between the caller and your code. The caller has to know what a return value of None means in this case. It's also not a good idea to have an ambiguous return type for your functions (in this case it's a hybrid Boolean/None).

Totally agree, return type should always be same

mengdong19: updated.

PR #147, opened 2019-08-20
1022 create affiliation endpoints

stevenc987:
*Issue #:*

*Description of changes:*

*API Code Review Checklist:*
 - #### Common
 - #### REST 
 - #### Code structure
 - #### Test Cases
 - #### Code Quality
    - [ ] Linting
- #### DB Migrations

*UI Code Review Checklist:*

### DIFF
--- a/auth-api/src/auth_api/models/affiliation.py
+++ b/auth-api/src/auth_api/models/affiliation.py
@@ -16,8 +16,8 @@
 An Affiliation is between an Org and an Entity.
 """
 
-from sqlalchemy import Column, ForeignKey, Integer
-
+from sqlalchemy import Column, ForeignKey, Integer, DateTime
+from sqlalchemy.orm import relationship
 from .base_model import BaseModel
 
 
@@ -26,6 +26,23 @@ class Affiliation(BaseModel):  # pylint: disable=too-few-public-methods # Tempor
 
     __tablename__ = 'affiliation'
 
-    id = Column(Integer, primary_key=True)
+    id = Column(Integer, primary_key=True, autoincrement=True)
     entity = Column(ForeignKey('entity.id'), nullable=False)
     org = Column(ForeignKey('org.id'), nullable=False)
+
+    entityInfo = relationship('Entity')
+
+    @classmethod
+    def find_affiliation_by_ids(cls, org_id: int, affiliation_id: int):
+        """Return the first Affiliation with the provided ids."""
+        return cls.query.filter_by(org=org_id).filter_by(id=affiliation_id).first()
+
+    @classmethod
+    def find_by_affiliation_id(cls, affiliation_id: int):
+        """Return the affiliation with the provided ids."""
+        return cls.query.filter_by(id=affiliation_id).first()
+
+    @classmethod
+    def find_affiliations_by_org_id(cls, org_id: int):
+        """Return the affiliations with the provided id."""
+        return cls.query.filter_by(org=org_id).all()
--- a/auth-api/src/auth_api/resources/affiliation.py
+++ b/auth-api/src/auth_api/resources/affiliation.py
@@ -1,0 +1,118 @@
+# Copyright © 2019 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""API endpoints for managing an affiliation resource."""
+
+from flask import current_app, g, request
+from flask_restplus import Namespace, Resource, cors
+from sqlalchemy import exc
+
+from auth_api import status as http_status
+from auth_api.exceptions import BusinessException
+from auth_api.jwt_wrapper import JWTWrapper
+from auth_api.schemas import utils as schema_utils
+from auth_api.services.affiliation import Affiliation as AffiliationService
+from flask import jsonify
+from auth_api.tracer import Tracer
+from auth_api.utils.roles import Role
+from auth_api.utils.util import cors_preflight
+
+
+API = Namespace('affiliations', description='Affiliations')
+TRACER = Tracer.get_instance()
+_JWT = JWTWrapper.get_instance()
+
+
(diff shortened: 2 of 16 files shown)


auth-api/src/auth_api/models/affiliation.py
```diff
@@ -26,6 +26,23 @@ class Affiliation(BaseModel):  # pylint: disable=too-few-public-methods # Tempor
...
 
-    id = Column(Integer, primary_key=True)
+    id = Column(Integer, primary_key=True, autoincrement=True)
     entity = Column(ForeignKey('entity.id'), nullable=False)
     org = Column(ForeignKey('org.id'), nullable=False)
+
+    entityInfo = relationship('Entity')
+
+    @classmethod
+    def find_affiliation_by_ids(cls, org_id: int, affiliation_id: int):
```
sutherlanda: I don't understand how this method would be useful.  If you already have the affiliation id, you should not need to query by org id and then filter.  If it is required, it should be named differently to better distinguish it from the two methods that follow.

auth-api/src/auth_api/resources/affiliation.py
```diff
@@ -0,0 +1,118 @@
...
+        except BusinessException as exception:
+            response, status = {'code': exception.code, 'message': exception.message}, exception.status_code
+        current_app.logger.debug('>Affiliation.get')
+        return jsonify(response), status
+
+
+@cors_preflight(['GET', 'PUT', 'OPTIONS'])
+@API.route('/<affiliation_id>', methods=['GET', 'PUT', 'OPTIONS'])
+class AffiliationResources(Resource):
+    """Resource for managing entities."""
```
sutherlanda: Docstring needs to be updated.

auth-api/src/auth_api/resources/affiliation.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @staticmethod
+    @_JWT.has_one_of_roles([Role.BASIC.value, Role.PREMIUM.value])
+    @TRACER.trace()
+    @cors.crossdomain(origin='*')
+    def get(org_id, affiliation_id):
+        """Get an existing affiliation by it's org_id and affiliation_id."""
```
sutherlanda: If you have the affiliation id, this is a unique identifier and it should be necessary to search by org and then affiliation.

auth-api/src/auth_api/schemas/schemas/affiliation.json
```diff
@@ -0,0 +1,22 @@
+{
```
sutherlanda: An affiliation should have an org id and an entity id

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -0,0 +1,166 @@
...
+        """Return the affiliation as a python dictionary.
+
+        None fields are not included in the dictionary.
+        """
+        affiliation_schema = AffiliationSchema()
+        obj = affiliation_schema.dump(self._model, many=False)
+        return obj
+
+    @staticmethod
+    def find_affiliation_by_ids(org_id, affiliation_id):
```
sutherlanda: See previous comments about org id and affiliation id

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -0,0 +1,166 @@
...
+
+        for affiliation_model in affiliation_models:
+            if affiliation_model:
+                data['items'].append(Affiliation(affiliation_model).as_dict())
+        current_app.logger.debug('>find_affiliations_by_org_id')
+
+        return data
+
+    @staticmethod
+    def create_affiliation(org_id, entity_info: Dict[str, Any]):
```
sutherlanda: This method should not take an entire entity dictionary.  It should only take the entity id since that is all it needs.  It should be the callers responsibility to retrieve the entity and extract its id.

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -0,0 +1,166 @@
...
+        affiliation.entity = entity.as_dict()['id']
+        affiliation.org = org_id
+
+        affiliation.save()
+        affiliation = Affiliation(affiliation)
+
+        return affiliation.as_dict()
+
+    @staticmethod
+    def update_affiliation(org_id, affiliation_id, entity_info: Dict[str, Any]):
```
sutherlanda: Same comment as `create_affiliation`

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -0,0 +1,166 @@
...
+        """Update an Affiliation."""
+        affiliation = AffiliationModel.find_by_affiliation_id(int(affiliation_id))
```
sutherlanda: This cast to `int` should not be necessary.  It could also throw a runtime exception if the passed `affiliation_id` is not castable to an `int`, which would need to be caught.

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -0,0 +1,166 @@
...
+        if affiliation is None:
+            raise BusinessException(Error.DATA_NOT_FOUND, None)
+
+        org = OrgService.find_by_org_id(int(org_id))
+        if org is None:
+            raise BusinessException(Error.DATA_NOT_FOUND, None)
+
+        entity = EntityService.find_by_entity_id(int(entity_info.get('entity')))
```
sutherlanda: Unnecessary cast to `int`

auth-api/tests/unit/api/test_affiliation.py
```diff
@@ -0,0 +1,248 @@
...
+    entity_info = {
+        'entity': entity.id
+    }
+
+    org = factory_org_model(name='My Test Affiliation', session=session)
+
+    token = jwt.create_jwt(get_claims(), token_header)
+    # print("######################## token {}", token)
+    headers = {'Authorization': f'Bearer {token}', 'content-type': 'application/json'}
+    rv = client.post(f'/api/v1/orgs/' + str(org.id) + '/affiliations', data=json.dumps(entity_info),
```
sutherlanda: This is nitpicky, but you should use string templating here rather than concatenation.  `/api/v1/orgs/{}/affilications'.format(org.id)`

auth-api/tests/unit/api/test_org.py
```diff
@@ -67,6 +67,8 @@
 }
 
 
+
```
sutherlanda: This should make pylint/flake8 complain.

auth-api/src/auth_api/models/affiliation.py
pksumesh: Better naming is needed as Andrew suggested, but I would keep this functionality there, since affiliation is a sub-resource of org, it's always better to check the org first to avoid people fetching all affiliations without org identifier.

PR #183, opened 2019-09-19
Invitation validation and confirmation endpoints

leksmall:
*Issue #:* bcgov/entity#1337, bcgov/entity#1338, bcgov/entity#1176, bcgov/entity#1334

*Description of changes:*
Two endpoints for validating a token and accepting an invitation
Changes in get invitations endpoints to accept a query param for the status.
Changes in the invitation model to include calculated attributes for expiry date and status
Added tests for the new endpoints
Changes to openshift templates to include new environment variables for generating the token
Bug fix - Sign out button on update business profile page
Invitation template styling

I will raise a separate PR for the swagger changes.

*Checklist:*
- [x] No lint errors
- [x] Updated the deployment configs for new environment variable(s)


auth-api/src/auth_api/services/invitation.py
```diff
@@ -46,11 +52,11 @@ def as_dict(self):
         return obj
 
     @staticmethod
-    def create_invitation(invitation_info: dict, user_id):
+    def create_invitation(invitation_info: dict, user_id, user):
```
sutherlanda: If we're passing in `user` now, can we just get `user.id` from that instead of passing in `user_id` separately?

auth-api/src/auth_api/resources/invitation.py
```diff
@@ -131,3 +132,48 @@ def delete(invitation_id):
         except BusinessException as exception:
             response, status = {'code': exception.code, 'message': exception.message}, exception.status_code
         return response, status
+
+
+@cors_preflight('GET,OPTIONS')
```
pksumesh: 'validate' and 'confirm' doesn't sound correct semantically, also its better to avoid verbs. 'validate' is basically a GET by token (and may be an url param to return only active). And 'confirm' is a PATCH on invitations which can return 204.

leksmall: As discussed, changed the verb in url to tokens.

PR #198, opened 2019-10-03
Force user and business profile emails to lowercase

sutherlanda:
*Issue #:* 1447
https://github.com/bcgov/entity/issues/1447

*Description of changes:*

The auth API expects lowercase emails so this change ensure any uppercase characters are converted prior to making the POST.

*Checklist:*
- [x] No lint errors


pksumesh:
We would need this in api side, in case if someone uses the endpoint in future.

sutherlanda:
Discussed with @pksumesh , we'll leave as is on API side since it's currently validating the schema and returning a 400.  This is technically correct behaviour.

PR #223, opened 2019-10-28
Changes to AUTH-WEB to support username display

sutherlanda:
*Issue #:* 1545
https://github.com/bcgov/entity/issues/1545

*Description of changes:*

Add session storage item to support username display in header (sbc-common-components)

*Checklist:*
- [x] No lint errors

### DIFF
--- a/auth-web/src/services/keycloak.services.ts
+++ b/auth-web/src/services/keycloak.services.ts
@@ -24,6 +24,7 @@ export default {
   initSession () {
     configHelper.addToSession(SessionStorageKeys.KeyCloakToken, this.kc.token)
     configHelper.addToSession(SessionStorageKeys.KeyCloakRefreshToken, this.kc.refreshToken)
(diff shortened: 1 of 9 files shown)


auth-web/src/services/keycloak.services.ts
```diff
@@ -24,6 +24,7 @@ export default {
   initSession () {
     configHelper.addToSession(SessionStorageKeys.KeyCloakToken, this.kc.token)
     configHelper.addToSession(SessionStorageKeys.KeyCloakRefreshToken, this.kc.refreshToken)
+    configHelper.addToSession(SessionStorageKeys.UserFullName, this.getUserInfo().firstName)
```
pksumesh: Can we use the 'name' attribute which is basically the display name for users. First Name alone won't be sufficient I guess.

sutherlanda: Right, done.

PR #226, opened 2019-10-29
Terms of use feature

saravankumarpa:
*Issue #:*
https://github.com/bcgov/entity/issues/1542

*Description of changes:*

1. Table to store documents as terms of use , privacy agreement etc ;related services
2. service changes in the user rest endpoint to patch the terms of service agreements 
3. Changes in the UI to support feature
4. Tests are include

  

*Checklist:*
- [ ] No lint errors


auth-api/src/auth_api/models/documents.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Documents(BaseModel):
+    """This is the model for a documents."""
+
+    __tablename__ = 'documents'
+
+    # TODO version concept is not well refined..this is the first version..refine it
+    version_id = Column(Integer, primary_key=True, autoincrement=False)
+    type = Column('type', String(20), nullable=False)
+    content = Column('content', Text)
```
pksumesh: Would it be better if we add start time and expiry time to the model, so that we can set the terms of service upfront with start date. The query would get whatever is latest for the current date.

auth-api/src/auth_api/models/user.py
```diff
@@ -44,6 +44,13 @@ class User(BaseModel):
     contacts = relationship('ContactLink', back_populates='user', primaryjoin='User.id == ContactLink.user_id')
     orgs = relationship('Membership', back_populates='user', primaryjoin='User.id == Membership.user_id')
 
+    is_terms_of_use_accepted = Column(Boolean(), default=False, nullable=True)
+    terms_of_use_accepted_version = Column(
+        ForeignKey('documents.version_id'), nullable=True
+    )
+    terms_of_use_version = relationship('Documents', foreign_keys=[terms_of_use_accepted_version], uselist=False,
```
pksumesh: This would cause the document to be loaded every time on the user query which may have some performance impact as the document is TEXT, removing the  `terms_of_use_version = relationship('Documents', foreign_keys=[terms_of_use_accepted_version], uselist=False,` would be better I think.

auth-api/src/auth_api/resources/user.py
```diff
@@ -218,3 +218,33 @@ def get():
         """Add a new contact for the Entity identified by the provided id."""
         sub = g.jwt_oidc_token_info.get('sub', None)
         return AuthorizationService.get_user_authorizations(sub), http_status.HTTP_200_OK
+
+
+@cors_preflight('PATCH, OPTIONS')
+@API.route('/termsofuse', methods=['PATCH', 'OPTIONS'])
```
pksumesh: This can be a patch on user resource (not a termsofuse endpoint), as it's basically patching the user resource with terms of use details.

auth-api/src/auth_api/resources/user.py
```diff
@@ -218,3 +218,33 @@ def get():
...
+    @TRACER.trace()
+    @cors.crossdomain(origin='*')
+    @_JWT.requires_auth
+    def patch():
+        """Update terms of service for the user."""
+        token = g.jwt_oidc_token_info
+        request_json = request.get_json()
+
+        valid_format, errors = schema_utils.validate(request_json, 'termsofuse')
+        if not token:
```
pksumesh: This token check is not needed, but I think we have techdebt ticket remove these unwanted token checks

auth-api/src/auth_api/services/documents.py
```diff
@@ -0,0 +1,55 @@
...
+        """Return the User as a python dict.
+
+        None fields are not included in the dict.
+        """
+        document_schema = DocumentSchema()
+        obj = document_schema.dump(self._model, many=False)
+        return obj
+
+    @classmethod
+    def fetch_terms_of_use_document(cls, document_type):
```
pksumesh: If Documents model/service is created to handle all the document related operations (not specific to terms of use), renaming this def to fetch document might be better, as the document type is always passed as argument.

saravankumarpa: Changing this

auth-api/src/auth_api/resources/user.py
saravankumarpa: Agreed. writing a decent patch method supporting the patch for a user is tough..Thats why went for this approach , though bad in terms of rest principles..will change this with a TODO
something like below is ideal , but havent see any

[
{ “op”: “update”, “path”: “/termsflag”, “value”: true }
]

PR #250, opened 2019-11-25
Switch email service

pwei1018:
*Issue #:*

*Description of changes:*

*Checklist:*
- [x] No lint errors
- [x] Updated the deployment configs for new environment variable(s)


pwei1018:
image

pwei1018:
image

pwei1018:
image

auth-api/src/auth_api/services/invitation.py
```diff
@@ -165,16 +153,13 @@ def send_invitation(invitation: InvitationModel, org_name, user, confirm_url):
...
-            thread = Thread(target=run_job)
-            thread.start()
-
-        except:  # noqa: E722
+        sent_response = send_email(subject, sender, recipient,
+                                   template.render(invitation=invitation,
+                                                   url=token_confirm_url,
+                                                   user=user,
+                                                   org_name=org_name))
+        print(sent_response)
```
pksumesh: Can we remove this print?

auth-api/src/auth_api/services/notification.py
```diff
@@ -12,14 +12,28 @@
...
-    mail.send(msg)
+    notify_url = current_app.config.get('NOTIFY_API_URL') + '/notify/'
+    notify_body = {
+        'recipients': recipients,
+        'contents': {
+            'subject': subject,
+            'body': html_body
+        }
+    }
+    notify_response = RestService.post(notify_url, data=notify_body)
```
pksumesh: Is notify open for public? Should we re-consider securing it with token from our realm?

pwei1018: I don't think we open the notify service to public for now. In the Openshift environment, I pan to use service address instead of public route. Next step, we can think about it if need open. FastAPI support JWT token pretty easy.

auth-api/src/auth_api/services/invitation.py
pwei1018: Fixed.

auth-api/src/auth_api/services/notification.py
pksumesh: Ohh ok. If route is not exposed, all good!

PR #269, opened 2019-11-30
Fix some contact email not save and wrong team name for some membership pending approval status user.

mengdong19:
*Issue #:*
https://github.com/bcgov/entity/issues/2001
https://github.com/bcgov/entity/issues/2019

*Description of changes:*

*Checklist:*
- [ ] No lint errors


auth-web/src/components/auth/NextPageMixin.vue
```diff
@@ -19,6 +19,8 @@ export default class NextPageMixin extends Vue {
         nextStep = Pages.USER_PROFILE
       } else if (organizations.length === 0) {
         nextStep = Pages.CREATE_TEAM
+      } else if (organizations.some(org => org.members[0].membershipStatus === 'ACTIVE')) {
```
sutherlanda: I'm not understanding why we would need this change or how it is related to #2001.  Currently, we check if a membership is PENDING, and then if it's not (i.e. we check in the else clause if it's an ACTIVE membership), we go to the dashboard already.  Let's not add this extra clause.

#2001 is a symptom of already being a member of a team and receiving an invite.  We should look at the logic that generates the invite email to make sure it's using the user's active team.

Another change that we should make is that instead of using checking membership on `org.members[0]`, we map the getter `myOrgMembership` from the store and check that.  I'm not sure that `members[0]` will always be the current user.

auth-web/package.json
```diff
@@ -56,7 +56,7 @@
     "stylus": "^0.54.7",
     "stylus-loader": "^3.0.2",
     "ts-jest": "^23.0.0",
-    "typescript": "^3.2.1",
+    "typescript": "^3.7.2",
```
sutherlanda: Please confirm that this will not generate any additional type errors in the solution - that's a big version jump.

auth-web/src/components/auth/NextPageMixin.vue
mengdong19: First of all, I'll help you to understand why this is related to #2001, and why this code will fix it. The cause of the wrong org name issue happens when user is member of multiple organizations, and some are active, some are pending. 
Say user A is already an active member of organization Org1 (org id 1), he got invited to organization  Org5 (org id 5), he accepted the invitation and login,  now he's member of two organizations, and they are ordered by org_id in array:
organizations[0]: org_id=1, members[0].membershipStatus = "ACTIVE"
organizations[1]: org_id=5, members[0].membershipStatus = "PENDING_APPROVAL"

With previous code, once there is "PENDING_APPROVAL" in any of the membership status, user will be directed to pendingapproval/organization[0].name, in the case described above,  user A got redirect to pendingapproval/Org1, so he'll see team "Org1"  in pendingapproval page after accepted invitation from team "Org5", this is what issue 2001 described. 
We had a discussion on RocketChat and we all agree to redirect user to Main page when user has an active membership, that's what the new code do, it redirect user to Main when there is an active team, no matter how many membership he has and what status he is in other team, so he can at least do something to manage his active org and business.
With the new code, Active status has higher priority than Pending status, in the case described above, user A will be redirect to main. 
The logic of generates the invite email has nothing to do with #2001. 
As for this.organizations, it checking organizations only for this user's membership, so org.members[0] is current user.

sutherlanda: `The cause of the wrong org name issue happens when user is member of multiple organizations, and some are active, some are pending.`

@mengdong19 We are operating on the assumption that a user cannot be a member of more than one org at a time.  If they are able to be a member of more than org somehow, then that is a bug and should be raised in a separate ticket.  We are not going to introduce extra logic to handle a case that should not happen.  Your fix for the email is valid and I'm happy to merge the PR if you remove the changes related to #2001, but the redirect change you are proposing is not valid.

mengdong19: Maybe it should not happen for now, but it would happen later, what's wrong with prepare for the future case when we can? I'm not going to remove this code because it fix #2001 and also works for the future.  Send Invitation logic will be another topic that is not related to #2001, accept invitation logic is also another topic.

sutherlanda: Feel free to leave your code here, but it will not be merged as is.

sutherlanda:
@saravankumarpa can you please review the above when you have a moment?

PR #271, opened 2019-12-01
#2016 - Prevent duplicate invites from being generated

sutherlanda:
*Issue #:* 2016
https://github.com/bcgov/entity/issues/2016

*Description of changes:*

Fix a bug where multiple invites could be generated for the same email address.  This was fixed through the following changes on the front-end:

a) Validation was added to the invite users form to prevent duplicate emails being specified simultaneously
b) The invite users form is reset whenever it is closed
c) Prior to creating an invitation, pending invitations are checked for the same email.  If found, the existing is just resent instead of creating a new one.

*Checklist:*
- [x] No lint errors

### DIFF
--- a/auth-web/src/components/auth/InviteUsersForm.vue
+++ b/auth-web/src/components/auth/InviteUsersForm.vue
@@ -79,11 +79,12 @@ interface InvitationInfo {
 
 @Component({
   computed: {
+    ...mapState('org', ['pendingOrgInvitations']),
     ...mapGetters('org', ['myOrg', 'myOrgMembership'])
   },
   methods: {
     ...mapMutations('org', ['resetInvitations']),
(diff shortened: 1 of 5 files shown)


auth-web/src/components/auth/InviteUsersForm.vue
```diff
@@ -135,8 +138,24 @@ export default class InviteUsersForm extends Vue {
     }
   }
 
+  private hasDuplicates (): boolean {
```
saravankumarpa: i am not very sure  , so just wondering if this be achieved using  new Set(array)).size !== array.length? If not , its good.

sutherlanda: Great idea - done :)

PR #290, opened 2019-12-09
Allow owners to change their role to admin if other owners on team

sutherlanda:
*Issue #:* 1966
https://github.com/bcgov/entity/issues/1966

*Description of changes:*

Make UI and API changes to support Owners downgrading their own role.  While implementing this ticket, I also tightened up checks on the API for Owners removing themselves (ensuring there is one other owner) and for auth checks on downgrading/upgrading roles.

NOTE: There isn't currently a set of Pytests for membership service calls, so I haven't modified any tests.  We should open up a separate ticket for updating the pytest suite for membership operations.

*Checklist:*
- [x] No lint errors


auth-api/src/auth_api/services/membership.py
```diff
@@ -143,8 +156,14 @@ def update_membership(self, updated_fields, token_info: Dict = None):
     def deactivate_membership(self, token_info: Dict = None):
         """Mark this membership as inactive."""
         current_app.logger.debug('<deactivate_membership')
+        # if this is a member removing another member, check that they admin or owner
         if self._model.user.username != token_info.get('username'):
             check_auth(token_info=token_info, one_of_roles=(ADMIN, OWNER))
```
pksumesh: I know this is old code. Just wondering whether this would work as there are no org_id or business_identifier passed as args.

auth-api/src/auth_api/services/membership.py
```diff
@@ -143,8 +156,14 @@ def update_membership(self, updated_fields, token_info: Dict = None):
...
+
+        # check to ensure that owner isn't removed by anyone but an owner
+        if self._model.membership_type == OWNER:
```
pksumesh: Same as above comment

auth-api/src/auth_api/services/membership.py
sutherlanda: Good catch, I've added the `org_id` param

PR #304, opened 2019-12-18
changes for vue cli 4

saravankumarpa:
*Issue #:*
https://github.com/bcgov/entity/issues/2171

*Description of changes:*

The changes should have been only package.json ideally.
But with the vue cli 4.1 , they removed a behavior which we relying on .They dont prepend slashes .So we have to add slashes in the process.env .That brought in changes in some other files.

*Checklist:*
- [ ] No lint errors


auth-web/package.json
```diff
@@ -55,10 +55,10 @@
     "sinon": "^6.3.5",
     "stylus": "^0.54.7",
     "stylus-loader": "^3.0.2",
-    "ts-jest": "^23.0.0",
-    "typescript": "^3.2.1",
+    "ts-jest": "23.0.0",
+    "typescript": "~3.5.3",
```
sutherlanda: Not going up to 3.6 or 3.7?

saravankumarpa: let me give a try with 3.7 ..if it works , its good to go with 3.7 ... this is the version which vue cli proposes.. But 3.7 is great..Let me give a try

saravankumarpa: I think we used it already.so will just work.. I will just try an optional chaining feature just to see how it goes

sutherlanda: Awesome

PR #324, opened 2020-01-08
Multiple email has to be separated by comma instead of spaces

saravankumarpa:
Multiple email has to be separated by comma instead of spaces

*Issue #:*
https://github.com/bcgov/entity/issues/2260

*Description of changes:*

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-api/src/auth_api/services/invitation.py
+++ b/auth-api/src/auth_api/services/invitation.py
@@ -213,11 +213,7 @@ class Invitation:
         admin_list = UserService.get_admins_for_membership(membership_id)
         invitation: InvitationModel = InvitationModel.find_invitation_by_id(invitation_id)
         context_path = CONFIG.AUTH_WEB_TOKEN_CONFIRM_PATH
-        admin_emails = ''
-        for contact in admin_list:
-            if contact.contacts:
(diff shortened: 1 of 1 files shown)


sutherlanda:
What was the root cause of the bug addressed by this?  The original ZenHub issue is unclear.

saravankumarpa:
> What was the root cause of the bug addressed by this? The original ZenHub issue is unclear.

Added the reply.
This got introduced when we switched to the notification service.The old mail service accepted spaces , but the new mail notify service needed commas instead

sutherlanda:
> > What was the root cause of the bug addressed by this? The original ZenHub issue is unclear.
> 
> Added the reply.
> This got introduced when we switched to the notification service.The old mail service accepted spaces , but the new mail notify service needed commas instead

Ah okay, thanks for clarifying @saravankumarpa

saravankumarpa:
> > > What was the root cause of the bug addressed by this? The original ZenHub issue is unclear.
> > 
> > 
> > Added the reply.
> > This got introduced when we switched to the notification service.The old mail service accepted spaces , but the new mail notify service needed commas instead
> 
> Ah okay, thanks for clarifying @saravankumarpa

I hope it fixes. Not 100% sure the bug. But there is a definite bug here which we have to fix

PR #358, opened 2020-01-22
#1981 Changes for Account settings page and Account menu

saravankumarpa:
*Issue #:*
https://github.com/bcgov/entity/issues/1981

Changes related to 
https://github.com/bcgov/entity/issues/2293
https://github.com/bcgov/entity/issues/2093
are also included

*Description of changes:*
Account settings page 
removed menu from main/dashboard
other related changes 

*Checklist:*
- [ ] No lint errors


auth-api/src/auth_api/services/org.py
```diff
@@ -74,6 +74,11 @@ def create_org(org_info: dict, user_id):
     def update_org(self, org_info):
         """Update the passed organization with the new info."""
         current_app.logger.debug('<update_org ')
+
+        existing_similar__org = OrgModel.find_similar_org_by_name(org_info['name'])
```
pwei1018: Need testcase for this change?

auth-web/src/services/keycloak.services.ts
```diff
@@ -46,6 +46,13 @@ class KeyCloakService {
     configHelper.addToSession(SessionStorageKeys.KeyCloakRefreshToken, this.kc.refreshToken)
     configHelper.addToSession(SessionStorageKeys.UserFullName, this.getUserInfo().fullName)
     this.parsedToken = this.kc.tokenParsed as UserToken
+
+    // Set flag in session storage so that common components will know account type
+    if (this.getUserInfo().userName.endsWith('@idir')) {
```
pksumesh: There is a 'loginSource' claim in token which can be used here to identify the login type. Its value would be either BCSC or IDIR.

auth-web/src/App.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <v-app id="app">
     <div class="header-group" ref="headerGroup">
-      <sbc-header :key="$route.fullPath"></sbc-header>
+      <sbc-header
+        :key="$route.fullPath"
+        :account-name="currentOrganization && currentOrganization.name"
+        :pending-approval-count="pendingActionCount"></sbc-header>
```
pksumesh: Just some thoughts here : since we are passing the values to header, may not work in cases where the user is in another application (coops, create-business, ppr etc.) and won't see the changes when someone else is working on the notification actions from another browser. Also with the new strategy where the applications can work standalone, this solution would cause some issues I guess. May be we should think about making the header stands by own, rather than passing the values from parent application.

auth-web/src/App.vue
```diff
@@ -12,33 +15,51 @@
...
-  mounted (): void {
-    // eslint-disable-next-line no-console
-    console.log('APP.vue mounted')
+  computed: {
+    ...mapState('org', ['currentOrganization', 'pendingOrgMembers', 'myOrgMembership']),
+    ...mapGetters('org', ['myOrgMembership'])
+  }
+})
+export default class App extends Vue {
+  private orgStore = getModule(OrgModule, this.$store)
```
pksumesh: ^^ orgStore - seems not needed.

auth-web/src/components/auth/AccountInfo.vue
```diff
@@ -0,0 +1,145 @@
...
+  private orgName = ''
+  private readonly orgCreateMessage
+  private touched = false
+
+  private isFormValid (): boolean {
+    return !!this.orgName
+  }
+
+  private async mounted () {
+    this.orgStore.setOrgCreateMessage('success') // reset
```
pksumesh: Can we use mapMutations rather than setting to store explicitly?

pksumesh:
There are 36 commits, whoever is merging this PR please do a squash merge!

auth-api/src/auth_api/services/org.py
saravankumarpa: adding

saravankumarpa: Added tests for this.

auth-web/src/App.vue
sutherlanda: Sorry @pksumesh I didn't see these comments until now.  I'm going to create a ticket to address this by adding a service layer in the common component so that it can make calls to our API as required.

PR #418, opened 2020-02-13
Bug fixes -owner count message corrected

saravankumarpa:
*Issue #:*
https://github.com/bcgov/entity/issues/1917

*Description of changes:*

Added maximum number of organisation reached message added
Removed the check for organsation membership

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/views/auth/DashboardView.vue
+++ b/auth-web/src/views/auth/DashboardView.vue
@@ -68,10 +68,9 @@ export default class DashboardView extends Vue {
     }
 
     // If user is not an active member of a team at all, redirect to create team page
-    /* if (!this.myOrgMembership) {
+    if (!this.myOrgMembership) {
       this.$router.push('/createaccount')
     }
-    */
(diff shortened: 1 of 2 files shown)


auth-web/src/views/auth/DashboardView.vue
```diff
@@ -68,9 +68,10 @@ export default class DashboardView extends Vue {
     }
 
     // If user is not an active member of a team at all, redirect to create team page
-    if (!this.myOrgMembership) {
+    /* if (!this.myOrgMembership) {
```
sutherlanda: I think we will still need this check in case a user without an account navigates directly to the Dashboard.

The problem must be that membership is not getting synced after the new account is created.  If we put a `syncMembership` call immediately after creating the account, that may be all we need.

saravankumarpa: i took this out..may be in another PR

PR #422, opened 2020-02-14
Add prop handling for redirect urls (#2563)

sutherlanda:
*Issue #:* 2563
https://github.com/bcgov/entity/issues/2563

*Description of changes:*

These are required changes in auth-web to support the changes to common components allowing a post-login URL to be specified.

*Checklist:*
- [x] No lint errors

### DIFF
--- a/auth-web/src/views/auth/SigninView.vue
+++ b/auth-web/src/views/auth/SigninView.vue
@@ -52,10 +52,11 @@ export default class Signin extends Mixins(NextPageMixin) {
 
   @Prop({ default: 'bcsc' }) idpHint: string
   @Prop({ default: '' }) redirectUrl: string
+  @Prop({ default: '' }) redirectUrlLoginFail: string
 
   private async mounted () {
     if (this.redirectUrl) {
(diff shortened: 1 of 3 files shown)


auth-web/src/views/auth/SigninView.vue
```diff
@@ -52,58 +52,44 @@ export default class Signin extends Mixins(NextPageMixin) {
...
+              })
+            }
+            resolve()
           }
-          resolve()
-        }
-      })
+        })
+        .error(() => {
+          if (this.redirectUrlLoginFail) {
```
pksumesh: If the user clicks on cancel in BC Services Card login, the it will take the user back to home page as it's a pre-configured thing in BCSC. Not sure, when this block will be executed.

sutherlanda: Oh okay, I didn't realize that.  I suppose this wouldn't be fired at all then, unless KeyCloak was unavailable or something.  Should I remove?

pksumesh: I think we can keep the code there, won't do any harm.

PR #432, opened 2020-02-20
Fix issue with pytests for resetting

sutherlanda:
Fix an issue causing the tests for reset users to fail.


pwei1018:
You guys need to fix that flake8 issue as well.

PR #442, opened 2020-02-25
Create signin and signout component in sbc-common-components

nitheesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/2657

* added sbc components for signin and signout
* moved keycloak service to sbc-common-components

*Checklist:*
- [x] No lint errors


auth-web/src/store/modules/org.ts
```diff
@@ -5,7 +5,7 @@ import { AccountSettings } from '@/models/account-settings'
 import { EmptyResponse } from '@/models/global'
 import InvitationService from '@/services/invitation.services'
 import OrgService from '@/services/org.services'
-import { UserInfo } from '@/models/userInfo'
+import { UserInfo } from 'sbc-common-components/src/models/userInfo'
```
sutherlanda: While we move this file to common, can we also rename to something other than `userInfo`? Maybe `kcUserProfile`?

sutherlanda: If this too big of a change, no problem - it's not high priority.

nitheesh-aot: I agree, that name makes sense.
will change that

nitheesh-aot: done

PR #447, opened 2020-02-27
Add navigation guard in router for user profile (#2379)

sutherlanda:
*Issue #:* 2379
https://github.com/bcgov/entity/issues/2379

*Description of changes:*

Add a navigation guard to prevent users from performing certain actions in the front-end before they have set up their user profile and accepted the terms of service.

Also remove unused interceptor file in auth-web

*Checklist:*
- [x] No lint errors


sutherlanda:
> I think we can keep the interceptor.ts even though its not used now.If we go back to interceptor for token refresh , the code can be just activated.Once its removed , the somebody will have to write it from scratch.

What are the chances we will move this back to from common-components?

saravankumarpa:
> > I think we can keep the interceptor.ts even though its not used now.If we go back to interceptor for token refresh , the code can be just activated.Once its removed , the somebody will have to write it from scratch.
> 
> What are the chances we will move this back to auth-web from common-components?

hm..Not sure of how/where its gonna end up .Just thought let it be there for a back up

sutherlanda:
> > > I think we can keep the interceptor.ts even though its not used now.If we go back to interceptor for token refresh , the code can be just activated.Once its removed , the somebody will have to write it from scratch.
> > 
> > 
> > What are the chances we will move this back to auth-web from common-components?
> 
> hm..Not sure of how/where its gonna end up .Just thought let it be there for a back up

Okay, no problem.  Added back in.

PR #477, opened 2020-03-10
Changes for Create a director search account

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/2693

*Description of changes:*

API changes for 

- create org by staff for director search

- changes for notification 

- product subscription api changes

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-api/src/auth_api/resources/org.py
+++ b/auth-api/src/auth_api/resources/org.py
@@ -31,7 +31,6 @@ from auth_api.utils.enums import NotificationType
 from auth_api.utils.roles import ALL_ALLOWED_ROLES, CLIENT_ADMIN_ROLES, MEMBER, Role, Status
 from auth_api.utils.util import cors_preflight
 
-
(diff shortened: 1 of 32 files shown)


auth-api/src/auth_api/resources/org.py
```diff
@@ -56,15 +55,18 @@ def post():
         valid_format, errors = schema_utils.validate(request_json, 'org')
         if not valid_format:
             return {'message': schema_utils.serialize(errors)}, http_status.HTTP_400_BAD_REQUEST
-
+        is_staff = token and 'staff' in token.get('realm_access').get('roles')
```
pwei1018: Should it be 'staff_admin' here?

saravanpa-aot: Good catch..But I am removing that code since now staff has entry in user table..so we dont have to check for it..But good catch

saravanpa-aot: changed it

PR #502, opened 2020-03-18
User creation APIs to support director search anonymous user

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/2694

*Description of changes:*

Two end points to support director search. One for the username/password owner user creation and other one for owner to submit bulk users.

I am still writing tests.Nitheesh and sumesh needs these APIs.so putting it here for review.Please have a look and review so that I can incorporate the comments.Will checkin the tests very soon.

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-api/migrations/versions/68e2f43b9d22_added_billable_column.py
+++ b/auth-api/migrations/versions/68e2f43b9d22_added_billable_column.py
@@ -18,7 +18,7 @@ depends_on = None
 
 
 def upgrade():
-    op.add_column('org', sa.Column('billable', sa.Boolean(), nullable=True))
+    op.add_column('org', sa.Column('billable', sa.Boolean(), server_default=sa.schema.DefaultClause("1"), nullable=False))
 
 
 def downgrade():
--- a/auth-api/migrations/versions/68e2f43b9d22_added_billable_column.py
+++ b/auth-api/migrations/versions/68e2f43b9d22_added_billable_column.py
@@ -1,0 +1,25 @@
+"""added billable column
+
+Revision ID: 68e2f43b9d22
+Revises: abc8cc7b64e1
+Create Date: 2020-03-17 23:00:51.705492
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import table, column
+from sqlalchemy import String, Boolean, Date
+
+# revision identifiers, used by Alembic.
+revision = '68e2f43b9d22'
+down_revision = 'abc8cc7b64e1'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('org', sa.Column('billable', sa.Boolean(), nullable=True))
+
+
+def downgrade():
+    op.drop_column('org', 'billable')
--- a/auth-api/src/auth_api/models/org.py
+++ b/auth-api/src/auth_api/models/org.py
@@ -17,7 +17,7 @@ Basic users will have an internal Org that is not created explicitly, but implic
 """
 
 from flask import current_app
-from sqlalchemy import Column, ForeignKey, Integer, String, and_, func
+from sqlalchemy import Column, ForeignKey, Integer, String, and_, func, Boolean
 from sqlalchemy.orm import relationship
 
 from auth_api.utils.roles import OrgStatus as OrgStatusEnum
@@ -48,6 +48,7 @@ class Org(BaseModel):  # pylint: disable=too-few-public-methods
     invitations = relationship('InvitationMembership', cascade='all,delete,delete-orphan', lazy='select')
     products = relationship('ProductSubscription', cascade='all,delete,delete-orphan', lazy='select')
     access_type = Column(String(250), index=True, nullable=True)  # for ANONYMOUS ACCESS
+    billable = Column('billable', Boolean(), default=False)
 
     @classmethod
     def create_from_dict(cls, org_info: dict):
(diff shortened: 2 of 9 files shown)


auth-api/src/auth_api/models/org.py
```diff
@@ -48,6 +48,7 @@ class Org(BaseModel):  # pylint: disable=too-few-public-methods
     invitations = relationship('InvitationMembership', cascade='all,delete,delete-orphan', lazy='select')
     products = relationship('ProductSubscription', cascade='all,delete,delete-orphan', lazy='select')
     access_type = Column(String(250), index=True, nullable=True)  # for ANONYMOUS ACCESS
+    billable = Column('billable', Boolean(), default=False)
```
sumesh-aot: Billable default should be True (all the accounts are billable except government accounts I guess). For dir search account it will be no-billable (billable=False). Also include a migration to add the value True for all existing accounts.

auth-api/src/auth_api/resources/__init__.py
```diff
@@ -86,6 +86,7 @@
 API.add_namespace(ACCOUNTS_API, path='/accounts')
 API.add_namespace(ORG_PRODUCTS_API, path='/orgs/<string:org_id>/products')
 API.add_namespace(NOTIFICATIONS_API, path='/users/<string:user_id>/org/<string:org_id>/notifications')
+API.add_namespace(USER_API, path='/users/<string:invitation_token>')
```
sumesh-aot: Can we add the invitation token as a header and not as a path param.

auth-api/src/auth_api/resources/user.py
```diff
@@ -32,12 +33,68 @@
...
+            InvitationService.accept_invitation(invitation['id'], None, None, False)
+            response, status = user, http_status.HTTP_201_CREATED
+
+        except BusinessException as exception:
+            response, status = {'code': exception.code, 'message': exception.message}, exception.status_code
+        return response, status
+
+
+@cors_preflight('POST,OPTIONS')
+@API.route('/bulk', methods=['POST', 'OPTIONS'])
```
sumesh-aot: Normally the last path in the endpoint is a resource. So to stick with standards, can we have the endpoint as /bulk/<resources> ?

auth-api/src/auth_api/services/org.py
```diff
@@ -75,6 +75,7 @@ def create_org(org_info: dict, user_id, token_info: Dict = None):
         org = OrgModel.create_from_dict(camelback2snake(org_info))
         if is_staff_admin:
             org.access_type = AccessType.ANONYMOUS.value
+            org.billable = True
```
sumesh-aot: This needs to be False

auth-api/migrations/versions/68e2f43b9d22_added_billable_column.py
```diff
@@ -0,0 +1,25 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '68e2f43b9d22'
+down_revision = 'abc8cc7b64e1'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('org', sa.Column('billable', sa.Boolean(), nullable=True))
```
sutherlanda: This shouldn't be nullable.

auth-api/src/auth_api/resources/user.py
```diff
@@ -32,12 +33,68 @@
...
+    @_JWT.requires_auth
+    def post():
+        """Admin users can post multiple users to his org.Use it for anonymous purpose only."""
+        try:
+
+            valid_format, errors = schema_utils.validate(request.get_json(), 'bulk_user')
+            if not valid_format:
+                return {'message': schema_utils.serialize(errors)}, http_status.HTTP_400_BAD_REQUEST
+
+            users = UserService.create_user_and_add_membership(request.get_json()['users'],
```
sutherlanda: You should call get_json once and store the result instead of calling it multiple times.

auth-api/src/auth_api/services/user.py
```diff
@@ -63,6 +67,65 @@ def as_dict(self):
...
+
+        current_app.logger.debug('create_user')
+        users = []
+        for membership in memberships:
+            create_user_request = KeycloakUser()
+            username = membership['username']
+            create_user_request.user_name = username
+            create_user_request.password = membership['password']
+            try:
+                # TODO may be this method itself throw the business exception
```
sutherlanda: Are we going to do this? Can we do it now?

auth-api/src/auth_api/services/user.py
```diff
@@ -63,6 +67,65 @@ def as_dict(self):
...
+            if existing_user:
+                raise BusinessException(Error.DATA_ALREADY_EXISTS, None)
+            user_model: UserModel = UserModel()
+            user_model.username = username
+            user_model.keycloak_guid = kc_user.id
+            user_model.is_terms_of_use_accepted = False
+            user_model.status = Status.ACTIVE.value
+            user_model.type = AccessType.ANONYMOUS.value
+            user_model.email = membership.get('email', None)
+            # TODO see how to get roles
```
sutherlanda: Same as above.

auth-api/src/auth_api/models/org.py
saravanpa-aot: sure..will do

auth-api/migrations/versions/68e2f43b9d22_added_billable_column.py
saravanpa-aot: yup..changint it

auth-api/src/auth_api/resources/user.py
saravanpa-aot: good catch

auth-api/src/auth_api/services/user.py
saravanpa-aot: Let me see..Not sure if its gonna affect any other functionalities.will come back on this

auth-api/src/auth_api/services/user.py
saravanpa-aot: just dicussed this with @sumesh-aot .Roles not needed at all..will remove the todo

auth-api/src/auth_api/services/org.py
saravanpa-aot: oh..is the director search account non-bilable? ?I thought otherwise

PR #527, opened 2020-03-25
3094 fix business profile errors, optimize syncing

sutherlanda:
*Issue #:* 3094

*Description of changes:*

* Fixes an issue with the business profile getting out of sync with co-ops.
* Improve performance by removing unnecessary syncing/routing.
* Fix issue with Create a New Account button on Home view

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/App.vue
+++ b/auth-web/src/App.vue
@@ -31,7 +31,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Mixins, Watch } from 'vue-property-decorator'
+import { Component, Mixins } from 'vue-property-decorator'
 import { Member, MembershipStatus, Organization } from '@/models/Organization'
 import { Pages, SessionStorageKeys } from '@/util/constants'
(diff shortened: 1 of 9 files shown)


sutherlanda:
> I think we need to make the business contact to accept business identifier as a path or route param (an not use the session storage value), the reason being since filings application accepts business identifier as a path param; there would be cases where the user can work on multiple business filings at the same time. Relying on on session storage won't work in this case.

Agreed, but can we do that as a separate ticket after this one? We'll have to coordinate that change with co-ops.

sumesh-aot:
> > I think we need to make the business contact to accept business identifier as a path or route param (an not use the session storage value), the reason being since filings application accepts business identifier as a path param; there would be cases where the user can work on multiple business filings at the same time. Relying on on session storage won't work in this case.
> 
> Agreed, but can we do that as a separate ticket after this one? We'll have to coordinate that change with co-ops.

Sure, will create a ticket

PR #540, opened 2020-03-26
redirection after login bcros

nitheesh-aot:
*Description of changes:*
redirection after login bcros

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/views/auth/SigninView.vue
+++ b/auth-web/src/views/auth/SigninView.vue
@@ -39,6 +39,7 @@ export default class Signin extends Mixins(NextPageMixin) {
   @Prop({ default: '' }) redirectUrlLoginFail: string
 
   private async authenticationComplete () {
+    await this.loadUserInfo()
     // Check if user is authenticated, and redirect according to specified redirect
     // or fallback to default route for their login source
     if (this.$store.getters['auth/isAuthenticated']) {
@@ -46,6 +47,7 @@ export default class Signin extends Mixins(NextPageMixin) {
       if (this.redirectUrl) {
         this.redirectTo(decodeURIComponent(CommonUtils.isUrl(this.redirectUrl) ? this.redirectUrl : `/${this.redirectUrl}`))
       } else {
+        await this.syncUser()
         switch (this.idpHint) {
           case IdpHint.BCSC:
             this.redirectTo(this.getNextPageUrl())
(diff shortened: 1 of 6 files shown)


auth-web/src/views/auth/SigninView.vue
```diff
@@ -39,13 +39,15 @@ export default class Signin extends Mixins(NextPageMixin) {
   @Prop({ default: '' }) redirectUrlLoginFail: string
 
   private async authenticationComplete () {
+    await this.loadUserInfo()
```
sutherlanda: Don't need await here - `loadUserInfo` is not an asynchronous call.

auth-web/src/views/auth/SigninView.vue
```diff
@@ -39,13 +39,15 @@ export default class Signin extends Mixins(NextPageMixin) {
...
     // Check if user is authenticated, and redirect according to specified redirect
     // or fallback to default route for their login source
     if (this.$store.getters['auth/isAuthenticated']) {
       this.$root.$emit('signin-complete')
       if (this.redirectUrl) {
         this.redirectTo(decodeURIComponent(CommonUtils.isUrl(this.redirectUrl) ? this.redirectUrl : `/${this.redirectUrl}`))
       } else {
+        await this.syncUser()
```
sutherlanda: So I don't think this call is necessary.  We are emitting `signin-conplete` above, which tells App.vue to load user info and sync the user/account.  Are you sure this is what was causing the issue in NextPageMixin?

auth-web/src/views/auth/SigninView.vue
nitheesh-aot: actually, we need that, that was the issue that the currentuser value were not setting before the call. either await or settimeout solved that.

auth-web/src/views/auth/SigninView.vue
sutherlanda: Actually,  you're right that calling syncUser here does fix it.  There's a race condition between the handler for `signin-complete` in App.vue and the logic in this comopnent.  Unfortunately, this fix means we'll be syncing the user/account twice.  I'm going to take another look at this.

nitheesh-aot: Yes, thats needed,
because the NextPageMixin needs the value for membership and userterms, since the event is fired along with the redirecturl call, the delay for all the api calls makes those values missing in the NextPageMixin 
calling await syncUser solves that issue

sutherlanda: Yeah, I get that now.  The only thing is we'll be syncing the user twice.  I'm okay to merge this now, but I'll take a look at improving that flow.  We probably shouldn't emit an event to trigger syncing.

PR #543, opened 2020-03-27
fix for member not redirecting to dir search once tos is accepted

nitheesh-aot:
*Description of changes:*
fix for member not redirecting to dir search once tos is accepted

*Checklist:*
- [ ] No lint errors


auth-web/src/components/auth/mixins/NextPageMixin.vue
```diff
@@ -49,7 +49,9 @@ export default class NextPageMixin extends Vue {
   protected getNextPageUrl (): string {
     switch (this.currentUser?.loginSource) {
       case LoginSource.IDIR:
-        return `/${Pages.SEARCH_BUSINESS}`
+        let idirNextStep = '/'
```
sutherlanda: Don't this change is necessary? Maybe leftover from something?

auth-web/src/views/auth/TermsOfServiceView.vue
```diff
@@ -89,6 +89,10 @@ export default class TermsOfServiceView extends Mixins(NextPageMixin) {
     this.atBottom = (e.target.scrollHeight - e.target.scrollTop) <= (e.target.offsetHeight + 25)
   }
 
+  mounted () {
+    this.$store.commit('updateHeader')
```
sutherlanda: Shouldn't need to do this, is something not updating?

auth-web/src/components/auth/mixins/NextPageMixin.vue
nitheesh-aot: I will change that

auth-web/src/views/auth/TermsOfServiceView.vue
nitheesh-aot: Its needed, the account was not updating without it

sutherlanda: Okay - I may try to find a better way to do this.  Forcing a header remount is expensive.

PR #547, opened 2020-03-29
changes to support transactions in bulk endpoint

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/3074

*Description of changes:*

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-api/src/auth_api/services/keycloak.py
+++ b/auth-api/src/auth_api/services/keycloak.py
@@ -30,7 +30,7 @@ class KeycloakService:
     """For Keycloak services."""
 
     @staticmethod
-    def add_user(user: KeycloakUser, return_if_exists: bool = False):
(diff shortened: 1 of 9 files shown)


auth-api/src/auth_api/services/keycloak.py
```diff
@@ -30,7 +30,7 @@ class KeycloakService:
     """For Keycloak services."""
 
     @staticmethod
-    def add_user(user: KeycloakUser, return_if_exists: bool = False):
+    def add_user(user: KeycloakUser, return_if_exists: bool = False, throw_error_if_exists: bool = False):
```
sutherlanda: Is there a case where we don't want to do anything if the user exists? (i.e. if both flags are False?)

If not, then these flags could be reduced to a singe flag.

PR #552, opened 2020-03-31
[entity2693] UI Clean-up

jeznorth:
*Issue #:*
https://github.com/bcgov/entity/issues/entity2693

Cleaning up the user-interface to conform to design specifications.

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/views/auth/TermsOfServiceView.vue
+++ b/auth-web/src/views/auth/TermsOfServiceView.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-container class="view-container">
+  <v-container class="terms-of-use-container view-container">
 
     <!-- Loading status -->
     <v-fade-transition>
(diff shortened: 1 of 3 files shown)


auth-web/src/views/auth/TermsOfServiceView.vue
```diff
@@ -1,52 +1,41 @@
...
+      <v-card class="profile-card" flat>
+        <v-card-text>
+          <terms-of-use></terms-of-use>
+        </v-card-text>
+        <v-card-actions class="terms-of-use-btns justify-center pt-0 pb-9">
+          <v-btn
+            large
+            color="primary"
+            class="font-weight-bold"
+            @click="clickAccepted"
```
nitheesh-aot: it seems like the disable property is removed here, should we have to enable the accept button even without reaching end of the line?

sutherlanda: I think the buttons will only be shown when the terms text has been scrolled (they are now placed in the same container, and we don't define a separate scroll area).  Correct me if I'm wrong @jeznorth

jeznorth: This is a full page scroll now. Do we need this anymore? By default they need to scroll down to actually see the buttons.

PR #554, opened 2020-04-01
User accepts the invite and inviter deactivates the account#3193 and hiding non accessible tabs for member#3278

nitheesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/3193

*Issue #:*
https://github.com/bcgov/entity/issues/3278

*Description of changes:*
- GIVEN I have received an invitation to join an account
- AND I have click on the link on the email, and sent request to join the account and reached the page says "Thanks for joining the account ..." 
- THEN the account owner deny the request from his pending approval list
- WHEN I refresh the "Thanks for joining..." page
- THEN I should see a message says my request has been denied/revoked
- AND I won't be able to join the account

*Checklist:*
- [ ] No lint errors


jeznorth:
@nitheesh-aot this ticket needs some clarification. Please see the description and comment as necessary.

nitheesh-aot:
> @nitheesh-aot this ticket needs some clarification. Please see the description and comment as necessary.

Hi @jeznorth , the scenario is a bit different from what you updated there, I will edit that description accordingly

auth-web/src/views/auth/PendingApprovalView.vue
```diff
@@ -1,26 +1,47 @@
...
+  }
 })
-export default class PendingApprovalView extends Vue { }
+
+export default class PendingApprovalView extends Vue {
+  protected readonly currentMembership!: Member
+  private isDenied: boolean = false
+
+  mounted () {
+    this.isDenied = !(this.currentMembership?.membershipStatus)
```
sutherlanda, line 44: Is `membershipStatus` undefined/null if the inviter has deactivated their account?

nitheesh-aot: Yes, it was a null response when its denied, thats why put the check like this.

sutherlanda: According to the TeamManagement mixin, this status should be `REJECTED`.  Part of this fix should include correcting the reason that is null at this stage.

nitheesh-aot: sorry I saw this now, I will check the backend for that and fix if anything not correct

sutherlanda: no problem! let me know if you need a hand

saravanpa-aot: @sutherlanda ..Was talking to @nitheesh-aot on this.. as of now , membership is returned only for active and pending members..  
Since this change makes sense , I think we can go ahead and return membership status for rejected users as well.. some more changes in API ,than expected it.. @nitheesh-aot is on it.. Thanks for the good review

PR #559, opened 2020-04-03
added missing Role for director search

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/3309

*Description of changes:*

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-api/src/auth_api/services/products.py
+++ b/auth-api/src/auth_api/services/products.py
@@ -53,7 +53,7 @@ class Product:
                 subscriptions_model_list.append(product_subscription)
             else:
                 raise BusinessException(Error.DATA_NOT_FOUND, None)
-            if subscription.get('product_roles') is not None:
+            if subscription.get('productRoles') is not None:
                 for role in subscription.get('productRoles'):
                     product_role_code = ProductRoleCodeModel.find_by_code_and_product_code(role, product_code)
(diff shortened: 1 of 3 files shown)


auth-api/src/auth_api/services/products.py
```diff
@@ -53,7 +53,7 @@ def create_product_subscription(org_id, subscription_data: Tuple[Dict[str, Any]]
                 subscriptions_model_list.append(product_subscription)
             else:
                 raise BusinessException(Error.DATA_NOT_FOUND, None)
-            if subscription.get('product_roles') is not None:
+            if subscription.get('productRoles') is not None:
```
sumesh-aot, line 56: Can we give all roles for that product by default? Else UI will need to query all the roles and then pass it.

saravanpa-aot: u mean if nothing is passed by default ? The only reason i put it is if we ever put the roles in UI as well..a role selection as well

saravanpa-aot: I added the new logic..Is that good ?
If no roles are passed from UI for a product , all roles gets added..
If the logic is okay , will add more tests..lmk

saravanpa-aot:
I will add the tests in the next PR..merging this

PR #563, opened 2020-04-04
Bugfix director search

nitheesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/2628
https://github.com/bcgov/entity/issues/2694

*Description of changes:*
bugfixes:
- An owner can add 0 team members which is incorrect- ideally this add button should be disabled initially if there is no member added
- Member can also see the team member screen - which is incorrect ( I changed the role of existing admin to member)
- owner can navigate to Manage businesses tab which is incorrect
- on logout - user should be redirected to the Director Search Login page
- Cancel button is not working for create user profile
- hide "Create a new BC Registries account" button in home page

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/components/auth/CreateUserProfileForm.vue
+++ b/auth-web/src/components/auth/CreateUserProfileForm.vue
@@ -19,7 +19,7 @@
               req
               persistent-hint
               :rules="usernameRules"
-              v-model="username"
+              v-model.trim="username"
               data-test="username"
               :disabled="isLoading"
           >
@@ -133,7 +133,7 @@ export default class CreateUserProfileForm extends Mixins(NextPageMixin) {
     }
 
     private usernameRules = [
-      v => !!v || 'Username is required'
+      v => !!v.trim() || 'Username is required'
     ]
 
     private passwordRules = [
@@ -157,7 +157,7 @@ export default class CreateUserProfileForm extends Mixins(NextPageMixin) {
       if (this.isFormValid()) {
         this.isLoading = true
         const requestBody: UserProfileRequestBody = {
-          username: this.username,
+          username: this.username.trim().toLowerCase(),
           password: this.password
         }
         try {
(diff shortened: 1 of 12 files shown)


auth-web/src/components/auth/CreateUserProfileForm.vue
```diff
@@ -191,7 +191,7 @@ export default class CreateUserProfileForm extends Mixins(NextPageMixin) {
     }
 
     private cancel () {
-      window.history.back()
+      this.$router.push('/signin/bcros/')
```
saravanpa-aot, line 194: where does it navigate if the user is signed in?

auth-web/src/router.ts
```diff
@@ -109,7 +111,7 @@ export function getRoutes (): RouteConfig[] {
     { path: '/returnpayment/:paymentId/transaction/:transactionId', name: 'returnpayment', component: PaymentReturnView, props: mapReturnPayVars, meta: { requiresAuth: false, requiresProfile: true } },
     { path: '/searchbusiness', name: 'searchbusiness', component: SearchBusinessView, props: true, meta: { requiresAuth: true, allowedRoles: [Role.Staff] } },
     { path: '/unauthorized', name: 'unauthorized', component: UnauthorizedView, props: true, meta: { requiresAuth: false } },
-    { path: '/unauthorizedtermsdecline', name: 'unauthorizedtermsdecline', component: UnauthorizedView, props: true, meta: { requiresAuth: true } },
+    { path: '/unauthorizedtermsdecline', name: 'unauthorizedtermsdecline', component: TermsOfServiceDeclineView, props: true, meta: { requiresAuth: true } },
```
saravanpa-aot, line 114: I hope this view is not used anywhere else

nitheesh-aot: nope, this view is only for the terms decline view, earlier I reused Unauthorized page conditionally for this, but it seems like there are a lot of different conditions needed in that, thats why moved to a new view page

auth-web/src/components/auth/CreateUserProfileForm.vue
nitheesh-aot: we can navigate to the home page if any other user is already logged in, also will that case happen in the real users?

PR #599, opened 2020-04-16
2946 update notification bell on switch

sutherlanda:
*Issue #:* 2946
https://github.com/bcgov/entity/issues/2946

*Description of changes:*

Force update of header after account switch.  Ensures notification bell is updated.

*Checklist:*
- [x] No lint errors


sutherlanda:
> If it works , all good..
> in quick look , this method completeAccountSwitch gets triggered from an event from header.. and we invoke the header again from this method.. is that correct?

That's correct.  It's a bit inefficient to do it this way, but I don't see another way at the moment.  It does work now though.

PR #712, opened 2020-05-26
Update Manage Business page - Feature flag and Remaining fixes

nitheesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/3651
https://github.com/bcgov/entity/issues/2689

*Description of changes:*

*Checklist:*
- [ ] No lint errors


auth-web/src/store/modules/business.ts
```diff
@@ -58,7 +58,39 @@ export default class BusinessModule extends VuexModule {
...
+          businessIdentifier: businessNumber,
+          name: businessResponse.data.business.legalName
+        })
+        if (updateBusinessResponse?.status === 200) {
+          return updateBusinessResponse
+        } else {
+          isUpdateFailed = true
+        }
+      } else {
+        isUpdateFailed = true
```
saravanpa-aot, line 80: feels like we don't have to set isUpdateFailed ...since you are returning when the status is 200 , all other cases we can delete the orphan records rite?

auth-web/src/store/modules/business.ts
```diff
@@ -58,7 +58,39 @@ export default class BusinessModule extends VuexModule {
...
+      }
+    } catch (error) {
+      isUpdateFailed = true
+    }
+    if (isUpdateFailed) {
```
saravanpa-aot, line 85: I mean its perfectly okay to use the isUpdateFailed variable..its so much readable..but do we have to set this at all places ?

auth-web/src/components/auth/AddNameRequestForm.vue
```diff
@@ -156,8 +156,8 @@ export default class AddNameRequestForm extends Vue {
       (!!this.applicantPhoneNumber || !!this.applicantEmail)
   }
 
-  private isInputEntered (value: any, ipType: string) {
-    return (!!((ipType === 'email') ? this.applicantPhoneNumber : this.applicantEmail) || !!value)
+  private isInputEntered (value: any, inputType: string) {
+    return (!!((inputType === 'email') ? this.applicantPhoneNumber : this.applicantEmail) || !!value)
```
saravanpa-aot, line 160: thanks for doing this.. makes it easy to grasp

auth-web/src/store/modules/business.ts
nitheesh-aot: got it, I will throw the error and do that in the catch, so that the variable can be eliminated

PR #722, opened 2020-05-27
couple of misc bug fixes related to role changes

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/3708

*Description of changes:*

packagelock updated
https://github.com/bcgov/entity/issues/3708
user was not able to see the active member list.fixed it
added a displayorder column so that list view order can be decided

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/components/auth/MemberDataTable.vue
+++ b/auth-web/src/components/auth/MemberDataTable.vue
@@ -188,7 +188,8 @@ export default class MemberDataTable extends Vue {
       case MembershipType.Admin:
         return true
       case MembershipType.Coordinator:
-        if (role.name !== 'Admin') {
+        // coordinator cant upgrade/change themselves to admin
+        if (role.name !== MembershipType.Admin) {
           return true
         }
         return false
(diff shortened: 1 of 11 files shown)


auth-web/src/components/auth/MemberDataTable.vue
```diff
@@ -214,8 +215,8 @@ export default class MemberDataTable extends Vue {
         }
         return false
       case MembershipType.Coordinator:
-        // Admins can change roles of their own
-        return this.isOwnMembership(memberBeingChanged)
+        // Admins can change roles of their own and users as well
+        return this.isOwnMembership(memberBeingChanged) || (memberBeingChanged.membershipTypeCode !== MembershipType.Admin)
```
nitheesh-aot, line 219: This check may fail. since its an 'or', it will allow the admin to change the other Coordinator as well, right?
also, the comment need to be updated to new role

saravanpa-aot: plz check now..I fixed it

PR #733, opened 2020-05-29
#2994 Launch Page Unit Tests

cameron-freshworks:
*Issue #:*
https://github.com/bcgov/entity/issues/2994

*Description of changes:*

* Implements Unit testing for additional Views and components.

* Fixed some styling.

* Fixed some duplicate id's.

* Removed some unused imports.

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/components/auth/BcscPanel.vue
+++ b/auth-web/src/components/auth/BcscPanel.vue
@@ -13,7 +13,7 @@
       <!-- Info Column -->
       <v-col cols="12" md="6" class="bcsc-info-col">
         <h2>It's Secure</h2>
-        <v-list-item class="list-item" v-for="(item, index) in secureBulletPoints" :key="index" >
+        <v-list-item class="list-item" v-for="(item, index) in secureBulletPoints" :key="`List-1-${index}`" >
           <v-icon size="8" class="list-item-bullet mt-5">mdi-square</v-icon>
           <v-list-item-content>
             <v-list-item-subtitle class="list-item-text">
@@ -22,7 +22,7 @@
           </v-list-item-content>
         </v-list-item>
         <h2>It's Quick and Easy</h2>
-        <v-list-item class="list-item" v-for="(item, index) in easeBulletPoints" :key="index" >
+        <v-list-item class="list-item" v-for="(item, index) in easeBulletPoints" :key="`List-2-${index}`" >
           <v-icon size="8" class="list-item-bullet mt-5">mdi-square</v-icon>
           <v-list-item-content>
             <v-list-item-subtitle class="list-item-text">
--- a/auth-web/tests/unit/components/BcscPanel.spec.ts
+++ b/auth-web/tests/unit/components/BcscPanel.spec.ts
@@ -1,12 +1,9 @@
-// Libraries
-import Vue from 'vue'
-import Vuex from 'vuex'
-import Vuetify from 'vuetify'
 import { createLocalVue, mount } from '@vue/test-utils'
-
-// Components
 import BcscPanel from '@/components/auth/BcscPanel.vue'
 import LearnMoreButton from '@/components/auth/common/LearnMoreButton.vue'
+import Vue from 'vue'
+import Vuetify from 'vuetify'
+import Vuex from 'vuex'
 
 Vue.use(Vuetify)
 const vuetify = new Vuetify({})
@@ -31,7 +28,7 @@ describe('MaintainBusinessView.vue', () => {
         vuetify,
         propsData: {
           ...propsData
-        },
+        }
       })
     }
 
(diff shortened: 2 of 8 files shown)


auth-web/tests/unit/views/HomeView.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+    })
+
+    wrapper = mount(HomeViewDev, {
+      store,
+      localVue,
+      router,
+      vuetify,
+      mocks: {
+        $t: (mock) => mock
```
severinbeauvais, line 67: If I understand correctly, this means there's a function $t which, given a value, returns that value. Why is this needed?

auth-web/tests/unit/views/HomeView.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+import Vuex from 'vuex'
+
+Vue.use(Vuetify)
+Vue.use(VueRouter)
+const router = new VueRouter()
+const vuetify = new Vuetify({})
+
+// Prevent the warning "[Vuetify] Unable to locate target [data-app]"
+document.body.setAttribute('data-app', 'true')
+jest.mock('../../../src/plugins/i18n', () => {})
```
severinbeauvais, line 19: Would `@` work here to replace "../../../src" ? Wondering if there's a specific requirement for this in the tests.

auth-web/tests/unit/views/HomeView.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    expect(wrapper.find(BcscPanel).exists()).toBe(true)
+  })
+
+  it('renders the correct buttons when authenticated', () => {
+    const bannerBtns = wrapper.vm.$el.querySelectorAll('.cta-btn')
+    const nameRequestBtn = bannerBtns[0]
+    const namedCompBtn = bannerBtns[1]
+    const numberedCompBtn = bannerBtns[2]
+    const manageBusinessBtn = bannerBtns[3]
+    const loginBtn = bannerBtns[4]
```
severinbeauvais, line 91: Oh, I like this way of naming the buttons up front! Makes the 'expect' code much easier to read :)

auth-web/src/components/auth/BcscPanel.vue
```diff
@@ -13,7 +13,7 @@
       <!-- Info Column -->
       <v-col cols="12" md="6" class="bcsc-info-col">
         <h2>It's Secure</h2>
-        <v-list-item class="list-item" v-for="(item, index) in secureBulletPoints" :key="index" >
+        <v-list-item class="list-item" v-for="(item, index) in secureBulletPoints" :key="`List-1-${index}`" >
```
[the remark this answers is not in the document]
cameron-freshworks: oh nice! This is good to know, thanks 👍

auth-web/tests/unit/components/BcscPanel.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+        vuetify,
+        propsData: {
+          ...propsData
+        }
+      })
+    }
+
+    wrapper = wrapperFactory({ userProfile: {} })
+
+    jest.resetModules()
```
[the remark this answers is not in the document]
cameron-freshworks: I can certainly echo that, sometimes what works in one test suite isn't always as expected in another. 
In this case, with the way auth-web has establish their store into modules, it seems to do the trick effectively. I will test this out tho!

auth-web/tests/unit/views/HomeView.spec.ts
cameron-freshworks: It just defines (mocks i should say) `$t` as a global method, otherwise when it tries to render `$t` it has no reference and will return `"TypeError: _vm.$t is not a function"`.

auth-web/tests/unit/views/HomeView.spec.ts
cameron-freshworks: Yes, good point! `@` would certainly work here. 

To be honest, i'm not sure 100% on this mock. I would have expected this mock to way for the mounted component to understand the global methods but it didn't work as how i expected. Let me do a little more research on this mock.

auth-web/tests/unit/views/HomeView.spec.ts
cameron-freshworks: awesome, thanks!

auth-web/tests/unit/views/HomeView.spec.ts
cameron-freshworks: > Great job Cam!

lol thanks! This message popped up about 7 times :) Really appreciate it heheh

auth-web/tests/unit/components/BcscPanel.spec.ts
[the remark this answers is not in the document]
cameron-freshworks: refactored in next commit. implemented in `afterEach()` 👍

auth-web/tests/unit/views/HomeView.spec.ts
cameron-freshworks: So, this wasn't required. I think? the intention was to give definition to such things such as the global methods ( ie $t ), but it didn't quite mock it as expected. Which is why i opted for the approach from above. Removed this in next commit!

PR #784, opened 2020-06-16
#4039 Launch UI Additions / Fixes

cameron-freshworks:
*Issue #:*
https://github.com/bcgov/entity/issues/4039

*Description of changes:*
* Also Tackles issues in #4017  & #4036

* Various style and performance updates to the launch page.

* Updated method of feature flagging to allow the LD key to be retrieved before loading the proper Home component.

* Updated Stepper for UX and added a small transition to the sub router view.

* Updated the NRO Urls to point to the respective environment your currently in. Required some config mapping in OS ( Saravan has already done i believe )

* Updated styling for Bcsc Panel.

* Updated unit test.

*Checklist:*
- [x] No lint errors


auth-web/public/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
-    <title>Cooperatives Online</title>
+    <title>Business Registry</title>
```
severinbeauvais, line 7: Not "BC Business Registry" ?

auth-web/src/assets/scss/base.scss
```diff
@@ -131,6 +131,10 @@ legend {
   }
 }
 
+.v-btn:hover {
+  opacity: .8;
```
severinbeauvais, line 135: I'm a bit concerned about this style globally, as it will affect all buttons in this UI, and isn't consistent with Filings UI or Create UI.
Is there a special case you're targeting, and could that be targeted more precisely?

auth-web/src/components/auth/BcscPanel.vue
```diff
@@ -56,13 +64,14 @@ import { User } from '@/models/user'
   }
 })
 export default class BcscPanel extends Vue {
+  private cardSetUpUrl = 'https://www2.gov.bc.ca/gov/content/governments/government-id/bc-services-card/log-in-with-card/mobile-card/set-up-mobile-card'
+  private learnMoreUrl = 'https://www2.gov.bc.ca/gov/content/governments/government-id/bc-services-card/log-in-with-card/mobile-card'
```
severinbeauvais, line 68: If you feel REALLY type-y, these could even be `private readonly ...`

auth-web/src/components/auth/TestimonialQuotes.vue
```diff
@@ -88,6 +80,20 @@ export default class TestimonialQuotes extends Vue {
     .fading {
       animation: fading 5s ease-in-out infinite;
     }
+
+    .quote-text {
+      max-width: 45rem;
+      color: #FFFFFF;
```
severinbeauvais, line 86: Aka "white" ? lol

auth-web/src/components/auth/LoginBCSC.vue
```diff
@@ -3,7 +3,7 @@
     <v-card-title>Create a BC Registries Account</v-card-title>
     <v-card-text>
       <p class="mb-7">To access this service, you must securely log in with your BC Services Card, government’s trusted way to access online services. It typically takes five minutes or less to set up your mobile card, and the only information BC Registries can access from your card is your legal name.</p>
-      <a href="https://www2.gov.bc.ca/gov/content/governments/government-id/bc-services-card/login-with-card" rel="noopener noreferrer"
+      <a href="https://www2.gov.bc.ca/gov/content/governments/government-id/bc-services-card/log-in-with-card" rel="noopener noreferrer"
```
severinbeauvais, line 6: I prefer the way you had properties for the URL in the other component. It keeps the template a bit cleaner.

auth-web/public/config/configuration.json
```diff
@@ -1,14 +1,14 @@
...
   "VUE_APP_STATUS_ROOT_API": "https://status-api-dev.pathfinder.gov.bc.ca/api/v1",
   "VUE_APP_PATH_NEW_BUSINESS": "https://business-create-dev.pathfinder.gov.bc.ca/businesses/",
   "DIRECTOR_SEARCH_URL": "https://search-web-dev-1rdehl-dev.pathfinder.gov.bc.ca/",
   "VUE_APP_AUTH_RESET_API": "http://auth-api-dev.pathfinder.gov.bc.ca/test/reset",
   "LAUNCH_DARKLY_ENV_KEY": "5db9da115f58e008123cd783",
   "BCEID_OSD_LINK": "https://www.test.bceid.ca/os/?7686&SkipTo=Basic",
-  "FILE_SERVER_URL":"http://minio-1rdehl-dev.pathfinder.gov.bc.ca/public"
-
+  "FILE_SERVER_URL":"http://minio-1rdehl-dev.pathfinder.gov.bc.ca/public",
+  "NRO_URL": "https://dev.bcregistrynames.gov.bc.ca/nro/"
```
severinbeauvais, line 13: Does this project have dc scripts for rebuilding the Openshift config? If so then this should be updated there as well.
And also, a note about adding this new configmap key needs to be in their release ticket.

auth-web/src/components/auth/common/ContactUsTooltip.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <v-tooltip top max-width="450px" light content-class="tooltip" v-model="showTip">
```
severinbeauvais, line 2: Content class should be a more specific name for this component so that it can be easily selected in the DOM for testing. Unless this is the only tooltip?

auth-web/src/components/auth/common/ContactUsTooltip.vue
```diff
@@ -0,0 +1,59 @@
...
+    <template v-slot:activator="{ attrs }">
+      <span :attrs="attrs" @click="showTip = !showTip" class="tooltip-text">Contact Us</span>
+    </template>
+    <v-card class="tooltip-content">
+      <h3 class="mb-3">Don't have a Passcode for your Cooperative Association?</h3>
+      <p>If you have no received your Access Letter from BC Registries, or have lost your Passcode, please contact us
```
severinbeauvais, line 8: no -> not

auth-web/src/components/auth/common/ContactUsTooltip.vue
```diff
@@ -0,0 +1,59 @@
...
+        at:</p>
+      <section>
+        <p class="mb-0">Toll Free: 1-877-526-1526</p>
+        <p class="mb-0">Phone: 250-387-7848</p>
```
severinbeauvais, line 12: This is not the same phone number as in Filings UI or Create UI. Which one(s) are incorrect?

Also, you could steal some of the markup from https://github.com/bcgov/business-filings-ui/blob/master/src/components/common/ErrorContact.vue ...

auth-web/src/components/auth/common/ContactUsTooltip.vue
```diff
@@ -0,0 +1,59 @@
...
+        <p>Email: bcregistries@gov.bc.ca</p>
+      </section>
+      <v-spacer></v-spacer>
+      <section>
+        <p class="mb-0">Hours of Operation:</p>
+        <p>Monday to Friday, 8:30am - 4:30pm PST</p>
+      </section>
+      <a href="">Frequently Asked Questions</a>
```
severinbeauvais, line 20: Can't go MVP with this.

auth-web/src/components/auth/common/LearnMoreButton.vue
```diff
@@ -21,4 +23,13 @@ export default class LearnMoreButton extends Vue {}
     font-weight: bold;
     width: 160px;
   }
+
+  .btn-learn-more:hover {
+    background-color: $BCgovBlue5;
+    opacity: .8;
```
severinbeauvais, line 29: Again with the opacity. Why is this different from the other UIs?

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
```diff
@@ -1,11 +1,13 @@
 <template>
   <v-tooltip top max-width="450px" light content-class="tooltip">
     <template v-slot:activator="{ on }">
-      <span v-on="on" class="tooltip-text"><u>numbered Benefit Company</u></span>
+      <span v-on="on" class="tooltip-text">numbered Benefit Company</span>
```
severinbeauvais, line 4: numbered -> Numbered?

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
```diff
@@ -1,11 +1,13 @@
...
     </template>
     <v-card class="tooltip-content">
       <h3 class="mb-3">Numbered Benefit Company</h3>
-      <span>Lorem ipsum... some definition of a numbered benefit company would go into this slot here</span>
+      <span>A Benefit Company can choose to use as its name, the incorporation number of the company followed by “B.C.
```
severinbeauvais, line 8: Remove comma.

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
```diff
@@ -1,11 +1,13 @@
...
+        Ltd.” The incorporation number is assigned by the Business Registry after the Incorporation Application is filed
```
severinbeauvais, line 9: Ltd. -> LTD. ?

auth-web/src/components/auth/stepper/InfoStepper.vue
```diff
@@ -2,24 +2,26 @@
...
             active-class="filled no-active"
+            :ripple="false"
             :id=step.id
             color="#1A5A96"
-            :to=step.to>
+            :to="step.to">
             <v-icon>{{step.step}}</v-icon>
           </v-btn>
-          <v-btn class="step__label" active-class="selected no-active" text color="#1A5A96" :to=step.to>
+          <v-btn class="step__label" :ripple="false" active-class="selected no-active" text color="#1A5A96" :to="step.to">
```
severinbeauvais, line 17: I'm not certain whether this works with v-btns, but in HTML if an element doesn't handle an event (eg, click) then it bubbles up to its parent. So the action on this button might not be needed. Or maybe it doesn't need to be a button at all.
(Bubbling is why we stop event propagation on some buttons.)

auth-web/src/views/auth/RequestNameView.vue
```diff
@@ -56,6 +61,9 @@ import NumberedCompanyTooltip from '@/components/auth/common/NumberedCompanyTool
   }
 })
 export default class RequestNameView extends Vue {
+  private nroUrl = ConfigHelper.getNroUrl()
+  private learnMoreUrl = 'https://www2.gov.bc.ca/gov/content/employment-business/business/managing-a-business/' +
```
nitheesh-aot, line 65: is this string concatenation needed here?, ok to be a single string right?

auth-web/src/routes/router.ts
```diff
@@ -47,6 +47,13 @@ function mapReturnPayVars (route: any) {
   }
 }
 
+// Feature Flagging HomeView Components
+// For Development only
```
severinbeauvais, line 51: Not for development only. Actually the opposite :)
But definitely temporary while leading up to MVP.

auth-web/src/views/auth/MaintainBusinessView.vue
```diff
@@ -26,35 +26,45 @@
...
 import { User } from '@/models/user'
 
 @Component({
   components: {
+    ContactUsTooltip,
     LearnMoreButton
   }
 })
 export default class MaintainBusinessView extends Vue {
+  private learnMoreUrl = 'https://www2.gov.bc.ca/assets/gov/employment-business-and-economic-development/business' +
```
nitheesh-aot, line 66: same here, ok to be a single string right?

auth-web/src/views/auth/HomeViewDev.vue
```diff
@@ -69,10 +69,13 @@
         <v-container>
           <h2>How does it work?</h2>
           <InfoStepper />
-          <RouterView
-            :userProfile="userProfile"
-            @account-dialog="accountDialog = true"
-            @manage-businesses="goToManageBusinesses($event)"/>
+          <transition name="slide-x-transition">
```
severinbeauvais, line 72: Nice 👍

auth-web/src/views/auth/IncorpOrRegisterView.vue
```diff
@@ -37,18 +37,21 @@
...
+          <v-btn large color="#fcba19" @click="emitLogin()" class="mt-5">
             Log in with BC Services Card
           </v-btn>
           <p class="my-5">New to BC Registries? <a @click="emitAccountDialog()" class="create-account-link">
             <u>Create a BC Registries Account</u></a>
           </p>
         </template>
-        <LearnMoreButton />
+        <learn-more-button
+        :redirect-url="'GO SOMEWHERE'"/>
```
severinbeauvais, line 48: Not ready for MVP. Also on line 52.
Also should indent or move to prev line.

auth-web/public/index.html
cameron-freshworks: 'Business Registry ' According to the req doc, but i'll follow up!

auth-web/src/assets/scss/base.scss
cameron-freshworks: Yea, i was on the fence with this. According to the invision style guides. All the buttons in our future state will be .8 op on hover. Government standard apparently. Though, i certainly dont doing this individually, would you prefer that?

auth-web/src/components/auth/BcscPanel.vue
cameron-freshworks: Ah! Will update this for sure.

auth-web/src/components/auth/LoginBCSC.vue
cameron-freshworks: Oops, will update this for consistency!

auth-web/public/config/configuration.json
cameron-freshworks: Yes, i believe Saravan has updated the Config Map but i'm not certain of the build scripts. Will follow up on this!

k, will do 👍

auth-web/src/components/auth/common/ContactUsTooltip.vue
cameron-freshworks: This is only the tool tip. I did consider adding an Prop to the tooltips that will be the ID of the parent, passed into the tooltip so we can attach it where use it. 

I forgot to mention. This ToolTip still had some outstanding information and isn't currently implemented. Hence why there is incomplete data as mentioned below :) This is just something we will need once we have the info regarding what we want for mvp.

auth-web/src/components/auth/common/ContactUsTooltip.vue
cameron-freshworks: Will update, thanks!

auth-web/src/components/auth/common/ContactUsTooltip.vue
cameron-freshworks: Sorry, this data was just placeholder, i forgot to specify this tooltip wasn't implemented yet. Just building the structure so when we have all the info i could update and implement. 
Should i take this out for now?

auth-web/src/components/auth/common/ContactUsTooltip.vue
cameron-freshworks: For sure, same as above. Should i remove the tooltip template until we have all the info for implementation.

auth-web/src/components/auth/common/LearnMoreButton.vue
cameron-freshworks: This is a mistake! Doubled up. Though like before, should we go local on these btn:hovers opposed to global. I'm 100% good with either. Maybe local would be safer..

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
cameron-freshworks: Designs actually show this tooltip lowercase on the `numbered`, perhaps because it's slotted into a sentence. Will follow up on this!

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
cameron-freshworks: Was in the design, but will follow up with Scott :)

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
cameron-freshworks: Lowercase by design, but will follow up on this!

auth-web/src/components/auth/stepper/InfoStepper.vue
cameron-freshworks: The `ripple` is to stop that default animation that the Vuetify `v-btns` have when being selected. 
Or are you referring to the `:to` property?

auth-web/src/routes/router.ts
cameron-freshworks: That is a good point :) My thinking is backwards!

auth-web/src/views/auth/RequestNameView.vue
cameron-freshworks: Definitely, will update this! In the Filings and Create UI's we have a max length on a line for the linters, so i've gotten used to that :)

auth-web/src/views/auth/MaintainBusinessView.vue
cameron-freshworks: Definitely, will update!

auth-web/src/views/auth/HomeViewDev.vue
cameron-freshworks: Yea! Kinda cool, applying a transition to a routerview. Vuetify is pretty solid :)

auth-web/src/views/auth/IncorpOrRegisterView.vue
cameron-freshworks: Sorry, another placeholder 🤦 . This won't make it to MVP 100% but i think to be safe i will put in another help link until the proper one is provided. I think this doc or static page is currently a wip.

auth-web/tests/unit/views/HomeView.spec.ts
```diff
@@ -17,11 +17,16 @@ const vuetify = new Vuetify({})
 // Prevent the warning "[Vuetify] Unable to locate target [data-app]"
 document.body.setAttribute('data-app', 'true')
 
+const mockSession = {
+  'NRO_URL': 'Mock Url'
```
severinbeauvais, line 21: You set this; is it possible to have a test that clicking the button attempts to redirect to it (same as we do in some Filings UI unit tests)?

auth-web/tests/unit/views/HomeView.spec.ts
```diff
@@ -73,7 +78,7 @@ describe('HomeView.vue', () => {
     jest.clearAllMocks()
   })
 
-  it('is a Vue instance', () => {
+  it('is a Vue instance', async () => {
```
severinbeauvais, line 81: Looks like async is not needed here.

auth-web/tests/unit/views/RequestNameView.spec.ts
```diff
@@ -12,10 +12,15 @@ const vuetify = new Vuetify({})
 // Prevent the warning "[Vuetify] Unable to locate target [data-app]"
 document.body.setAttribute('data-app', 'true')
 
+const mockSession = {
+  'NRO_URL': 'Mock Url'
```
severinbeauvais, line 16: Same comment/question as in HomeView about testing this.

auth-web/src/views/auth/RequestNameView.vue
```diff
@@ -16,36 +16,41 @@
...
         <!-- Panel Btns -->
         <div class="request-name-info-btns mt-5">
           <v-btn large color="#003366" class="white--text"
-            href="https://www.bcregistrynames.gov.bc.ca/nro/" target="_blank" rel="noopener noreferrer">
+                 :href="nroUrl" target="_blank" rel="noopener noreferrer">
             Request a Name
           </v-btn>
           <p class="mt-5">Have an existing Name Request?
-            <a href="https://www.bcregistrynames.gov.bc.ca/nro/nro.htm?_flowId=anonymous-monitor-flow&_flowExecutionKey=e1s1"
+            <a :href="`${nroUrl}nro.htm?_flowId=anonymous-monitor-flow&_flowExecutionKey=e1s1`"
```
severinbeauvais, line 31: While this works, I suggest using a getter so it looks cleaner.

The principle I'm trying to follow is to keep "code" out of the template. This instance isn't serious enough to preempt my approval but, you know, it would be cleaner 😀

auth-web/src/views/auth/RequestNameView.vue
```diff
@@ -16,36 +16,41 @@
...
-          <LearnMoreButton />
+          <learn-more-button
+            :redirect-url="learnMoreUrl"
+          />
         </div>
       </v-col>
       <!-- Image Column -->
       <v-col cols="12" md="6">
-        <v-img src="../../assets/img/Step2-Name-Request.png" aspect-ratio="1.2" contain></v-img>
+        <a :href="nroUrl" target="_blank">
```
severinbeauvais, line 43: You know what would be really sexy (if we had time)? A new component type or directive to suppose the DRY principle (since it seems there are lots of anchor links to NR). Just an idea.

auth-web/src/assets/scss/base.scss
severinbeauvais: My first concern is that this base style will affect all buttons, which we may want to do. My second concern is, if this is a standard, then we should be doing it for the other UIs also (but not willy-nilly -- will need regression testing).
I'm unsure of the best approach here... maybe to keep risk low, just apply this style where it's needed? That may result in a bit of tech debt but oh well, MVP is close!

auth-web/src/components/auth/common/ContactUsTooltip.vue
severinbeauvais: Sounds like you're on top of this already.

My 2c: name the class "contact-us-tooltip" ?

auth-web/src/components/auth/common/ContactUsTooltip.vue
severinbeauvais: I'm OK with leaving this code in.

auth-web/src/components/auth/common/ContactUsTooltip.vue
severinbeauvais: Ah, um...

Either way there is tech debt (either to add the file or update it later). I suggest leaving in, possibly with a big comment block at the top that it's incomplete and not used yet. You may avoid future merge issues if you do this instead of keeping this file on another branch to merge later.

auth-web/src/components/auth/common/LearnMoreButton.vue
severinbeauvais: Go safer to preempt full regression testing.

auth-web/src/components/auth/stepper/InfoStepper.vue
severinbeauvais: Yes, I was thinking that this button could "do nothing" and its parent div could handle the bubbled-up event. Not sure if this is possible here, but it's a common technique for divs that work as buttons.

auth-web/src/assets/scss/base.scss
cameron-freshworks: Totally agree :) Updated in next commit!

auth-web/public/index.html
cameron-freshworks: Updated in next commit!

auth-web/src/components/auth/BcscPanel.vue
cameron-freshworks: Updated in next commit!

auth-web/src/components/auth/TestimonialQuotes.vue
cameron-freshworks: Updated!

auth-web/src/components/auth/LoginBCSC.vue
cameron-freshworks: Done in next commit!

auth-web/src/components/auth/common/ContactUsTooltip.vue
cameron-freshworks: Agreed! Done in next commit

auth-web/src/components/auth/common/ContactUsTooltip.vue
cameron-freshworks: Updated in next commit

auth-web/src/components/auth/common/ContactUsTooltip.vue
cameron-freshworks: Done and done, in next commit!

auth-web/src/components/auth/common/LearnMoreButton.vue
cameron-freshworks: Updated in next commit. Removed the duplication!

auth-web/src/components/auth/stepper/InfoStepper.vue
cameron-freshworks: Great question, when i remove the `:to` binding, it removes the btns access to its `active-class` binding. I have no idea why or how, but the result is the underline styling that follows the selected step, doesn't work. It will require some restructuring to figure out, since we totally made this custom :) 
Can we defer this?

auth-web/src/views/auth/RequestNameView.vue
cameron-freshworks: Fixed in next commit!

auth-web/src/views/auth/MaintainBusinessView.vue
cameron-freshworks: Updated in next commit!

auth-web/src/views/auth/IncorpOrRegisterView.vue
cameron-freshworks: Updated in next commit!

auth-web/tests/unit/views/HomeView.spec.ts
cameron-freshworks: Absolutely, for the sake of unblocking the other team, are you ok if i defer expanding the unit tests to a future commit?

auth-web/tests/unit/views/HomeView.spec.ts
cameron-freshworks: Updated in next commit!

auth-web/tests/unit/views/RequestNameView.spec.ts
cameron-freshworks: Totally, its a great idea. Will do this, in a future commit ( If thats ok :) )

auth-web/src/views/auth/RequestNameView.vue
cameron-freshworks: Absolutely agree, Will update this!

auth-web/src/components/auth/stepper/InfoStepper.vue
severinbeauvais: Yes, let's defer this, as it's currently working correct (just not perfect code-wise). It's low priority so I'm not even going to create a ticket for it atm.

auth-web/src/views/auth/RequestNameView.vue
cameron-freshworks: Is a great idea, i was thinking the same after i did this a few times, i realized my folly. Will update this now!

auth-web/tests/unit/views/HomeView.spec.ts
severinbeauvais: Yes, absolutely.

auth-web/src/views/auth/RequestNameView.vue
severinbeauvais: It's an academic exercise. Only do if there is time :)

cameron-freshworks: Created a Name Request Button component and implemented where it's used 👍 
Tho, there was still one instance where i pulled the nroUrl other than the component.

auth-web/src/views/auth/RequestNameView.vue
cameron-freshworks: Updated!

auth-web/src/routes/router.ts
```diff
@@ -324,18 +330,14 @@ export function getRoutes (): RouteConfig[] {
...
 }
 
-// Feature Flagging HomeView Components
-// For Development only
-const flagCondition = LaunchDarklyService.getFlag('incorporations-launch-feature')
-
 // Get the correct Homeview depending on Environment
 const getEnvHomeView = () => {
-  return flagCondition ? HomeViewDev : HomeView
+  return getFeatureFlag ? HomeViewDev : HomeView
```
severinbeauvais, line 335: `getFeatureFlag` is a function not a getter.
This ternary will always return the first response since the function always exists.
Ie, you need () to call the function, not test its existence.

severinbeauvais: Or change the function to a getter.

cameron-freshworks: good catch, thanks. I didn't update this when i refactored!

PR #856, opened 2020-07-07
SBC_COMMON : Redirect the user to create account if the user doesn't have valid role#3296

nitheesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/3296

*Description of changes:*
SBC_COMMON : Redirect the user to create account if the user doesn't have valid role

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/components/auth/NonBcscAccounts/AffidavitDownload.vue
+++ b/auth-web/src/components/auth/NonBcscAccounts/AffidavitDownload.vue
@@ -73,24 +73,40 @@
         Back
       </v-btn>
       <v-spacer></v-spacer>
-      <v-btn
-        large
-        color="primary"
-        class="font-weight-bold mr-2"
-        data-test="use-existing-bceid-btn"
-        @click="signinUsingBCeID"
-      >
-        Login using existing BCeID
-      </v-btn>
-      <v-btn
-        large
-        color="primary"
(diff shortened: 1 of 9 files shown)


auth-web/src/components/auth/NonBcscAccounts/AffidavitDownload.vue
```diff
@@ -73,37 +73,64 @@
...
-      <v-btn
-        large
-        color="primary"
-        class="font-weight-bold"
-        data-test="register-bceid-btn"
-        @click="redirectToBceId"
-      >
-        Register a new BCeID
-      </v-btn>
+      <template v-if="isAuthenticated">
```
saravanpa-aot, line 76: is this change related to this ticket?

saravanpa-aot: the changes looks like if isAuthenticated , then show the signin buttons ? a bit confused here..could be a correct change , just wants to understand it more

nitheesh-aot: Yes, its related to this.
actually if isAuthenticated, then we show the Continue button, otherwise the signin/register buttons

PR #927, opened 2020-07-29
Remove LD flag checks and unwanted codes#4481

nitheesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4481

*Description of changes:*
- Remove LD flag checks and unwanted codes#4481

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/components/auth/EntityManagement.vue
+++ b/auth-web/src/components/auth/EntityManagement.vue
@@ -11,7 +11,7 @@
       <div class="view-header align-center">
         <h1 class="view-header__title">Manage Businesses</h1>
         <div class="view-header__actions">
-          <v-menu v-if="showAddNameRequestDropdown">
+          <v-menu>
             <template v-slot:activator="{ on }">
               <v-btn
                 color="primary"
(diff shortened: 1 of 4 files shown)


auth-web/src/components/auth/EntityManagement.vue
```diff
@@ -43,16 +43,6 @@
               </v-list-item>
             </v-list>
           </v-menu>
-          <v-btn
```
saravanpa-aot, line 46: I dont know the logic here..but hope its fine to remove this code

auth-web/src/routes/router.ts
```diff
@@ -63,9 +63,34 @@ export function getRoutes (): RouteConfig[] {
...
+      component: HomeView,
+      children: [
+        {
+          path: '',
+          redirect: 'decide-business'
+        },
+        {
+          path: 'decide-business',
+          component: DecideBusinessView,
+          meta: { showNavBar: false }
```
cameron-freshworks, line 75: We might not need the `meta` data for the child routes if the `showNavBar` is false on the parent route. No biggie either way though, might be worth a test.

auth-web/src/components/auth/EntityManagement.vue
nitheesh-aot: Yes its fine, as the flag is to toggle the button in the else part

PR #960, opened 2020-08-17
[entity4373] UI Updates to GL Code Interfaces

jeznorth:
*Issue #:*
https://github.com/bcgov/entity/issues/entity4373

UI Updates to General Ledger Code Interfaces

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/components/auth/staff/GLCodeDetailsModal.vue
+++ b/auth-web/src/components/auth/staff/GLCodeDetailsModal.vue
@@ -1,185 +1,211 @@
 <template>
-  <v-row justify="center">
-    <v-dialog
-      v-model="isOpen"
-      persistent
-      max-width="900px"
-    >
-      <v-card class="pa-2">
-        <v-card-title>
-          <h3>General Ledger Code Details</h3>
-          <v-btn
-            large
-            icon
-            @click="close"
-          >
-            <v-icon>mdi-close</v-icon>
-          </v-btn>
-        </v-card-title>
-        <v-card-text>
-          <v-row class="mb-4">
-            <v-col cols="12" sm="4">
-              <div class="gl-code-label">Effective Dates</div>
-            </v-col>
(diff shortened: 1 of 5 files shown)


auth-web/src/components/auth/staff/GLCodeDetailsModal.vue
```diff
@@ -1,185 +1,220 @@
...
+            <v-data-table
+              :headers="filingTypeHeaders"
+              :items="filingTypes"
+              item-key="feeScheduleId"
+              hide-default-footer
+            >
+              <template v-slot:loading>
+                Loading...
+              </template>
+              <template v-slot:item.corpTypeName="{ item }">
```
nitheesh-aot, line 197: We don't need the template here unless we need to format the column specifically.
just make sure the 'value' field in the 'filingTypeHeaders' has the correct variable name to map the data

auth-web/src/components/auth/staff/GLCodeDetailsModal.vue
```diff
@@ -1,185 +1,220 @@
...
+                  {{item.corpType}}
+              </template>
+              <template v-slot:item.filingTypeName="{ item }">
```
nitheesh-aot, line 200: same as the above

auth-web/src/components/auth/staff/GLCodeDetailsModal.vue
```diff
@@ -208,6 +244,21 @@ export default class GLCodeDetailsModal extends Vue {
     }
   }
 
+  private readonly filingTypeHeaders = [
+    {
+      text: 'Corporation Type',
+      align: 'left',
+      sortable: false,
+      value: 'corpTypeName'
```
nitheesh-aot, line 252: 'corpType' instead of 'corpTypeName' can avoid the v-slot template in the html

PR #976, opened 2020-08-28
4780: Update the user profile only during initial login

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4780

*Description of changes:*
- Removing update user profile fro sync operation 

*Checklist:*
- [ ] No lint errors


auth-web/package.json
```diff
@@ -21,7 +21,7 @@
     "moment": "^2.24.0",
     "regenerator-runtime": "^0.13.3",
     "register-service-worker": "^1.6.2",
-    "sbc-common-components": "^2.3.1",
+    "sbc-common-components": "^2.3.4",
```
saravanpa-aot, line 24: u will have to check in the package-lock as well..

auth-web/src/store/modules/user.ts
```diff
@@ -140,7 +140,7 @@ export default class UserModule extends VuexModule {
 
   @Action({ rawError: true })
   public async syncUserProfile () {
-    const userResponse = await UserService.syncUserProfile()
+    const userResponse = await UserService.getUserProfile('@me')
```
saravanpa-aot, line 143: is this correct?

nitheesh-aot: we need to have the syncUserProfile during the initial login, right?
this change might never update the user profile

sumesh-aot: User profile update happens from common components, not from auth-web. The change in common components is to make sure user profile is updated regardless of which app the user logs in.

sumesh-aot: There were lot of calls to syncUserProfile and was causing unwanted update to user profile on backend.

auth-web/package.json
sumesh-aot: Added package-lock

auth-web/src/store/modules/user.ts
nitheesh-aot: oh, you added the post call there. I wasn't aware of that.
That will be good.
I am approving this.

PR #1043, opened 2020-10-01
4772 add view for PAD tos + dialog

sutherlanda:
*Issue #:* 4772
https://github.com/bcgov/entity/issues/4772

*Description of changes:*

Adds views for the PAD agreement TOS dialog.  Before this feature can be used, the Payment Selection feature must be completed.

NOTE: This also does not include the data model or API changes needed on the account model to determine if the TOS has already been accepted.  Placeholders have been put in where appropriate.

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/components/auth/PAD/PADAgreementDialog.vue
+++ b/auth-web/src/components/auth/PAD/PADAgreementDialog.vue
@@ -1,0 +1,178 @@
+<template>
+  <v-row>
+    <v-col sm="12">
+      <v-tooltip bottom>
+        <template v-slot:activator="{ on }">
+          <v-checkbox
+            v-on:change="emitStatus()"
+            v-on="on"
+            class="terms-checkbox"
+            color="default"
+            v-model="termsAccepted"
+            required
+            :disabled="termsAccepted || !canCheckTerms"
+          >
+            <template v-slot:label>
+              <div class="terms-checkbox-label" v-on="on">
+                <span>I have read and agreed to the</span>
+                <v-btn
+                  text
+                  link
+                  color="primary"
+                  class="pr-1 pl-1"
+                  @click.stop="openDialog()"
+                  data-test="terms-of-use-checkbox"
+                >Terms of Use</v-btn>
+              </div>
+            </template>
+          </v-checkbox>
+        </template>
+        <span>{{tooltipTxt}}</span>
+      </v-tooltip>
+      <v-dialog scrollable width="1024" v-model="termsDialog" :persistent="true">
+        <v-card>
+          <v-card-title>Terms of Use</v-card-title>
+          <v-card-text id="scroll-target" data-test="scroll-area">
+            <div v-scroll:#scroll-target="onScroll" style="height: 2000px;">
+              <PADTermsOfUse
+                :content="content"
+              ></PADTermsOfUse>
+            </div>
+          </v-card-text>
+          <v-card-actions>
+            <v-btn
+              large
+              depressed
+              color="primary"
+              class="agree-btn"
+              :disabled="!atBottom"
+              @click="termsDialog = false; termsAccepted = true; emitStatus()"
+              data-test="accept-button"
+            >
+              <span>Agree to Terms</span>
+            </v-btn>
+            <v-btn
+              large
+              depressed
+              color="primary"
+              class="agree-btn"
+              @click="termsDialog = false"
+              data-test="close-button"
+            >
+              <span>Close</span>
+            </v-btn>
+          </v-card-actions>
+        </v-card>
+      </v-dialog>
+    </v-col>
+  </v-row>
+</template>
+
+<script lang="ts">
+import { Component, Emit, Prop, Vue, Watch } from 'vue-property-decorator'
+import PADTermsOfUse from '@/components/auth/PAD/PADTermsOfUse.vue'
+import { User } from '@/models/user'
+import documentService from '@/services/document.services.ts'
+import { getModule } from 'vuex-module-decorators'
+import { mapState } from 'vuex'
+
+@Component({
+  components: {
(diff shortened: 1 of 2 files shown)


nitheesh-aot:
Can we reuse the existing TOS components for this? not sure whether we need a new component since the behaviour is the same.

nitheesh-aot (commented):
Can we reuse the existing TOS components for this? not sure whether we need a new component since the behaviour is the same.

saravanpa-aot (commented):
@sutherlanda i think the api will be same like other component ;just with a different type of document..so we can reuse the existing component rite?  I guess the designs , behaviour etc will be similar to other TOS component

sutherlanda:
@nitheesh-aot @saravanpa-aot would it make more sense to add configuration to the existing user TOS to support retrieval of alternative documents?  If you agree, I can make this change on this PR instead of adding a new component.

sutherlanda:
@saravanpa-aot @nitheesh-aot Making the User TOS configurable is proving to be quite complex because the business logic is split across two different components: `TermsOfUseDialog` and `TermsOfUse` and the PAD terms are going to need to be stored on the Account data model, rather than the User Profile.  Unless you are planning to have many different versions of the TOS, it may be simpler to have a bit of duplication (as I've done in this PR) rather than try to create a complex configuration for the TOS components.  Let me know your thoughts.  The TOS components could also be simplified, but that is probably beyond the scope of this ticket.

saravanpa-aot:
> @saravanpa-aot @nitheesh-aot Making the User TOS configurable is proving to be quite complex because the business logic is split across two different components: `TermsOfUseDialog` and `TermsOfUse` and the PAD terms are going to need to be stored on the Account data model, rather than the User Profile. Unless you are planning to have many different versions of the TOS, it may be simpler to have a bit of duplication (as I've done in this PR) rather than try to create a complex configuration for the TOS components. Let me know your thoughts. The TOS components could also be simplified, but that is probably beyond the scope of this ticket.

Hm..yeah..I agree that we cant re-use the TermsOfUseDialog.For sure. Since its the page which hosts the component , we have to create new one.
was hoping we can reuse the core component **TermsOfUse**. May be we have to pass in a prop and identify the type of the document and pass it to backend.Would be great if we can do that since it has logic around hasAcceptedLatestTos.

And yeah; the TOS has different versions. We get frequent updates to TOS ; new version comes in often.3 versions for normal users so far. This will be the case with PAD agreement as well I am guessing.We might have to even stop the user from proceeding if they have accepted the older version and new version in place. This was the case with normal TOS.

Hm..good thinking about where the data goes. we moved all the payment related data to pay-db .Not even payment method is in auth-db. The single source of truth is pay-db now.  So have to think about it. will get back to you on that.

sutherlanda:
> > @saravanpa-aot @nitheesh-aot Making the User TOS configurable is proving to be quite complex because the business logic is split across two different components: `TermsOfUseDialog` and `TermsOfUse` and the PAD terms are going to need to be stored on the Account data model, rather than the User Profile. Unless you are planning to have many different versions of the TOS, it may be simpler to have a bit of duplication (as I've done in this PR) rather than try to create a complex configuration for the TOS components. Let me know your thoughts. The TOS components could also be simplified, but that is probably beyond the scope of this ticket.
> 
> Hm..yeah..I agree that we cant re-use the TermsOfUseDialog.For sure. Since its the page which hosts the component , we have to create new one.
> was hoping we can reuse the core component **TermsOfUse**. May be we have to pass in a prop and identify the type of the document and pass it to backend.Would be great if we can do that since it has logic around hasAcceptedLatestTos.
> 
> And yeah; the TOS has different versions. We get frequent updates to TOS ; new version comes in often.3 versions for normal users so far. This will be the case with PAD agreement as well I am guessing.We might have to even stop the user from proceeding if they have accepted the older version and new version in place. This was the case with normal TOS.
> 
> Hm..good thinking about where the data goes. we moved all the payment related data to pay-db .Not even payment method is in auth-db. The single source of truth is pay-db now. So have to think about it. will get back to you on that.

@saravanpa-aot what's the final decision on this? Do you want me to refactor or leave as is?

saravanpa-aot:
> > > @saravanpa-aot @nitheesh-aot Making the User TOS configurable is proving to be quite complex because the business logic is split across two different components: `TermsOfUseDialog` and `TermsOfUse` and the PAD terms are going to need to be stored on the Account data model, rather than the User Profile. Unless you are planning to have many different versions of the TOS, it may be simpler to have a bit of duplication (as I've done in this PR) rather than try to create a complex configuration for the TOS components. Let me know your thoughts. The TOS components could also be simplified, but that is probably beyond the scope of this ticket.
> > 
> > 
> > Hm..yeah..I agree that we cant re-use the TermsOfUseDialog.For sure. Since its the page which hosts the component , we have to create new one.
> > was hoping we can reuse the core component **TermsOfUse**. May be we have to pass in a prop and identify the type of the document and pass it to backend.Would be great if we can do that since it has logic around hasAcceptedLatestTos.
> > And yeah; the TOS has different versions. We get frequent updates to TOS ; new version comes in often.3 versions for normal users so far. This will be the case with PAD agreement as well I am guessing.We might have to even stop the user from proceeding if they have accepted the older version and new version in place. This was the case with normal TOS.
> > Hm..good thinking about where the data goes. we moved all the payment related data to pay-db .Not even payment method is in auth-db. The single source of truth is pay-db now. So have to think about it. will get back to you on that.
> 
> @saravanpa-aot what's the final decision on this? Do you want me to refactor or leave as is?

I think its good the way it is.. Will see later , if the existing components can be reused. There is a follow up ticket to work on when actual TOS is ready. Merging this now.. Thanks

saravanpa-aot (approved):

PR #1099, opened 2020-10-28
PAD bank account validation for premium accounts#4773

nitheesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4773

*Description of changes:*
- PAD bank account validation for premium accounts#4773

*Checklist:*
- [ ] No lint errors

### DIFF
--- a/auth-web/src/store/modules/org.ts
+++ b/auth-web/src/store/modules/org.ts
@@ -12,6 +12,7 @@ import {
   MembershipType,
   Organization,
   PADInfo,
+  PADInfoValidation,
   UpdateMemberPayload
 } from '@/models/Organization'
 import { BcolAccountDetails, BcolProfile } from '@/models/bcol'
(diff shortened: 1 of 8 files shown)


auth-web/src/store/modules/org.ts
```diff
@@ -350,6 +352,14 @@ export default class OrgModule extends VuexModule {
     return response?.data
   }
 
+  @Action({ rawError: true })
+  public async validatePADInfo (): Promise<PADInfoValidation> {
+    const padInfo: PADInfo = { ...this.context.state['currentOrgPADInfo'] }
+    delete padInfo.isTOSAccepted
+    const response = await PaymentService.verifyPADInfo(padInfo)
```
saravanpa-aot, line 359: no try catch here?

saravanpa-aot: i saw that logic in other place..is it better if its handled here?

saravanpa-aot (commented):
just about the createaccount in the catch..is there a way to move it out of catch?

nitheesh-aot:
> just about the createaccount in the catch..is there a way to move it out of catch?

The there are multiple calls need to be done in the create account, it will be better to call the create account method in the catch in this case

saravanpa-aot (approved):
dicussed with nitheesh and suggested some options

PR #1212, opened 2020-12-16
quick fix for the broken payment update method

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/5064

*Description of changes:*

Just temporary work around to make the update of the payment method work.
The code needs a rewrite.I will probably do it along the next ticket.I tried to rewrite ;but need to touch a lot of tests.So parking it for later.
This is done so that front end dev can test the ticket.

### DIFF
--- a/auth-api/src/auth_api/resources/org.py
+++ b/auth-api/src/auth_api/resources/org.py
@@ -158,7 +158,7 @@ class Org(Resource):
                     response, status = org.change_org_ype(request_json, action,
                                                           bearer_token).as_dict(), http_status.HTTP_200_OK
                 else:
-                    response, status = org.update_org(request_json, bearer_token).as_dict(), http_status.HTTP_200_OK
+                    response, status = org.update_org(request_json, token, bearer_token).as_dict(), http_status.HTTP_200_OK
             else:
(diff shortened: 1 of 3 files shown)


auth-api/src/auth_api/resources/org.py
```diff
@@ -158,7 +158,8 @@ def put(org_id):
                     response, status = org.change_org_ype(request_json, action,
                                                           bearer_token).as_dict(), http_status.HTTP_200_OK
                 else:
-                    response, status = org.update_org(request_json, bearer_token).as_dict(), http_status.HTTP_200_OK
+                    response, status = org.update_org(request_json, token, bearer_token).as_dict(),\
```
sumesh-aot, line 161: trivial : May be rename token to user_info or something? token and bearer token are confusing, and I thought both are same.

sumesh-aot (approved):

auth-api/src/auth_api/resources/org.py
saravanpa-aot: yeah.makes sense

saravanpa-aot: i changed it.

PR #1245, opened 2020-12-24
5721-Update User Terms.

mengdong19:
*Issue #:*
https://github.com/bcgov/entity/issues/5721

*Description of changes:*
Add new version of terms and conditions.
Add PAD terms of use page.


shabeeb-aot (approved):

auth-web/src/routes/router.ts
```diff
@@ -509,6 +510,13 @@ export function getRoutes (): RouteConfig[] {
       component: TermsOfServiceView,
       meta: { requiresAuth: true }
     },
+    {
+      path: '/PAD-terms-and-conditions',
```
saravanpa-aot, line 514: the approach has to be revised so that all the kind of tos documents can be exposed as urls... may be accept the type in the url and rename it as termsofconditions view or soemthing...
since the url wont be used now , good to merge

auth-web/src/views/auth/PADTermsAndConditionsView.vue
```diff
@@ -0,0 +1,43 @@
...
+</script>
+
+<style lang="scss" scoped>
+  .terms-of-use-container {
+    max-width: 65rem;
+  }
+
+  // TODO: Payor Authoriztion section of the PAD Terms does not apply in this content so we don't display it
+  // TODO: Better strategy to show/hide this content VS hacking it out with styling
+  ::v-deep {
```
saravanpa-aot, line 37: this has to be handled in a different way rather than style hide...

saravanpa-aot (approved):

auth-web/src/routes/router.ts
mengdong19: Might need to adjust styling and format to the terms if we do this way.

PR #1250, opened 2021-01-04
[entity6010] Develop Static Price List Interface

jeznorth:
*Issue #:*
https://github.com/bcgov/entity/issues/entity6010

Implement static content / route for initial price list (aka 'Fee Schedule')

### DIFF
--- a/auth-web/src/views/auth/PriceListView.vue
+++ b/auth-web/src/views/auth/PriceListView.vue
@@ -1,0 +1,218 @@
+<template>
+  <main class="view-container container">
+      <h1 class="mb-4">Price List</h1>
+      <p class="mb-8">Where applicable, tax is included in the price shown. Prices are subject to change without notice.</p>
+      <section class="mb-10">
+        <h2 class="mb-5">Business Registry</h2>
+        <v-card flat>
+          <table aria-label="Price List Table">
+            <thead>
+              <th
+                class="product text-left"
+                scope="col"
+              >
+                Product
+              </th>
+              <th
+                class="fee"
+                scope="col"
+              >
+                Price
+              </th>
+              <th
+                class="fee--service"
+                scope="col"
+              >
+                Service Fee
+              </th>
+              <th
+                class="fee--priority"
+                scope="col"
+              >
+                Priority Fee
+              </th>
+              <th
+                class="fee--future"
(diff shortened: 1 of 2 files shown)


auth-web/src/views/auth/PriceListView.vue
```diff
@@ -0,0 +1,181 @@
+<template>
+  <main class="view-container container">
+      <h1 class="mb-4">Price List</h1>
+      <p class="mb-8">Where applicable, tax is included in the price shown. Prices are subject to change without notice.</p>
+      <section class="mb-10">
+        <h2 class="mb-5">Business Registry</h2>
+        <v-card flat>
+          <table>
```
saravanpa-aot, line 8: curious , why not v-data-table?

auth-web/src/views/auth/PriceListView.vue
```diff
@@ -0,0 +1,181 @@
...
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+import { VueConstructor } from 'vue'
+
+@Component({
+  name: 'PriceListView'
+})
+
+export default class PriceListView extends Vue {
+  data () {
```
shabeeb-aot, line 42: No need of data return. we can use private pricelist:

auth-web/src/views/auth/PriceListView.vue
jeznorth: No need. Table will suffice here.

auth-web/src/views/auth/PriceListView.vue
jeznorth: Updated as requested.

auth-web/src/views/auth/PriceListView.vue
```diff
@@ -0,0 +1,219 @@
...
+            </tbody>
+          </table>
+        </v-card>
+      </section>
+  </main>
+</template>
+
+<script lang="ts">
+import { Component, Vue } from 'vue-property-decorator'
+import { VueConstructor } from 'vue'
```
shabeeb-aot, line 77: Looks like this is not needed. we are not using it anywhere, can you crosscheck?

shabeeb-aot (approved):

PR #1273, opened 2021-01-11
staff level access made read only

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/5893

*Description of changes:*

1. API changes to remove staff access from PUT/PATCH
2. new permissions added for update pad info, change payment method etc
3. There was an issue with bank account validation.so thats fixed.
4. v- can changes. 
5. On creating new account , removed existing account id from session


auth-web/src/directives/can.ts
```diff
@@ -27,8 +37,12 @@ function canAccess (binding: DirectiveBinding, el: HTMLElement, node: VNode) {
   if (!okayToAccess) {
     if (behaviour === 'hide') {
       commentNode(el, node)
-    } else if (behaviour === 'disable') {
+    } else if (behaviour === 'disable' && isCard === false) {
```
shabeeb-aot, line 40: best practice is !isCard

shabeeb-aot (approved):

mengdong19 (approved):

PR #1347, opened 2021-02-08
Added suspend/activate accounts

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6121

*Description of changes:*

Added suspend/activate account patch end point. 
New role will be added in keycloak.

### DIFF
--- a/auth-api/migrations/versions/8260fac9943d_suspended_status.py
+++ b/auth-api/migrations/versions/8260fac9943d_suspended_status.py
@@ -1,0 +1,39 @@
+"""suspended status 
+
+Revision ID: 8260fac9943d
+Revises: b23a04d2e331
+Create Date: 2021-02-08 10:41:23.530382
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.sql import column, table
+from sqlalchemy import Boolean, String
+
+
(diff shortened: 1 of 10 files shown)


auth-api/migrations/versions/8260fac9943d_suspended_status.py
```diff
@@ -0,0 +1,39 @@
...
+down_revision = 'b23a04d2e331'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Insert codes and descriptions for organization status
+    org_status_table = table('org_statuses',
+                             column('code', String),
+                             column('desc', String),
```
mengdong19, line 25: column name is not the same as insert statement below. 'desc' vs 'description'.

auth-api/tests/unit/services/test_org.py
```diff
@@ -290,7 +290,8 @@ def test_suspend_org(session):  # pylint:disable=unused-argument
 def test_update_duplicate_org(session):  # pylint:disable=unused-argument
     """Assert that an Org cannot be updated."""
     org = factory_org_service()
-
+    8260
```
mengdong19, line 293: extra code here.

mengdong19 (approved):

PR #1356, opened 2021-02-10
 Staff manually suspend account#6121

shabeeb-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6121

*Description of changes:*


saravanpa-aot (commented):
This will work.But we want that to happen in the common-components so that when people logs into entities applicaiton , they get redirected .

shabeeb-aot:
> This will work.But we want that to happen in the common-components so that when people logs into entities applicaiton , they get redirected .

Its already done in common component. This is to handle after redirection we need user to redirect user suspend page or admin suspend page

saravanpa-aot (approved):

PR #1395, opened 2021-03-03
Changes for removing product role codes and adding ppr authorizations

sumesh-aot:
Changes for removing product role codes and adding ppr authorizations

Test cases for authorization changes

*Issue #:*
https://github.com/bcgov/entity/issues/6579

*Description of changes:*
- Removed product role code and product roles, and using permissions table for easy maintenance
- Removed some unused functions and its test cases
- Adding PPR auth only for premium accounts


karthik-aot (approved):

auth-api/src/auth_api/models/product_code.py
```diff
@@ -36,3 +37,8 @@ def find_by_code(cls, code):
     def get_all_products(cls):
         """Get all of the products codes."""
         return cls.query.all()
+
+    @classmethod
+    def find_by_type_code(cls, type_code: str) -> List[ProductCode]:
+        """Find an Org instance that matches the provided id and not in INACTIVE status."""
+        return cls.query.filter(ProductCode.type_code == type_code).all()
```
mengdong19, line 44: The comment mentioned not "INACTIVE", missed this filter here.

sumesh-aot: good catch! copy-paste error :) Fixed it

mengdong19 (approved):

PR #1417, opened 2021-03-09
Feature/6472 new UI behaviour

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6472

*Description of changes:*

- Change UI behavior to show error message rather than disable button


saravanpa-aot (commented):
which ticket is this? i think the ticket mentioned is wrong

karthik-aot:
Corrected the task number 6472 :)

PR #1424, opened 2021-03-10
6746-Two fields appearing with same information for staff

mengdong19:
*Issue #:*
https://github.com/bcgov/entity/issues/6746

*Description of changes:*
Duplicated status is removed.

### DIFF
--- a/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
+++ b/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
@@ -35,19 +35,8 @@
 
         <template v-show="!anonAccount">
           <div v-if="isStaff" class="nv-list-item mb-10">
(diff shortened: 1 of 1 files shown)


auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
```diff
@@ -35,19 +35,8 @@
 
         <template v-show="!anonAccount">
           <div v-if="isStaff" class="nv-list-item mb-10">
-            <div class="name" id="accountStatusStaff">Account Status</div>
+            <div class="name" id="accountStatusStaff"></div>
```
shabeeb-aot, line 38: I think we need to keep Account Status and remove Status code block

shabeeb-aot (approved):

PR #1436, opened 2021-03-16
Feature/6706

shabeeb-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6706

*Description of changes:*
* create account for govm users.
* stepper

### DIFF
--- a/auth-web/src/components/auth/create-account/GovmContactInfoForm.vue
+++ b/auth-web/src/components/auth/create-account/GovmContactInfoForm.vue
@@ -1,0 +1,146 @@
+<template>
+  <v-form ref="form"  data-test="form-govm-contact">
+    <p class="mb-9">Enter the IDIR email address of the ministry's employee.
+        An email will be sent this user to verify and activate this account. This user will be the admin of this account.</p>
+    <v-row >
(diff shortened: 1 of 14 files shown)


auth-web/src/components/auth/create-account/GovmContactInfoForm.vue
```diff
@@ -0,0 +1,146 @@
...
+  public confirmedEmailAddress = ''
+  // @Prop({ default: false }) isStepperView: boolean
+
+  $refs: {
+    form: HTMLFormElement
+  }
+
+  public emailRules = CommonUtils.emailRules()
+
+  public emailMustMatch (): string {
```
sumesh-aot, line 113: If the fields are readonly why do we need rules? Also is there a better pattern to carve out the email and confirm email as a component and re-use everywhere. Feels like lot of duplicate codes.
Also in this case it's readonly, so no need for rules as this will only confuse devs.

sumesh-aot (commented):
Please include test cases too

shabeeb-aot:
> Please include test cases too

Still working on it. will update in next PR

karthik-aot (approved):

saravanpa-aot (approved):

PR #1447, opened 2021-03-18
changes for org update for govm accounts

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6706

*Description of changes:*

Changes to support org updation for govm accounts


auth-api/src/auth_api/services/org.py
```diff
@@ -369,8 +372,16 @@ def update_org(self, org_info, token_info: Dict = None,  # pylint: disable=too-m
         current_app.logger.debug('<update_org ')
 
         has_org_updates: bool = False  # update the org table if this variable is set true
+        has_status_changing: bool = False
```
sumesh-aot, line 375: Org create and org update are getting complicated, we may need to think of a better pattern like adding some validator classes to abstract away all pre-validation. Now everything is kind of baked in code. Not a comment specific to this ticket.

sumesh-aot (approved):

auth-api/src/auth_api/services/org.py
saravanpa-aot: True.100% agreee.. have been thinking about this

PR #1452, opened 2021-03-18
6220 UI changes to consume remove affiliations API

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6220

*Description of changes:*

- Updated org service to pass resetPasscodeEmail and resetPasscodeFlag as body to delete affiliations request.
- Updated backend to resetpasscode based on the incoming flag from UI
- Added deletion confirmation modal


mengdong19 (approved):

auth-api/src/auth_api/resources/org.py
```diff
@@ -365,8 +365,10 @@ def delete(org_id, business_identifier):
         """Delete an affiliation between an org and an entity."""
         request_json = request.get_json()
```
sumesh-aot, line 366: Please make these fail-safe, as DELETE can be called without any payload

sumesh-aot: Also add or update test cases to cover these scenario.

auth-api/src/auth_api/resources/org.py
```diff
@@ -365,8 +365,10 @@ def delete(org_id, business_identifier):
...
         email_addresses = request_json.get('passcodeResetEmail')
+        should_passcode_reset = request_json.get('passcodeResetFlag')
```
saravanpa-aot, line 368: may be avoid the word flag in the variable ; call it resetPasscode or something similar to that

sumesh-aot (approved):

PR #1475, opened 2021-03-25
Adding int type for org_id in routes

sumesh-aot:
*Issue #:*

*Description of changes:*
- Adding int type for org_id in routes


auth-api/src/auth_api/resources/org.py
```diff
@@ -410,7 +410,7 @@ def get(org_id):
 
 
 @cors_preflight('DELETE,PATCH,OPTIONS')
-@API.route('/<string:org_id>/members/<string:membership_id>', methods=['DELETE', 'PATCH', 'OPTIONS'])
+@API.route('/<int:org_id>/members/<string:membership_id>', methods=['DELETE', 'PATCH', 'OPTIONS'])
```
mengdong19, line 413: Should membership_id also be int?

sumesh-aot: Yes, can make it int. This PR was done as there were lot of errors on the backend in production, as in some cases the org_id comes as undefined and causes lot of errors. making it int, flask will return a 404 without interrupting the code.

saravanpa-aot (approved):

PR #1486, opened 2021-03-31
Added product subscription for clients

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6872

*Description of changes:*

 added default subscription status to products table
added retrieving and adding of products for public users


auth-api/migrations/versions/b22ee0cd18ad_productsubscriptionsstatus_added.py
```diff
@@ -0,0 +1,88 @@
...
+
+    product_subscriptions_statuses_table = table('product_subscriptions_statuses',
+                                                 column('code', String),
+                                                 column('description', String),
+                                                 column('default', Boolean)
+                                                 )
+    op.bulk_insert(
+        product_subscriptions_statuses_table,
+        [
+            {'code': 'REJECTED', 'description': 'Status for a rejected account',
```
sumesh-aot, line 48: instead of account, can you add the description as product.

auth-api/migrations/versions/b22ee0cd18ad_productsubscriptionsstatus_added.py
```diff
@@ -0,0 +1,88 @@
...
+             'default': False},
+            {'code': 'ACTIVE', 'description': 'Status for a active account',
+             'default': True},
+            {'code': 'PENDING_STAFF_REVIEW', 'description': 'Status for a PENDING_STAFF_REVIEW account',
+             'default': False},
+            {'code': 'INACTIVE', 'description': 'Status for a inactive account',
+             'default': False},
+            {'code': 'SUSPENDED', 'description': 'Status for a SUSPENDED account',
```
sumesh-aot, line 56: Do we need SUSPENDED ?

auth-api/migrations/versions/b22ee0cd18ad_productsubscriptionsstatus_added.py
```diff
@@ -0,0 +1,88 @@
...
+             'default': False}
```
sumesh-aot, line 57: Also you will need to change the authorizations_view to pull only products ACTIVE

auth-api/src/auth_api/models/product_subscription.py
```diff
@@ -16,10 +16,11 @@
 The ProductSubscription object connects Org models to one or more ProductSubscription models.
 """
 
-from sqlalchemy import Column, ForeignKey, Integer
+from sqlalchemy import Column, ForeignKey, Integer, and_
 from sqlalchemy.orm import relationship
 
 from .base_model import VersionedModel
+from ..utils.roles import VALID_SUBSCRIPTION_STATUSES
```
sumesh-aot, line 23: We may need to think about a pattern for importing modules.. from a readability point of view its better to start from the root module except for files from same package..

auth-api/src/auth_api/services/authorization.py
```diff
@@ -164,8 +163,11 @@ def as_dict(self, expanded: bool = False):
 
     @staticmethod
     def _is_product_based_auth(product_code):
+
         check_product_based_auth = False
         if product_code:
+            from auth_api.services.products import \
```
sumesh-aot, line 169: why is it needed here? I dont see any usage

saravanpa-aot: cyclic import. now authorization uses products and vice versa

auth-api/migrations/versions/b22ee0cd18ad_productsubscriptionsstatus_added.py
saravanpa-aot: I missed it..Will add in the next PR ? i mean right now doing it ..but will get it with the next PR

auth-api/migrations/versions/b22ee0cd18ad_productsubscriptionsstatus_added.py
saravanpa-aot: removed

auth-api/migrations/versions/b22ee0cd18ad_productsubscriptionsstatus_added.py
saravanpa-aot: fixed

sumesh-aot (approved):

PR #1491, opened 2021-04-06
6886 - Add Task Patch endpoints

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6886

*Description of changes:*

- Added migration script to add new column related-to in tasks table.
- Added Patch endpoint for task resource and updateTask in services and model files.
- Updated the tests 


auth-api/src/auth_api/resources/task.py
```diff
@@ -23,27 +24,52 @@
...
-
 API = Namespace('tasks', description='Endpoints for tasks management')
 TRACER = Tracer.get_instance()
 _JWT = JWTWrapper.get_instance()
 
 
-@cors_preflight('GET,POST,OPTIONS')
-@API.route('', methods=['GET', 'POST', 'OPTIONS'])
+@cors_preflight('GET,OPTIONS')
+@API.route('/<string:task_relationship_type>', methods=['GET', 'POST', 'OPTIONS'])
```
sumesh-aot, line 33: May be its better to get the tasks based on type rather than task_relationship_type, or make it as query param.

auth-api/src/auth_api/resources/task.py
```diff
@@ -23,27 +24,52 @@
...
+            tasks = TaskService.fetch_tasks(task_relationship_type=task_relationship_type)
             response, status = {'tasks': TaskSchema().dump(tasks, many=True)}, http_status.HTTP_200_OK
         except BusinessException as exception:
             response, status = {'code': exception.code, 'message': exception.message}, exception.status_code
         return response, status
+
+
+@cors_preflight('PATCH,OPTIONS')
+@API.route('/<int:task_id>/status', methods=['PATCH', 'OPTIONS'])
+class TaskStatus(Resource):
```
sumesh-aot, line 54: May be make it a PUT on task, PATCH may become complicated later [lesson learned from org patch].

auth-api/src/auth_api/services/org.py
```diff
@@ -112,6 +114,11 @@ def create_org(org_info: dict, user_id,  # pylint: disable=too-many-locals, too-
             org.status_code = OrgStatus.PENDING_STAFF_REVIEW.value
             user = UserModel.find_by_jwt_token(token=token_info)
             Org.send_staff_review_account_reminder(user, org.id, origin_url)
+            # create a staff review task for this account
+            task_info = {'relationshipName': org.name,
```
sumesh-aot, line 118: try to stick with consistent naming.. use `name` instead of relationshipName. Also I think here it should pass the task name [default to org name], relationship_type : Org, relationship_id : org id.

auth-api/src/auth_api/utils/enums.py
```diff
@@ -230,3 +230,23 @@ class ProductCode(Enum):
...
+
+    ORG = 'ORG'     # Task related to Org staff review
+    AFFIDAVIT = 'AFFIDAVIT'
+
+
+class TaskStatus(Enum):
+    """Task relationship type."""
+
+    OPEN = 'OPEN'     # Open Task - needs to be taken action
+    CLOSE = 'CLOSE'   # Task has been acted upon
```
sumesh-aot, line 246: may be make it CLOSED or COMPLETED ?

auth-api/src/auth_api/services/task.py
```diff
@@ -55,21 +60,54 @@ def as_dict(self):
...
+                               relationship_id=org_info.get('relationshipId'),
+                               type=TaskType.PENDING_STAFF_REVIEW.value,
+                               status=TaskStatus.OPEN.value,
+                               related_to=org_info.get('relatedTo'))
         task_model.save()
+        return Task(task_model)
 
+    def update_task(self, task_info: Dict = None, token_info: Dict = None):
+        """Update a task record."""
+        current_app.logger.debug('<update_task ')
```
saravanpa-aot, line 78: hm.i thought we will update the status of the org as well when task is updated?

auth-api/src/auth_api/services/task.py
```diff
@@ -55,21 +60,54 @@ def as_dict(self):
         return obj
 
     @staticmethod
-    def create_task(task_model: TaskModel):
+    def create_task(org_info: dict):
```
saravanpa-aot, line 63: isnt it task_info rather than org_info?

auth-api/src/auth_api/services/task.py
```diff
@@ -55,21 +60,54 @@ def as_dict(self):
...
         """Create a new task record."""
-        current_app.logger.debug('<create_staff_task ')
-
+        current_app.logger.debug('<create_task ')
+        task_model = TaskModel(name=org_info.get('name'),
```
saravanpa-aot, line 66: this should be expandable to other kind of relations as well rite? like affidavit approval..Not just org approval

auth-api/src/auth_api/resources/task.py
```diff
@@ -40,8 +44,55 @@ class Tasks(Resource):
...
+            task = TaskService(TaskModel.find_by_task_id(task_id))
+            if task:
+                # Update task
+                origin = request.environ.get('HTTP_ORIGIN', 'localhost')
+                relationship_status = request_json.pop('relationshipStatus')
+                response, status = task.update_task(task_info=request_json,
+                                                    token_info=token).as_dict(), http_status.HTTP_200_OK
+
+                # Update relationships with the status
+                if request_json.get('relationshipType') == TaskRelationshipType.ORG.value:
```
saravanpa-aot, line 86: hm.. Would service layer be a better place to handle this ?

sumesh-aot (approved):

PR #1515, opened 2021-04-14
Feature/4118 backend part2

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4118

*Description of changes:*

- Added getTaskById endpoint in task resource
- Updated unit tests
- Added alembic migration script for inserting pending orgs in tasks(for prod)


auth-api/migrations/versions/885632ab6357_insert_tasks_table_with_pending_orgs.py
```diff
@@ -0,0 +1,58 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    org_res = conn.execute(f"SELECT * FROM orgs WHERE status_code = 'PENDING_STAFF_REVIEW';")
+    org_list: List[Org] = org_res.fetchall()
+
+    for org in org_list:
```
saravanpa-aot, line 27: mostly it will be no orgs..hope this code will not break if org list is emoty

auth-api/migrations/versions/885632ab6357_insert_tasks_table_with_pending_orgs.py
```diff
@@ -0,0 +1,58 @@
...
+        org_id = org.id
+        user_id = org.created_by_id
+        created_time = datetime.datetime.now()
```
saravanpa-aot, line 30: use created_time as orgs created time

auth-api/migrations/versions/885632ab6357_insert_tasks_table_with_pending_orgs.py
```diff
@@ -0,0 +1,58 @@
...
+                   f"VALUES "
+                   f"('{created_time}', '{created_time}', '{name}', '{date_submitted}', '{task_relationship_type}',"
+                   f" {org_id}, {user_id}, {user_id}, {user_id}, '{status}', '{task_type}')")
+
+    pass
+
+
+def downgrade():
+    conn = op.get_bind()
+    org_res = conn.execute(f"SELECT * FROM orgs WHERE status_code = 'PENDING_STAFF_REVIEW';")
```
saravanpa-aot, line 49: you could just delete from tasks since this will be the first time of migration; but good to do this as well

auth-api/tests/unit/api/test_task.py
```diff
@@ -156,3 +156,15 @@ def test_put_task_product(client, jwt, session, keycloak_mock):  # pylint:disabl
...
+def test_fetch_task(client, jwt, session):  # pylint:disable=unused-argument
+    """Assert that the task can be fetched by id."""
+    user = factory_user_model()
+    task = factory_task_service(user.id)
+    task_id = task._model.id
+
+    headers = factory_auth_header(jwt=jwt, claims=TestJwtClaims.staff_role)
+    rv = client.get('/api/v1/tasks/{}'.format(task_id), headers=headers, content_type='application/json')
+    assert rv.status_code == http_status.HTTP_200_OK
+    assert rv.json.get('name') == task._model.name
```
saravanpa-aot, line 170: not in this PR ;but can you write a big test case for asserting org status and mail notification when task is completed... like a full workflow... create org ; make sure task is getting created ; update task ;make sure org is udpated and mail is being sent [mock the mail code and check number of invocation]... this is low priority..but such a test case will be really helpful

saravanpa-aot (approved):

auth-api/tests/unit/api/test_task.py
karthik-aot: It is already there with the previous PR :)

PR #1526, opened 2021-04-15
4118 - fetch tasks and task type update

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/4118

*Description of changes:*
- Fetch tasks implementation in Staff Dashboard UI
- Task Type update as per UX pin

### DIFF
--- a/auth-api/migrations/versions/885632ab6357_insert_tasks_table_with_pending_orgs.py
+++ b/auth-api/migrations/versions/885632ab6357_insert_tasks_table_with_pending_orgs.py
@@ -9,7 +9,7 @@ import datetime
 from typing import List
 from alembic import op
 from auth_api.models import Org
-from auth_api.utils.enums import TaskStatus, TaskType, TaskRelationshipType
+from auth_api.utils.enums import TaskStatus, TaskRelationshipType
 
 # revision identifiers, used by Alembic.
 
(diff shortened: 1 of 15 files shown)


auth-api/migrations/versions/885632ab6357_insert_tasks_table_with_pending_orgs.py
```diff
@@ -31,7 +31,7 @@ def upgrade():
         date_submitted = org.created
         name = org.name
         status = TaskStatus.OPEN.value
-        task_type = TaskType.PENDING_STAFF_REVIEW.value
+        task_type = "NEW ACCOUNT"
```
saravanpa-aot, line 34: can we use a default config ; from the environment

auth-api/src/auth_api/services/org.py
```diff
@@ -158,7 +158,7 @@ def _handle_bceid_status_and_notification(org, origin_url, token_info):
                      'relatedTo': user.id,
                      'dateSubmitted': datetime.today(),
                      'relationshipType': TaskRelationshipType.ORG.value,
-                     'type': TaskType.PENDING_STAFF_REVIEW.value,
+                     'type': "NEW ACCOUNT",
```
saravanpa-aot, line 161: again from environment

saravanpa-aot (approved):

PR #1559, opened 2021-04-23
Feature/6895 backend 1

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6895

*Description of changes:*

- Updated org.py to create task during govm update status process
- updated unit tests


auth-api/src/auth_api/resources/org.py
```diff
@@ -146,6 +146,7 @@ def put(org_id):
         valid_format, errors = schema_utils.validate(request_json, 'org')
         toke_info = g.jwt_oidc_token_info
         bearer_token = request.headers['Authorization'].replace('Bearer ', '')
+        origin = request.environ.get('HTTP_ORIGIN', 'localhost')
```
sumesh-aot, line 149: what is the use of origin URL ?

sumesh-aot (approved):

auth-api/src/auth_api/resources/org.py
karthik-aot: We use the origin_url to send notification mail. Since, am yet to get nod on whether to send mail to staff ( similar to bceid), so removing it

PR #1565, opened 2021-04-26
Open NR UI in same tab

severinbeauvais:
*Issue #:* https://github.com/bcgov/entity/issues/7304

*Description of changes:*
- open NR UI in current tab instead of in new (blank) tab (3 places)
- it is hoped this will retain all session/cookie data so that account + user are retained from sbc-auth to NR UI
- also fixed/updated unit tests

### DIFF
--- a/auth-web/src/views/auth/home/RequestNameView.vue
+++ b/auth-web/src/views/auth/home/RequestNameView.vue
@@ -25,8 +25,7 @@
         <div class="request-name-info-btns mt-5">
           <name-request-button />
           <p class="mt-5">Have an existing Name Request?
-            <a :href="nameRequestExistingUrl"
-              target="_blank" rel="noopener noreferrer" class="status-link">
+            <a class="status-link" @click="goToNameRequestExisting()">
               Check your Name Request Status
             </a>
           </p>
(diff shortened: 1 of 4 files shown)


sumesh-aot (approved):

auth-web/src/views/auth/home/RequestNameView.vue
```diff
@@ -37,9 +36,13 @@
       </v-col>
       <!-- Image Column -->
       <v-col cols="12" md="6">
-        <a :href="nameRequestUrl" target="_blank">
-          <v-img src="../../../assets/img/Step2_NameRequest_x2.png" aspect-ratio="1.2" contain></v-img>
-        </a>
+        <v-img
+          src="../../../assets/img/Step2_NameRequest_x2.png"
```
cameron-freshworks, line 40: Could this work with `@/assets`? Or is the assets outside the `src`?

cameron-freshworks (approved):
Looks good!

auth-web/src/views/auth/home/RequestNameView.vue
severinbeauvais: I don't know if Vue would do that replacement in the template. Maybe.

Hey wait a minute, didn't you write this code originally? 😁

PR #1583, opened 2021-04-30
Provide access to BCOL Vital Stats user If they had access before

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6887

*Description of changes:*


auth-api/migrations/versions/31ec16d4f1e9_product_migration_for_vs.py
```diff
@@ -0,0 +1,49 @@
...
+down_revision = 'a37f90e6802d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Query all orgs which are linked to BCOL.
+    conn = op.get_bind()
+    org_res = conn.execute(
+        "select o.id, o.bcol_user_id from orgs o where bcol_user_id is not null and bcol_account_id is not null;"
```
saravanpa-aot, line 29: may be check the status_code as well ;there could be orgs which are inactive... its ok otherwise as well...

auth-api/migrations/versions/31ec16d4f1e9_product_migration_for_vs.py
```diff
@@ -0,0 +1,49 @@
...
+        token = RestService.get_service_account_token()
+    for org_id in orgs:
+        try:
+            print('Getting bcol profile for ', org_id[0], org_id[1])
+            bcol_response = RestService.get(endpoint=current_app.config.get('BCOL_API_URL') + f'/profiles/{org_id[1]}',
+                                            token=token)
+            print('BCOL Response', bcol_response.json())
+            Product.create_subscription_from_bcol_profile(org_id[0], bcol_response.json().get('profileFlags'))
+        except Exception as exc:
+            print('Profile Error')
```
saravanpa-aot, line 43: shud we raise error ? incase if some connection issues , we can stop the migration

saravanpa-aot (approved):

auth-api/migrations/versions/31ec16d4f1e9_product_migration_for_vs.py
sumesh-aot: ok, can raise an error.

PR #1607, opened 2021-05-11
Feature/7381

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7381

*Description of changes:*

- Added Autocomplete component to get similar names from von api.
- added integration with premium and basic accounts
- added unit tests

### DIFF
--- a/auth-web/src/components/auth/create-account/AccountCreateBasic.vue
+++ b/auth-web/src/components/auth/create-account/AccountCreateBasic.vue
@@ -19,11 +19,12 @@
         :readonly="govmAccount"
         autocomplete="off"
       />
-      <auto-complete-view
+      <org-name-auto-complete
+      v-if="enableOrgNameAutoComplete"
       :searchValue="autoCompleteSearchValue"
       :setAutoCompleteIsActive="autoCompleteIsActive"
       @auto-complete-value="setAutoCompleteSearchValue">
-      </auto-complete-view>
+      </org-name-auto-complete>
       <v-text-field
         filled
         label="Branch/Division (If applicable)"
@@ -91,11 +92,11 @@ import { Component, Mixins, Prop, Watch } from 'vue-property-decorator'
 import { Member, Organization } from '@/models/Organization'
 import { mapActions, mapMutations, mapState } from 'vuex'
 import { Address } from '@/models/address'
-import AutoCompleteView from '@/views/auth/AutoCompleteView.vue'
 import BaseAddressForm from '@/components/auth/common/BaseAddressForm.vue'
 import ConfirmCancelButton from '@/components/auth/common/ConfirmCancelButton.vue'
 import LaunchDarklyService from 'sbc-common-components/src/services/launchdarkly.services'
 import OrgModule from '@/store/modules/org'
+import OrgNameAutoComplete from '@/views/auth/OrgNameAutoComplete.vue'
 import Steppable from '@/components/auth/common/stepper/Steppable.vue'
 import { addressSchema } from '@/schemas'
 import { getModule } from 'vuex-module-decorators'
@@ -104,7 +105,7 @@ import { getModule } from 'vuex-module-decorators'
   components: {
     BaseAddressForm,
     ConfirmCancelButton,
-    AutoCompleteView
+    OrgNameAutoComplete
   },
   computed: {
     ...mapState('org', [
@@ -168,6 +169,10 @@ export default class AccountCreateBasic extends Mixins(Steppable) {
 
   private get enablePaymentMethodSelectorStep (): boolean {
     return LaunchDarklyService.getFlag(LDFlags.PaymentTypeAccountCreation) || false
+  }
+
+  private get enableOrgNameAutoComplete (): boolean {
+    return LaunchDarklyService.getFlag(LDFlags.EnableOrgNameAutoComplete) || false
   }
 
   private get address () {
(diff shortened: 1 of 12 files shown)


auth-web/src/services/von.services.ts
```diff
@@ -0,0 +1,10 @@
+import { AutoCompleteResponseIF } from '@/models/AutoComplete'
```
saravanpa-aot, line 1: whats IF?

auth-web/tests/unit/views/AutoCompleteView.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  it('does not display when inactive', async () => {
+    wrapper.setProps({ setAutoCompleteIsActive: false })
+    wrapper.setProps({ searchValue: 'test' })
+    // 3 ticks: watcher update, method run, results update
+    await Vue.nextTick()
+    await Vue.nextTick()
+    await Vue.nextTick()
+    expect(wrapper.find('[data-test="auto-complete-card"]').exists()).toBe(false)
+  })
+  it('display when active', async () => {
```
saravanpa-aot, line 80: can we have a test to assert the silcing if its not too hard?

auth-web/src/views/auth/AutoCompleteView.vue
```diff
@@ -0,0 +1,120 @@
...
+        this.$emit('auto-complete-value', searchValue)
+      }
+    }
+
+    private async getAutoCompleteResults (searchValue: string) {
+      try {
+        const response: AutoCompleteResponseIF = await this.getAutoComplete(searchValue)
+        if (searchValue === this.searchValue && response?.results) {
+        // will take up to 5 results - similar to PPR
+          this.autoCompleteResults = response?.results.slice(0, 5)
```
saravanpa-aot, line 88: make 5 a constant please ; say like MAX_RESULTS_COUNT or something like that

auth-web/src/views/auth/AutoCompleteView.vue
```diff
@@ -0,0 +1,120 @@
...
+
+    @Watch('autoCompleteIsActive', { deep: true })
+    updateAutoCompleteIsActive (val) {
+      if (!val) {
+        this.autoCompleteResults = []
+      }
+    }
+
+    @Watch('autoCompleteSelected', { deep: true })
+    async emitSelectedValue (val) {
```
saravanpa-aot, line 75: is it possible to type the event? I forgot how it works;but if we can type ,it will be good

auth-web/src/views/auth/AutoCompleteView.vue
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Watch('searchValue', { deep: true })
+    async GetAutoCompleteResults (val, oldVal) {
+      if (oldVal !== val && this.autoCompleteIsActive) {
+        await this.getAutoCompleteResults(val)
+      }
+    }
+
+    @Watch('setAutoCompleteIsActive', { deep: true })
```
saravanpa-aot, line 62: do we need deep true?

auth-web/src/components/auth/create-account/AccountCreateBasic.vue
```diff
@@ -17,8 +17,14 @@
         :disabled="saving"
         data-test="input-org-name"
         :readonly="govmAccount"
+        autocomplete="off"
       />
-       <v-text-field
+      <auto-complete-view
```
saravanpa-aot, line 22: should we have org name auto complete or something like that ?
when we have <auto-complete-view> as the name ; its very generalised.. in our case , its the auto complete of the org name alone rite? just my opinion..feel free to disagree

auth-web/src/models/AutoComplete.ts
```diff
@@ -0,0 +1,11 @@
+// To handle Von autocomplete responses
+export interface AutoCompleteResponseIF {
+    total: number,
+    results: Array<AutoCompleteResultIF>
+}
+
+export interface AutoCompleteResultIF {
```
saravanpa-aot, line 7: now I know IF is for interfaces!! it kind of read like an if statement so far...
Can you check if the prefix/postfix to denote interface is a good pattern? I think not..may be worth a discussion

karthik-aot: Latest trend to include IF to imply the response may be occuring from API call. But yeah, I shall refactor to reflect the existing best practices

auth-web/src/services/von.services.ts
karthik-aot: :-P

shabeeb-aot (approved):

auth-web/src/services/von.services.ts
```diff
@@ -0,0 +1,10 @@
+import { AutoCompleteResponse } from '@/models/AutoComplete'
+import { AxiosResponse } from 'axios'
+import ConfigHelper from '@/util/config-helper'
+import { axios } from '@/util/http-util'
+
+export default class VonService {
+  public static async getAutoComplete (searchValue: string): Promise<AxiosResponse<AutoCompleteResponse>> {
+    return axios.get(`${ConfigHelper.getVonAPIUrl()}/search/autocomplete?q=${searchValue}`)
```
shabeeb-aot, line 8: check service will accept limit on get, so we can avoid slicing from frontend

karthik-aot: Unfortunately, there is no provision to get paginated results 
https://github.com/bcgov/TheOrgBook/blob/master/APISpec/openapi.yml

saravanpa-aot (approved):

PR #1640, opened 2021-05-27
7533 - Adding subscription as part org create payload

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7533

*Description of changes:*


auth-api/src/auth_api/services/org.py
```diff
@@ -167,14 +174,6 @@ def create_membership(access_type, org, user_id, **kwargs):
             # Add the user to account_holders group
             KeycloakService.join_account_holders_group()
 
-    @staticmethod
```
sumesh-aot, line 170: this is not in use anymore

pwei1018 (approved):

PR #1653, opened 2021-06-02
user settings sorting and updating header

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6921
*Description of changes:*
- Duplicate account warning view would display existing accounts in ascending order
- Update sbc header on switching account

### DIFF
--- a/auth-web/src/store/modules/user.ts
+++ b/auth-web/src/store/modules/user.ts
@@ -261,7 +261,18 @@ export default class UserModule extends VuexModule {
   public async getUserAccountSettings () {
     const response = await UserService.getUserAccountSettings(this.context.state['userProfile'].keycloakGuid)
     if (response && response.data) {
-      const orgs = response.data.filter(userSettings => (userSettings.type === 'ACCOUNT'))
+      // filter by account type and sort by name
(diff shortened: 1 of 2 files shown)


auth-web/src/store/modules/user.ts
```diff
@@ -261,7 +261,18 @@ export default class UserModule extends VuexModule {
   public async getUserAccountSettings () {
     const response = await UserService.getUserAccountSettings(this.context.state['userProfile'].keycloakGuid)
     if (response && response.data) {
-      const orgs = response.data.filter(userSettings => (userSettings.type === 'ACCOUNT'))
+      // filter by account type and sort by name
+      const orgs = response.data.filter(userSettings => (userSettings.type === 'ACCOUNT')).sort((a, b) => {
```
saravanpa-aot, line 265: use localeCompare with sensitivity for checking case insensitive..
also arrow function will make the whole code in one line i think

saravanpa-aot (changes requested):

auth-web/src/store/modules/user.ts
karthik-aot: addressed. :)

saravanpa-aot (approved):

PR #1702, opened 2021-06-25
pending orgs need to see payment methods

saravanpa-aot:
*Description of changes:*
added missing permission for pending orgs so that they can payment methods

### DIFF
--- a/auth-api/migrations/versions/c44fff21c830_pending_approval_view_permission.py
+++ b/auth-api/migrations/versions/c44fff21c830_pending_approval_view_permission.py
@@ -1,0 +1,46 @@
+"""pending_approval view permission
+
+Revision ID: c44fff21c830
+Revises: e6e295695b9a
+Create Date: 2021-06-25 09:30:56.448783
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql import column, table
+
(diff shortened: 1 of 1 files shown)


karthik-aot (approved):

auth-api/migrations/versions/c44fff21c830_pending_approval_view_permission.py
```diff
@@ -0,0 +1,46 @@
...
+    conn = op.get_bind()
+    res = conn.execute(
+        f"select max(id) from permissions;")
+    latest_id = res.fetchall()[0][0]
+
+    # Insert code values
+    op.bulk_insert(
+        permissions_table,
+        [
+            {'id': latest_id + 1, 'membership_type_code': 'ADMIN', 'org_status_code': 'PENDING_AFFIDAVIT_REVIEW',
```
sumesh-aot, line 39: have you tried without adding `id` as part of the migration? It would work if a sequence is already there.

sumesh-aot (approved):

auth-api/migrations/versions/c44fff21c830_pending_approval_view_permission.py
saravanpa-aot: Yeah..I think we need to improve it. karthik also had the same comment... will check it..

PR #1706, opened 2021-06-28
account staff review fix

karthik-aot:
*Issue #:*

- Staff accoutn mailer takes org_id rather than the related task id

*Description of changes:*

- extract task id in current method and send to mailer.
- add find task by relationship id in task model

### DIFF
--- a/auth-api/src/auth_api/services/org.py
+++ b/auth-api/src/auth_api/services/org.py
@@ -140,7 +140,7 @@ class Org:  # pylint: disable=too-many-public-methods
         org.commit()
 
         if is_bceid_status_handling_needed:
-            Org.send_staff_review_account_reminder(org.id, origin_url)
+            Org.send_staff_review_account_reminder(relationship_id=org.id, origin_url=origin_url)
 
         current_app.logger.info(f'<created_org org_id:{org.id}')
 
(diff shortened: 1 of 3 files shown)


auth-api/src/auth_api/services/org.py
```diff
@@ -815,26 +815,30 @@ def approve_or_reject(org_id: int, is_approved: bool, origin_url: str = None):
...
+        context_path = f'review-account/{task.id}'
         app_url = '{}/{}'.format(origin_url, current_app.config.get('AUTH_WEB_TOKEN_CONFIRM_PATH'))
         review_url = '{}/{}'.format(app_url, context_path)
         user_from_context: UserContext = kwargs['user_context']
         first_name = user_from_context.first_name
         last_name = user_from_context.last_name
 
         data = {
-            'accountId': org_id,
+            'accountId': task.id,
```
sumesh-aot, line 834: if it's not accountId, then ideally we should rename it, as passing task id as account id is not good for maintenance

auth-api/src/auth_api/services/org.py
```diff
@@ -815,26 +815,30 @@ def approve_or_reject(org_id: int, is_approved: bool, origin_url: str = None):
...
             'emailAddresses': recipient,
             'contextUrl': review_url,
             'userFirstName': first_name,
             'userLastName': last_name
         }
         try:
-            publish_to_mailer('staffReviewAccount', org_id=org_id, data=data)
+            publish_to_mailer('staffReviewAccount', org_id=task.id, data=data)
```
sumesh-aot, line 841: here too, task id is not org_id

auth-api/src/auth_api/services/org.py
karthik-aot: yeah. Removed it as we do not need account id for this particular template. Context url where task id is appended is only required

auth-api/src/auth_api/services/org.py
karthik-aot: yeah. Removed it

sumesh-aot (approved):

saravanpa-aot (approved):

PR #1708, opened 2021-06-29
show login options in team members tab

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7698

*Description of changes:*


auth-web/src/components/auth/account-settings/team-management/MemberDataTable.vue
```diff
@@ -478,6 +500,17 @@ export default class MemberDataTable extends Vue {
     )
   }
 
+  private isBceidOrBcscAccount (): boolean {
```
sumesh-aot, line 503: can it be a generic function name like isRegularAccount() ? we never know how IDP would change over time.

sumesh-aot (approved):

auth-web/src/components/auth/account-settings/team-management/MemberDataTable.vue
saravanpa-aot: Sure.. Just rename the method name or u mean dont check using accessType?

sumesh-aot: just the name, as it will be easy to change a code block within a function without impacting anywhere else.

karthik-aot: If we have a generic function, then we can have a getter as well and use it in template :)

karthik-aot (approved):

shabeeb-aot (approved):

PR #1711, opened 2021-06-29
 Remove unused code for account upgrade/downgrade/account name change…

shabeeb-aot:
…/bcol account updates

*Issue #:*
https://github.com/bcgov/entity/issues/7717

*Description of changes:*
* Remove codes related to account upgrade


saravanpa-aot (approved):

saravanpa-aot (approved):
tthis change looks good ;but need to remove the name update code from account info page as welll ;but i think we can do that as a part of next ticket..

PR #1713, opened 2021-06-30
Feature/7717 WIP

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7717

*Description of changes:*

- Remove upgrade/downgrade account logic
- Remove org name change logic


auth-api/tests/unit/api/test_org.py
```diff
@@ -584,10 +584,6 @@ def test_update_org(client, jwt, session, keycloak_mock):  # pylint:disable=unus
     # assert no contacts
     assert len(dictionary.get('contacts')) == 0
 
-    rv = client.put('/api/v1/orgs/{}'.format(org_id), data=json.dumps({'name': 'helo-duplicate'}),
-                    headers=headers, content_type='application/json')
```
saravanpa-aot, line 588: so what happens now when they try to update the name?
can we assert that condition as well?

auth-api/tests/unit/services/test_org.py
```diff
@@ -479,18 +467,6 @@ def test_find_org_by_name(session, auth_mock):  # pylint:disable=unused-argument
     assert found_org.get('orgs')[0].get('name') == org_name
 
 
-def test_raise_error_if_duplicate_name(session, auth_mock):  # pylint:disable=unused-argument
-    """Assert that when org name is duplicate, DATA_CONFILICT error will be raised."""
-    org = factory_org_service()
-    dictionary = org.as_dict()
```
saravanpa-aot, line 485: same here, what happens to this test?

auth-api/tests/unit/api/test_org.py
karthik-aot: Yeah. With this change, an invalid exception would be raised if the input json consists of name property.

auth-api/tests/unit/services/test_org.py
karthik-aot: Yeah, fixed in this instance as well :)

auth-api/src/auth_api/services/org.py
```diff
@@ -397,6 +334,7 @@ def update_org(self, org_info, origin_url: str = None):  # pylint: disable=too-m
 
         # Update mailing address Or create new one
         if mailing_address:
+            has_org_updates = True
```
saravanpa-aot, line 337: if mailing address is getting updated , why set has_org_updates=True? it goes to different table rite?

saravanpa-aot (commented):
if mailing address is getting updated , why set has_org_updates=True? it goes to different table rite?

auth-api/tests/unit/services/test_org.py
```diff
@@ -410,14 +410,24 @@ def test_create_org_with_duplicate_name_bcol(session, keycloak_mock, monkeypatch
...
 def test_update_org(session, monkeypatch):  # pylint:disable=unused-argument
     """Assert that an Org can be updated."""
     org = factory_org_service()
-
-    org.update_org(TestOrgInfo.org2)
+    org.update_org(TestOrgInfo.update_org_with_business_type)
 
     dictionary = org.as_dict()
-    assert dictionary['name'] == TestOrgInfo.org2['name']
+    print('----------')
```
saravanpa-aot, line 428: remove the prints

karthik-aot: Sorry yeah. Removed them

auth-api/src/auth_api/services/org.py
karthik-aot: There are two methods where we call this add_contact_to_org, create_org and update_org 
In the existing system, we call save at the end of create_org and update_org  that also saves contact_links. 
Is there other alternatives for this?

saravanpa-aot (approved):

sumesh-aot (approved):

PR #1721, opened 2021-07-08
Fix dissolve buttons not shown to staff

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/6270

*Description of changes:*

Fix dissolve buttons 

### DIFF
--- a/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
+++ b/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
@@ -126,7 +126,7 @@
           </template>
         </div>
 
-        <div v-if="editEnabled">
+        <div>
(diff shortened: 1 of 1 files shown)


shabeeb-aot (approved):

auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
```diff
@@ -140,7 +140,7 @@
             >
               Deactivate Account
             </v-btn>
-
+            <span v-if="editEnabled">
```
karthik-aot, line 143: Is there a reason for span, can we use v-if directly in v-btn ?

karthik-aot (approved):

PR #1746, opened 2021-08-23
8572 Old Manage business dashboard updates

cameron-freshworks:
*Issue #:* /bcgov/entity#8572

*Description of changes:*
- Open button redirects to the Name Request app and queries Name Request for Display.

- Adds a button called "Use this name request" in drop down for Incorporations and this will launch draft IA. This button is not shown for unsupported NRs and BEN change of name/alteration (ie, only show for BEN IAs). Set Using a FF.

- "Remove" button relocated into Drop down menu.

 - We are using the combo button design. Open/drop down menu including remove and the 'Use this name Request' conditionally and "Remove"

 - Makes request to retrieve and populate Name Request data for any affiliation of NR Type.

 - Removes auto draft creation upon manually affiliating a name request and does so ONLY when selecting "Use this Name Request"

- Open action to open businesses regularly OR navigate to Name Request for NR affiliations.

- Set FF for controlling IA creations.

- When navigating to the NAME REQUEST - pass parameters required to pull up Existing NR Summary. ** stretch goal.

### DIFF
--- a/auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
+++ b/auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
@@ -29,15 +29,73 @@
             <div class="meta">
               <v-list-item-title v-if="isNumberedIncorporationApplication(item)">Numbered Benefit Company</v-list-item-title>
               <v-list-item-title v-if="!isNumberedIncorporationApplication(item)">{{ item.name }}</v-list-item-title>
+              <v-list-item-title v-if="!item.name && isNameRequest(item.corpType.code)">{{ item.nameRequest.names[0].name }}</v-list-item-title>
               <v-list-item-subtitle v-if="isIncorporationNumber(item.corpType.code)">Incorporation Number: {{ item.businessIdentifier }}</v-list-item-subtitle>
               <v-list-item-subtitle v-if="isNameRequest(item.corpType.code)">Name Request ({{ item.businessIdentifier }})</v-list-item-subtitle>
               <v-list-item-subtitle v-if="isTemporaryBusinessRegistration(item.corpType.code)">Incorporation Application</v-list-item-subtitle>
             </div>
           </template>
-          <template v-slot:[`item.action`]="{ item }">
+
+          <!-- Actions -->
+          <template v-slot:[`item.action`]="{ item, index }">
             <div class="actions">
-              <v-btn small color="primary" @click="goToDashboard(item)" title="Go to Business Dashboard" data-test="goto-dashboard-button">Open</v-btn>
-              <v-btn v-can:REMOVE_BUSINESS.disable small depressed @click="removeBusiness(item)" title="Remove Business" data-test="remove-button">Remove</v-btn>
+              <span class="open-action">
+                <v-btn
+                    small
+                    color="primary"
+                    min-width="5rem"
+                    min-height="2rem"
+                    class="open-action-btn"
+                    data-test="open-action-button"
+                    @click="open(item)"
+                >
+                  Open
+                </v-btn>
+              </span>
+
+              <!-- More Actions Menu -->
+              <span class="more-actions mr-4">
+                <v-menu
+                    offset-y left nudge-bottom="4"
+                    v-model="index"
+                >
+                  <template v-slot:activator="{ on }">
+                    <v-btn
+                        small
+                        color="primary"
+                        min-height="2rem"
+                        class="more-actions-btn"
+                        v-on="on"
+                    >
+                      <v-icon>{{index ? 'mdi-menu-up' : 'mdi-menu-down'}}</v-icon>
+                    </v-btn>
+                  </template>
+                  <v-list>
+                    <v-list-item
+                        v-if="isNameRequest(item.corpType.code) && isApprovedForIA(item)"
+                        class="actions-dropdown_item"
+                        data-test="use-name-request-button"
+                        @click="createDraftGoToDashboard(item)"
+                    >
+                      <v-list-item-subtitle>
+                        <v-icon small>mdi-file-certificate-outline</v-icon>
+                        <span class="pl-1">Use this Name Request</span>
+                      </v-list-item-subtitle>
+                    </v-list-item>
+                    <v-list-item
+                        class="actions-dropdown_item"
+                        data-test="remove-button"
+                        v-can:REMOVE_BUSINESS.disable
+                        @click="removeBusiness(item)"
+                    >
+                      <v-list-item-subtitle>
+                        <v-icon small>mdi-delete</v-icon>
+                        <span class="pl-1">Remove</span>
+                      </v-list-item-subtitle>
+                    </v-list-item>
+                  </v-list>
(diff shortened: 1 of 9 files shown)


auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
```diff
@@ -111,6 +170,12 @@ export default class AffiliatedEntityList extends Vue {
     return corpType === CorpType.NAME_REQUEST
   }
 
+  private isApprovedForIA (business: Business): boolean {
+    return this.isNameRequest(business.corpType.code) &&
+      business.nameRequest?.state === NrState.APPROVED &&
+      LaunchDarklyService.getFlag(LDFlags.IaSupportedEntities)?.includes(business.nameRequest?.legalType)
```
sumesh-aot: shouldn't this be using the new attributes Thor added on NR GET response instead of checking against a configured value?

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
```diff
@@ -29,15 +29,73 @@
             <div class="meta">
               <v-list-item-title v-if="isNumberedIncorporationApplication(item)">Numbered Benefit Company</v-list-item-title>
               <v-list-item-title v-if="!isNumberedIncorporationApplication(item)">{{ item.name }}</v-list-item-title>
+              <v-list-item-title v-if="!item.name && isNameRequest(item.corpType.code)">{{ item.nameRequest.names[0].name }}</v-list-item-title>
```
sumesh-aot, line 32: Does NR always sends the approved name as first one in array? If not, it cannot be assumed that the name is first one.

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
```diff
@@ -111,6 +170,15 @@ export default class AffiliatedEntityList extends Vue {
     return corpType === CorpType.NAME_REQUEST
   }
 
+  private isApprovedForIA (business: Business): boolean {
+    // Split string tokens into an array to avoid false string matching
+    const supportedEntityFlags = LaunchDarklyService.getFlag(LDFlags.IaSupportedEntities)?.split(' ')
```
severinbeauvais, line 175: Is the default value falsey if LD can't get this flag (LD id not configured or network issue)?

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
```diff
@@ -111,6 +170,15 @@ export default class AffiliatedEntityList extends Vue {
...
+
+    return this.isNameRequest(business.corpType.code) &&
+      business.nameRequest?.state === NrState.APPROVED &&
```
severinbeauvais, line 178: Is Conditionally Approved a value state for using a NR?

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
cameron-freshworks: I thought i heard something about this but in a GET on `https://legal-api-dev.apps.silver.devops.gov.bc.ca/api/v1/nameRequests/<nrNumber>` we currently only do a proxy request through LEAR->NAMEX and back, i didn't see any additional properties. 
Where should i look for these?

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
cameron-freshworks: In this Dashboard (old), we will just be taking the first name, the approved name.
My understanding is that these Auto Affiliated NR's wont be published until APPROVED on their side, so this should do in the meantime.

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
cameron-freshworks: For sure, with optional chaining it would return `undefined` ie a falsey value in that event.

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
cameron-freshworks: I'm not sure, good question. Was not called out in the tickets🤷

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
sumesh-aot: Not sure if it is implemented yet. If not in place, I guess it's good to go with configured value.

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
sumesh-aot: Names publishes any state change to the NR and we add the NR to the account when the status is DRAFT, so it's not only APPROVED names but other states can also appear.
Also you may need to check the status as well, before you start a draft IA.

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
cameron-freshworks: Totally, i believe once it is implemented we would replace the first two properties but the third (feature flag) would most likely still be required for control on what we allow as per environment. 

Thanks for bringing this up, will look into these new properties and where to find them 👍

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
cameron-freshworks: Oh interesting, Andre said they were only publishing on APPROVED states. I'll look into it. 

But yes, we do look at State before enabling the ability to Create the Draft IA.

cameron-freshworks: Will update this condition, Namex is passing the individual name state 👍

cameron-freshworks: Updated to display the APPROVED name or the first name only in the event of a DRAFT.

Future state on the new dashboard WILL display all the draft/approved names, however that wasn't a requirement of the OLD dashboard.

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
cameron-freshworks: Will update this in the future if the requirements are updated 👍

sumesh-aot (approved):
LGTM

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
severinbeauvais: OK, I'll resolve this comment, but please ask BA if maybe this was a missed req.

auth-web/src/components/auth/manage-business/AffiliatedEntityList.vue
severinbeauvais: OK, and you obviously handle that 👍

PR #1754, opened 2021-08-31
Applied target values for Unsupported NR Routing

cameron-freshworks:
*Issue #:*
https://github.com/bcgov/entity/issues/8570

*Description of changes:*

* Applied NR Target value to use in determine redirects and routing for supported and unsupported nrs.
* Updated Configuration for onestop and corporate online urls.
- This will require config map additions in Dev/Test/Prod for both onestop and corp online urls.** THIS WAS DONE BY RICHARD
- Includes the following UX updates from the document below.

https://docs.google.com/document/d/1CHcdQej6JofFdazuyIIGpMSfBq9BrfjpsgSCA2wGorA/edit?usp=sharing

### DIFF
--- a/auth-web/public/config/configuration.json
+++ b/auth-web/public/config/configuration.json
@@ -22,5 +22,7 @@
   "PAY_API_URL": "https://pay-api-dev.apps.silver.devops.gov.bc.ca",
   "PAY_API_VERSION": "/api/v1",
   "LEGAL_API_URL": "https://legal-api-dev.apps.silver.devops.gov.bc.ca",
-  "LEGAL_API_VERSION": "/api/v1"
+  "LEGAL_API_VERSION": "/api/v1",
+  "ONE_STOP_URL": "https://dev.onestop.gov.bc.ca/",
+  "CORPORATE_ONLINE_URL": "https://dev.corponline.gov.bc.ca/"
 }
--- a/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
+++ b/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
@@ -71,7 +71,7 @@
                         v-if="isApprovedForIA(item)"
                         class="actions-dropdown_item"
                         data-test="use-name-request-button"
-                        @click="createDraft(item)"
+                        @click="useNameRequest(item)"
                       >
                         <v-list-item-subtitle>
                           <v-icon>mdi-file-certificate-outline</v-icon>
@@ -113,6 +113,7 @@ import {
   FilingTypes,
   LDFlags,
   NrState,
+  NrTargetTypes,
   SessionStorageKeys
 } from '@/util/constants'
 import { Business, BusinessRequest, NameRequest, Names } from '@/models/business'
@@ -185,9 +186,9 @@ export default class AffiliatedEntityTable extends Vue {
     // Split string tokens into an array to avoid false string matching
     const supportedEntityFlags = LaunchDarklyService.getFlag(LDFlags.IaSupportedEntities)?.split(' ') || []
 
-    return this.isNameRequest(business.corpType.code) &&
-      business.nameRequest?.enableIncorporation &&
-      supportedEntityFlags.includes(business.nameRequest?.legalType)
+    return this.isNameRequest(business.corpType.code) && // Is this a Name Request
+      business.nameRequest?.enableIncorporation && // Is the Nr state approved or conditionally approved
+      supportedEntityFlags.includes(business.nameRequest?.legalType) // Feature flagged Nr types
   }
 
   /** Returns the Name value for the affiliation */
@@ -280,17 +281,25 @@ export default class AffiliatedEntityTable extends Vue {
   }
 
   /** Handler method for draft IA creation and navigation */
-  async createDraft (item: Business) {
-    let businessIdentifier = item.businessIdentifier
-    // 3806 : Create new IA if the selected item is Name Request
-    // If the business is NR, indicates there is no temporary business. Create a new IA for this NR and navigate.
-    if (item.corpType.code === CorpType.NAME_REQUEST) {
-      this.isLoading = true
-      // Find business with name as the NR number and use it for redirection
-      businessIdentifier = await this.createBusinessRecord(item)
-      this.isLoading = false
+  async useNameRequest (item: Business) {
+    switch (item.nameRequest.target) {
+      case NrTargetTypes.LEAR:
+        // Create new IA if the selected item is Name Request
+        let businessIdentifier = item.businessIdentifier
+        if (item.corpType.code === CorpType.NAME_REQUEST) {
+          this.isLoading = true
+          businessIdentifier = await this.createBusinessRecord(item)
+          this.isLoading = false
+        }
+        this.goToDashboard(businessIdentifier)
+        break
+      case NrTargetTypes.ONESTOP:
+        this.goToOneStop() // Navigate to onestop for firms
+        break
+      case NrTargetTypes.COLIN:
+        this.goToCorpOnline() // Navigate to Corporate Online for Corps
+        break
     }
-    this.goToDashboard(businessIdentifier)
   }
 
   /** Navigation handler for entities dashboard */
@@ -305,6 +314,16 @@ export default class AffiliatedEntityTable extends Vue {
   private goToNameRequest (nameRequest: NameRequest): void {
     ConfigHelper.setNrCredentials(nameRequest)
     window.location.href = `${ConfigHelper.getNameRequestUrl()}nr/${nameRequest.id}`
+  }
+
+  /** Navigation handler for OneStop application */
+  private goToOneStop (): void {
+    window.location.href = ConfigHelper.getOneStopUrl()
+  }
+
+  /** Navigation handler for Corporate Online application */
+  private goToCorpOnline (): void {
+    window.location.href = ConfigHelper.getCorporateOnlineUrl()
   }
 
   /** Create a business record in Lear */
--- a/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
+++ b/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
@@ -16,7 +16,11 @@
       >
         <template v-slot:item="{ item, index }">
           <tr>
-            <td v-if="isNameRequest(item.corpType.code)" class="header" :class="{ 'col-wide': getHeaders.length < 7 }">
+            <td
+              v-if="isNameRequest(item.corpType.code) && item.nameRequest"
+              class="header"
+              :class="{ 'col-wide': getHeaders.length < 7 }"
+            >
               <div v-for="(name, i) in item.nameRequest.names" :key="`nrName: ${i}`" class="pb-1 names-block">
                 <v-icon v-if="isRejectedName(name)" color="red" class="names-text pr-1" small>mdi-close</v-icon>
                 <v-icon v-if="isApprovedName(name)" color="green" class="names-text pr-1" small>mdi-check</v-icon>
@@ -116,7 +120,8 @@ import {
   BusinessState,
   CorpType,
   FilingTypes,
-  LDFlags, NrDisplayStates,
+  LDFlags,
+  NrDisplayStates,
   NrState,
   NrTargetTypes,
   SessionStorageKeys
@@ -218,7 +223,7 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
       case (this.isTemporaryBusinessRegistration(item.corpType.code)):
         return item.nrNumber
       case this.isNameRequest(item.corpType.code):
-        return item.nameRequest.nrNumber
+        return item.nameRequest?.nrNumber
       default:
         return item.businessIdentifier
     }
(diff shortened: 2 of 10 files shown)


auth-web/public/config/configuration.json
```diff
@@ -22,5 +22,7 @@
   "PAY_API_URL": "https://pay-api-dev.apps.silver.devops.gov.bc.ca",
   "PAY_API_VERSION": "/api/v1",
   "LEGAL_API_URL": "https://legal-api-dev.apps.silver.devops.gov.bc.ca",
-  "LEGAL_API_VERSION": "/api/v1"
+  "LEGAL_API_VERSION": "/api/v1",
+  "ONE_STOP_URL": "https://dev.onestop.gov.bc.ca/",
+  "CORPORATE_ONLINE_URL": "https://dev.corponline.gov.bc.ca/"
```
severinbeauvais, line 27: If these aren't yet populated in Openshift, ask some Ope guy (Patrick, Richard) to update 1Password accordingly.
And/or add this item to the respective release ticket (so we remember to do it).

auth-web/src/models/business.ts
```diff
@@ -61,7 +61,8 @@ export interface NameRequest {
     state?: string,
     applicantEmail?: string,
     applicantPhone?: string,
-    enableIncorporation?: boolean
+    enableIncorporation?: boolean,
+    target?: NrTargetTypes
```
severinbeauvais, line 65: Why not update and use the shared interface for a Name Request?

auth-web/public/config/configuration.json
cameron-freshworks: oh right, i forgot about those changes after the migration!
Will do, thanks 👍

auth-web/src/models/business.ts
cameron-freshworks: This interface is used for instances beyond just the response from the NR, so the optional properties might not be exactly the same. 
Will have a look though and use if able!

severinbeauvais: I think TS allows the use of Partial interfaces, which may allow you to use a shared interface and add in your own properties... but I haven't used that yet.

severinbeauvais: Or this: https://www.typescripttutorial.net/typescript-tutorial/typescript-extend-interface/

auth-web/public/config/configuration.json
cameron-freshworks: Contacted some SRE team members but i can't seem to find a release ticket. 
Just labelled as `Program Increment 10 - All` 🤷

auth-web/src/models/business.ts
cameron-freshworks: The discrepancy here is this interface is to define the Namex Response object (partially) not the NameRequest used by our UI's for Filings/Corrections etc. 

I could add a Namex Response interface in the BCRS repo and import it back.

auth-web/public/config/configuration.json
severinbeauvais: Ask James.

Or if the config for all environments is updated now, adding to release ticket isn't needed.

cameron-freshworks: Sounds like Richard is already on it, so will confirm and comment back here before resolving and/or merge etc 👍

auth-web/src/models/business.ts
severinbeauvais: 🤷‍♂️

severinbeauvais: OK, I think you should not add a one-time Namex Response interface to the shared repo.

If you can't find a way to leverage the usual NR contents (interface) with this interface (which is unfortunately named the same as the other NR interfaces) then just leave it, I guess. I was hoping to introduce some consistency / schema-like thing but oh well.

cameron-freshworks: 100% i know where you're coming from. It's caused me a little angst trying to sort this out because i do agree!

That being said, there is some interfaces overlap here (as well as other ui's) which require a bit more sorting out. 
What i mean is removing any localized NameRequest related Interfaces and just relying on the shared Interfaces, but would be a refactor stretching a little farther outside the scope here.

severinbeauvais: So what do you want to do?

cameron-freshworks: If we feel strongly enough about this, we should make a ticket.
I still feel like a guest in this repo, so i'm less inclined to refactor where it isn't absolutely required.

severinbeauvais: > I still feel like a guest in this repo
Haha, me too!

> If we feel strongly enough about this, we should make a ticket.
Sure, make a ticket. They can always decide to not do it (or make it low priority).

cameron-freshworks: Ticket is made! 
https://app.zenhub.com/workspaces/entity-5bf2f2164b5806bc2bf60531/issues/bcgov/entity/8794

severinbeauvais (approved):

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -111,40 +116,38 @@ import {
   BusinessState,
   CorpType,
   FilingTypes,
-  LDFlags,
+  LDFlags, NrDisplayStates,
```
cameron-freshworks, line 119: Fixed in next commit

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -213,7 +216,7 @@ export default class AffiliatedEntityTable extends Vue {
       case (this.isNumberedIncorporationApplication(item)):
         return 'Pending'
       case (this.isTemporaryBusinessRegistration(item.corpType.code)):
-        return 'Pending'
+        return item.nrNumber
```
cameron-freshworks, line 219: This is a new property returned from affiliations request, to preserve NR number once the affiliation has moved into a draft filing.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -243,7 +246,7 @@ export default class AffiliatedEntityTable extends Vue {
     switch (true) {
       case (this.isNameRequest(item.corpType.code)):
         // Format name request state value for Display
-        return item.nameRequest.state.charAt(0).toUpperCase() + item.nameRequest.state.slice(1).toLowerCase()
+        return NrDisplayStates[NrState[item.nameRequest.state]]
```
cameron-freshworks, line 249: Mapped these to make it cleaner and more consistent.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -253,22 +256,25 @@ export default class AffiliatedEntityTable extends Vue {
...
-    return item.lastModified
-      ? new Date(item.lastModified).toLocaleDateString('en-CA', {
-        timeZone: 'America/Vancouver' })
-      : this.$t('notAvailable').toString()
+    switch (true) {
+      case (!!item.lastModified):
+        return this.dateToPacificDate(new Date(item.lastModified))
+      case (!!item.modified):
+        return this.dateToPacificDate(new Date(item.modified))
+      default: return this.$t('notAvailable').toString()
```
cameron-freshworks, line 269: Using the date mixing to format these dates.

auth-web/src/util/constants.ts
```diff
@@ -143,9 +143,24 @@ export enum CorpType {
...
+
+export enum NrDisplayStates {
+    APPROVED = 'Approved',
+    HOLD = 'Pending Staff Review',
+    DRAFT = 'Pending Staff Review',
+    REJECTED = 'Rejected',
+    CONDITIONAL = 'Conditional Approval',
+    REFUND_REQUESTED = 'Cancelled, Refund Requested',
+    CANCELLED = 'Cancelled',
+    EXPIRED = 'Expired'
```
cameron-freshworks, line 163: Just for mapping states to display variations for those states.

auth-web/public/config/configuration.json
cameron-freshworks: Richard has applied these values to dev and the work order is in for TEST/PROD 😁

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -16,7 +16,7 @@
           <v-btn
             id="add-name-request-btn"
             class="font-weight-regular"
-            color="primary"
+            color="#1669BB"
```
karthik-aot, line 19: Can we move it to scoped scss at the bottom ?

karthik-aot (approved):

auth-web/src/components/auth/manage-business/EntityManagement.vue
cameron-freshworks: Absolutely, i could also define this blue (app-blue) on the vue instance, as primary === app-blue in our approach to keeping things consistent across all the apps.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -242,7 +251,7 @@ export default class AffiliatedEntityTable extends Vue {
     switch (true) {
```
severinbeauvais, line 251: What the...?

Can you fix this please?

cameron-freshworks: Can you elaborate?

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: Yes, please define a Vuetify theme colour as needed.

cameron-freshworks: Updated in next commit 👍

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -551,18 +550,37 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
...
   }
 
   // Vuetify Overrides
   ::v-deep {
-
     #dashboard-actions {
       .v-input .v-label {
         top: 30px;
         color: $gray7;
+        font-size: .875rem
```
severinbeauvais, line 583: ;

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
cameron-freshworks: This is just another technique to use when you have varying conditions not tied to a single type of value.
https://medium.com/trabe/using-switch-true-in-javascript-986e8ad8ae4f

I'm open to suggestions but personally i don't see anything wrong with this.

auth-web/src/components/auth/manage-business/EntityManagement.vue
cameron-freshworks: Thanks, will update!

auth-web/src/components/auth/mixins/DateMixin.vue
```diff
@@ -0,0 +1,28 @@
+// You can declare a mixin as the same style as components.
+<script lang="ts">
+import Component from 'vue-class-component'
+import Vue from 'vue'
+
+@Component({
+  name: 'DateMixin'
```
severinbeauvais, line 7: Are you sure this is needed?

cameron-freshworks: This syntax? It's how the rest of the mixins are declared in this repo, so i followed suite. 

If you are referring to the mixin itself, yes, this date formatting method could/would be re used i imagine.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: I see.

I'd prefer to see "if then else-if then else" here as it's less obtuse.

cameron-freshworks: I disagree, i feel like 2-3 layers of if - else - if - else is obtuse 🤷

shabeeb-aot (approved):

auth-web/src/components/auth/mixins/DateMixin.vue
severinbeauvais: Yes, I meant the syntax.

auth-web/src/plugins/vuetify.ts
```diff
@@ -15,6 +15,7 @@ export default new Vuetify({
     },
     themes: {
       light: {
+        primary: '#1669bb',
```
severinbeauvais, line 18: Wait - this sets "primary" colour for everyone. Is that OK? If your usage is a special case then I think you should declare a new name (not a built in name).

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: Agree to disagree 😁

auth-web/src/plugins/vuetify.ts
cameron-freshworks: No special case, i believe this is what we want app wide to bring it inline with the rest of our UI's. 
Vuetify primary blue is overridden everywhere else (other ui's) if i'm not mistaken.

auth-web/src/components/auth/manage-business/EntityManagement.vue
cameron-freshworks: Fixed in last commit 👍

auth-web/src/plugins/vuetify.ts
severinbeauvais: ok

PR #1763, opened 2021-09-10
Allow multiple affiliation of same entity

sumesh-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/8863

*Description of changes:*
- Changes for multiple affiliation of same entity


auth-api/src/auth_api/services/affiliation.py
```diff
@@ -178,19 +179,18 @@ def create_new_business_affiliation(org_id,  # pylint: disable=too-many-argument
...
             raise BusinessException(Error.NR_INVALID_CONTACT, None)
 
+        # Validate if org_id is valid by calling Org Service.
         org = OrgService.find_by_org_id(org_id, allowed_roles=CLIENT_AUTH_ROLES)
         if org is None:
             raise BusinessException(Error.DATA_NOT_FOUND, None)
 
         entity = EntityService.find_by_business_identifier(business_identifier, skip_auth=True)
-        # If entity already exists and is already affiliated to an org, throw error
+        # If entity already exists and passcode is already claimed, throw error
```
saravanpa-aot, line 193: i am guessing the code is alrite..whats the user action for this method?i just cant grasp quickly

saravanpa-aot (approved):

auth-api/src/auth_api/services/affiliation.py
sumesh-aot: The change here is to allow the affiliation even if it's claimed.  As per the latest discussion, they can just share the passcode and use it in another account. (but the original code may come back, thats why I didn't remove it)

PR #1764, opened 2021-09-10
Backend changes - 8664

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/8664

*Description of changes:*

- created migration script for updating remarks field to array(postgres) and list(python)
- refactor task service layer accordingly
- updated tests

How it would be in DB
image

Before Upgrade
image
After upgrade
image
After downgrade
image


auth-api/migrations/versions/304705f06971_tasks_multiple_reasons.py
```diff
@@ -0,0 +1,84 @@
...
+
+    conn = op.get_bind()
+    task_res = conn.execute("SELECT * FROM tasks;")
+    task_list: List[Task] = task_res.fetchall()
+
+    for task in task_list:
+        if task.remarks is None:
+            continue
+        # A check if there are task remarks that is already wrapped in '{}', ideally, we would not hit this scenario
+        if task.remarks.startswith('{') and task.remarks.endswith('}'):
```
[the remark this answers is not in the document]
karthik-aot: yeah. I shall refactor the same way like downgrade, check if it an array :)

saravanpa-aot, line 36: you can remove the whole thing rite? its an upgrade script ;so we will never have this scenario
also select * from can check for tasks which remarks exists

auth-api/migrations/versions/304705f06971_tasks_multiple_reasons.py
```diff
@@ -0,0 +1,84 @@
...
+            task_id = task.id
+            remarks = task.remarks.split(',')
+            # Now updating the new remarks value back in the table
+            update_sql = text("UPDATE tasks SET remarks = :remarks where id = :task_id") \
+                .params(
+                remarks=remarks, task_id=task_id)
+            op.execute(update_sql)
+
+    # ### alter command auto generated by Alembic ###
+    op.alter_column('tasks', 'remarks',
```
saravanpa-aot, line 48: shudnt this happen before the update statement?

docker/docker-compose.yml
```diff
@@ -60,19 +60,19 @@ services:
...
+      mem_limit: 512m
+      expose:
+        - 4222
+        - 8222
+      labels:
+        - entity.services=nats
+      ports:
+        - 4222:4222
+        - 8222:8222
+      tty: true
```
saravanpa-aot, line 76: why is this change?

auth-api/migrations/versions/304705f06971_tasks_multiple_reasons.py
```diff
@@ -0,0 +1,84 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # We get the list of tasks at first.
+    # we convert tasks.remarks from character type to array form ( check if there are arrays already, edge case )
+    # we convert the column remarks type from character to ARRAY
+
+    conn = op.get_bind()
```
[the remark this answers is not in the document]
karthik-aot: ARRAY type expects JSON value and would not accept string literal. 
image
Therefore, I transform the existing column value to JSON array type and then alter the column :)

auth-api/migrations/versions/304705f06971_tasks_multiple_reasons.py
karthik-aot: yeah refactored the condition now

auth-api/migrations/versions/304705f06971_tasks_multiple_reasons.py
karthik-aot: ARRAY type expects JSON value and would not accept string literal.

auth-api/migrations/versions/304705f06971_tasks_multiple_reasons.py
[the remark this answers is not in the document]
karthik-aot: Before Upgrade
image
After upgrade
image
After downgrade
image

sumesh-aot (approved):

saravanpa-aot (approved):

PR #1798, opened 2021-10-29
changes for supporting user role for revocation of api key

saravanpa-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/7071

*Description of changes:*

Just added staff and user role to api key delete end point


sumesh-aot (commented):
Need same change for GET keys too

sumesh-aot (approved):

PR #1817, opened 2021-11-19
9510 Reduced My Business Registry load time

severinbeauvais:
*Issue #:*
https://github.com/bcgov/entity/issues/9510

*Description of changes:*
- deleted obsolete AffiliatedEntityList component (and unit tests)
- deleted non-business table template code
- refactored isLoading
- ~don't call setup() initially~ _reverted (sometimes needed for initial load)_
- don't call syncBusinesses() if already running
- don't call syncBusinesses() if just run
- deleted obsolete FF code
- refactored syncBusinesses() to run using iterables and to make all network calls concurrently
- some lint fixes and minor cleanup


auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -1,12 +1,11 @@
 <template>
   <div>
```
severinbeauvais, line 2: Hide whitespace to see this diff correctly.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -291,22 +241,23 @@ import PasscodeResetOptionsModal from '@/components/auth/manage-business/Passcod
...
   private primaryBtnText = ''
   private secondaryBtnText = ''
   private primaryBtnHandler: () => void = undefined
   private secondaryBtnHandler: () => void = undefined
+  private lastSyncBusinesses = 0
 
   /** V-model for dropdown menu. */
   private addAffiliationDropdown: boolean = false
 
-  protected readonly currentAccountSettings!: AccountSettings
```
severinbeauvais, line 309: This is inherited from some mixins.

auth-web/src/store/modules/business.ts
```diff
@@ -25,10 +24,6 @@ export default class BusinessModule extends VuexModule {
   currentBusiness: Business = undefined
   businesses: Business[] = []
 
-  public get businessAffiliations (): Business[] {
```
severinbeauvais, line 28: Never used.

auth-web/src/store/modules/business.ts
```diff
@@ -40,72 +35,102 @@ export default class BusinessModule extends VuexModule {
     this.businesses = businesses
   }
 
-  @Action({ commit: 'setBusinesses', rawError: true })
```
severinbeauvais, line 43: I took out the "auto-commit" so it doesn't overwrite `businesses` on error. I now call setBusinesses() explicitly on line 133.

auth-web/src/store/modules/business.ts
```diff
@@ -40,72 +35,102 @@ export default class BusinessModule extends VuexModule {
...
-  public async syncBusinesses (): Promise<Business[]> {
-    const organization = this.context.rootState.org.currentOrganization
+  @Action({ rawError: true })
+  public async syncBusinesses (): Promise<void> {
+    /** Returns NR's approved name. */
```
severinbeauvais, line 40: I moved the helper functions to the top.

auth-web/src/store/modules/business.ts
```diff
@@ -40,72 +35,102 @@ export default class BusinessModule extends VuexModule {
...
+        return [] as Business[]
+      })
+    // console.log('*** affiliated entities =', affiliatedEntities) // eslint-disable-line no-console
 
-      return affiliatedEntities
-    }
+    // get only NR affiliations
+    const nameRequests = affiliatedEntities.map(entity => {
+      return (entity.corpType.code === CorpType.NAME_REQUEST) ? entity : null
+    })
```
severinbeauvais, line 82: I did this instead of filtering so that `nameRequests` is the same length as `affiliatedEntities`. I also did this in other statements below. The reason is so that I can write back into affiliatedEntries using an index on line 100.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -532,6 +489,13 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
 <style lang="scss" scoped>
   @import '$assets/scss/theme.scss';
 
+  .loading-container.grayed-out {
+    // these are the same styles as dialog overlay:
+    opacity: 0.46;
+    background-color: rgb(33, 33, 33); // grey darken-4
+    border-color: rgb(33, 33, 33); // grey darken-4
```
severinbeauvais, line 496: This changes the spinner to semi-transparent, so the user can see that the page is loading. It makes this "feel" faster :)

auth-web/src/store/modules/business.ts
```diff
@@ -40,72 +35,105 @@ export default class BusinessModule extends VuexModule {
...
+          return response.data.entities
         }
-      }
+        throw new Error(`Invalid response = ${response}`)
+      })
+      .catch(error => {
+        console.log('Error getting affiliated entities:', error) // eslint-disable-line no-console
+        return [] as Business[]
+      })
+    // console.log('*** affiliated entities =', affiliatedEntities) // eslint-disable-line no-console
```
severinbeauvais, line 77: Obviously I will remove all these logs in the final PR.

auth-web/src/store/modules/business.ts
```diff
@@ -40,72 +35,105 @@ export default class BusinessModule extends VuexModule {
...
 
-      return affiliatedEntities
-    }
+    // update store with initial results
+    this.setBusinesses(affiliatedEntities)
```
severinbeauvais, line 80: This displays the initial table (without all the NR updates). (The spinner is still displayed, but it gives feedback to the user that something is happening.)

severinbeauvais:
You guys can check out my code and run it locally to see it in action: `gh pr checkout 1817`

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -352,24 +300,31 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
...
   private async setup () {
-    this.isLoading = true
+    // ensure syncBusinesses isn't already running
+    if (this.isLoading === 1) {
+      return
+    }
+
+    // ensure syncBusinesses hasn't just been run
+    if (Date.now() - this.lastSyncBusinesses < 2000) {
+      return
```
severinbeauvais, line 314: [Vue Debounce Decorator](https://www.npmjs.com/package/vue-debounce-decorator) did not work for `syncBusinesses()` so I implemented this instead.

severinbeauvais: I used a 2-second duration because I found that the account changed handler often took almost 1 second to trigger this (setup method) a second time.

auth-web/src/store/modules/business.ts
```diff
@@ -37,75 +32,98 @@ export default class BusinessModule extends VuexModule {
 
   @Mutation
   public setBusinesses (businesses: Business[]) {
-    this.businesses = businesses
+    this.businesses = [...businesses]
```
severinbeauvais, line 35: This is needed to trigger reactivity.

auth-web/src/store/modules/business.ts
```diff
@@ -37,75 +32,98 @@ export default class BusinessModule extends VuexModule {
...
+    const isConditionallyApproved = (nr): boolean =>
+      (nr.state === NrState.CONDITIONAL &&
+        [NrConditionalStates.RECEIVED, NrConditionalStates.WAIVED].includes(nr.consentFlag))
+
+    /** Returns True if NR is approved for one stop registration. */
+    const isApprovedForRegistration = (nr): boolean =>
+      nr.actions.some(action => action.filingName === LearFilingTypes.REGISTRATION)
+
+    // initialize store
+    this.setBusinesses([])
```
severinbeauvais, line 59: This clears the old table (which may be visible through the semi-transparent loading spinner.

auth-web/src/components/auth/manage-business/EntityManagement.vue
cameron-freshworks: Looks good.
Is this load time roughly consistent across environments?

cameron-freshworks (commented):
Looks fine to me! Obviously a little more complexity but it appears necessary to avoid frequent setups.  
Will run locally!

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: I don't actually know. How would I test this? I think it's partly dependent on the environment's resources (web server/db).

That said, I expect it to be about the same. I don't see why it would be much different.

cameron-freshworks: Yea, i really don't think you could other than just watch the account change handler in Test (and prod if possible). 
I imagine it would be the same. 
If it was slower by 2-3 seconds for example would that be an issue though?

severinbeauvais: Yes, if it was longer than 2 seconds then the data/table would get loaded twice. That would suck. To test this, we would need a Prod user with an account with many affiliated entities/NRs. I'm thinking we wait for a user to complain about this.

FWIW, as far as I can tell, the delay in Test is the same as in Dev.

cameron-freshworks: kk, sounds good!

severinbeauvais: I could bump this up to 3000ms. It just has to be less than the time a speedy user would take to switch between accounts.

cameron-freshworks (approved):
Looks good! The ux is much improved with the loading transparency and the reduced affiliations requests.

severinbeauvais:
(I just spoke with Saravan and he's OK with this change.)

PR #1874, opened 2022-01-12
Auth Web - Implements Shared Breadcrumb

cameron-freshworks:
*Issue #:*
https://github.com/bcgov/entity/issues/10236

*Description of changes:*
* Imports and implements Shared Breadcrumb
* Removes Navigation tab bar
* Updated routes with Meta + Breadcrumb data


auth-web/src/App.vue
```diff
@@ -32,7 +32,7 @@
           <v-icon>mdi-close</v-icon>
         </v-btn>
       </v-snackbar>
-      <navigation-bar :configuration="navigationBarConfig" :hide="!showNavigationBar" />
+      <BreadCrumb v-if="showNavigationBar" class="pl-8" :breadcrumbs="breadcrumbs" />
```
severinbeauvais, line 35: Are you sure this needs "pl-8"? Or should it be in the same container as the body and only go to the container margins? Eg,

image

auth-web/src/App.vue
```diff
@@ -117,27 +107,16 @@ export default class App extends Mixins(NextPageMixin) {
...
-      ]
-    }
+  /** The route breadcrumbs list. */
+  get breadcrumbs (): Array<BreadcrumbIF> {
+    const breadcrumbs = [...(this.$route?.meta?.breadcrumb || [])]
+
+    // Apply dynamic return crumb depending on user role
+    // This will be replaced when we have our UBER dashboard / high level marketing page
+    const defaultCrumb = this.currentUser?.roles?.includes(Role.Staff) ? StaffDashboardBreadcrumb : HomeBreadCrumb
+    if (this.$route?.name === 'business') { breadcrumbs.unshift(defaultCrumb) }
```
severinbeauvais, line 117: Move block content to new line (or don't use curly braces) (but I prefer the former).

auth-web/src/assets/styles/theme.scss
```diff
@@ -0,0 +1,69 @@
+// Gray Palette
```
severinbeauvais, line 1: A full, duplicate copy of the theme variables isn't what I had in mind.

I thought this file would only contain:
```
import "@/assets/scss/theme.scss"
```

shabeeb-aot, line 1: I suppose we need this file since we didn't build shared component. I think we have to prioritize compiled shared components else we will be duplicating codes. Can we add only those colour codes needed for these components in this file and remove everything else. Else it may change the existing style in sbc-auth

auth-web/src/resources/BreadcrumbResources.ts
```diff
@@ -0,0 +1,19 @@
+import { BreadcrumbIF } from '@bcrs-shared-components/interfaces'
```
shabeeb-aot, line 1: Is this common for all applications? if so we can import from shared components only. We don't have a resources folder yet. Can we use this file inside the util folder for better structure? This will be one file inside folder

auth-web/src/assets/styles/theme.scss
cameron-freshworks: ah i see you what you mean now.

auth-web/src/App.vue
cameron-freshworks: will do!

auth-web/src/App.vue
cameron-freshworks: Where it is right now, it does require it. I think there is additional padding inside the views, which required some additional padding on the app level. Can have another look though.

severinbeauvais: The reason I'm asking is that the margins change depending on screen width.

The breacrumb should line up with the left side as above.

auth-web/src/routes/router.ts
```diff
@@ -120,35 +121,70 @@ export function getRoutes (): RouteConfig[] {
...
+            showNavBar: true
+          }
         },
         {
           path: 'incorporate-or-register',
+          name: 'incorporate-or-register',
           component: IncorpOrRegisterView,
-          meta: { showNavBar: false }
+          meta: {
+            breadcrumb: [
```
shabeeb-aot, line 164: will it be better to keep breadcrumb in one place? either in resources file or inline like this?

auth-web/src/App.vue
cameron-freshworks: Done in next commit

auth-web/src/assets/styles/theme.scss
cameron-freshworks: Done in next commit

cameron-freshworks: Refactored this to provide an access file for our shared components that imports the theme.file from it's current location. 
No changes to original theme files other than add an additional theme colour that has been introduced prior. 
`app-dk-blue`.

auth-web/src/resources/BreadcrumbResources.ts
cameron-freshworks: This isn't common for all as of yet. 
I believe once the Uber Dashboard or highest level landing/marketing page is complete, we can then then create a common resource. Could look at doing this now though, whatever you prefer.

auth-web/src/routes/router.ts
cameron-freshworks: It can be done like that when the crumb is common but each of these routes require a unique `to` defined. 
Not sure exactly what you mean. 
Do you want to define all of these in a resource and import then into the routes?

cameron-freshworks: I think in these cases it's ok in the routes but if we were to re use these specific crumbs it would make sense to make it a resource. I just can't think of where else we would reuse at this time.

auth-web/src/App.vue
cameron-freshworks: For sure, this isn't achievable on the app level from what i can see even when placed inside the`app-body`. 
Would require to be about 4 levels down in the Views themselves.. so i'm not sure. Will keep looking.

severinbeauvais: What if you wrap it in a container?

Or move it elsewhere?

cameron-freshworks: So this worked without issue in the other UI's and i think it's due to something locally overriding the `no-gutters` class that the shared component uses. 
Will keep looking in what the root cause of the override is, otherwise i will try to explicitly set. Don't want to change the Shared Component in this instance as i believe it is some local styling. 
👀

cameron-freshworks: OK yea, found it in the `layout` overrides.. basically nullifies any instance of `no-gutters` so i will look into the implications of remove this override.

cameron-freshworks: Removed the override and fixed some elements that were compensating for this previously

auth-web/src/assets/styles/theme.scss
cameron-freshworks: Updated in prior commit

severinbeauvais (approved):

auth-web/src/resources/BreadcrumbResources.ts
cameron-freshworks: To follow up on this, it will be tricky as until each UI has a set configuration object that differs between them. 
ie the filings UI refers to Auth Web home as the `BUSINESS_URL` and auth web refers to itself as `AUTH_WEB_URL` etc. 
So in order to make 1 common resource we would either need..
a) all the configuration objects to follow a standard naming convention
b) have local customization of these resourced objects (which we sort of land us in the same spot we are now anywho)

auth-web/src/routes/router.ts
shabeeb-aot: In someplace, used as breadcrumb: [DashboardBreadcrumb], and other places using only text from resources (text: HomeBreadCrumb.text). So I was thinking to use one pattern. or even if we are using only text, better to add in the language file and use it in both places. (ie in resource text  and HomeBreadCrumb.text). I think its fine now as you mentioned above :)

shabeeb-aot (approved):

auth-web/src/routes/router.ts
cameron-freshworks: Ahh i see you what you mean, absolutely that makes sense.

auth-web/src/assets/styles/theme.scss
```diff
@@ -0,0 +1 @@
+@import "$assets/scss/theme.scss";
```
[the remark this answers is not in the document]
cameron-freshworks: Awesome, yea i look forward to learning more about that approach!

PR #1878, opened 2022-01-14
9517 Registration of SP, GP to lear

vysakh-menon-aot:
*Issue #:* bcgov/entity#9517

*Description of changes:*


auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -155,7 +156,7 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
   // Local Properties
   private readonly businesses!: Business[]
   private readonly currentOrganization!: Organization
-  private readonly createNamedBusiness!: (filingBody: BusinessRequest) => any
+  private readonly createNamedBusiness!: (filingBody: any, nrNumber: string) => Promise<any>
```
cameron-freshworks, line 159: Why remove the typing?

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -192,13 +193,13 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
     return item.corpType.code === CorpType.NEW_BUSINESS && item.name === item.businessIdentifier
   }
 
-  /** Returns true if the affiliation is approved to start an IA */
-  private isApprovedForIA (business: Business): boolean {
+  /** Returns true if the affiliation is approved to start an IA or Registration */
+  private showUseNameRequest (business: Business): boolean {
```
severinbeauvais, line 197: Very small nit, but it's clearer to use "is" or "has" for boolean function names, and "do" or "show" (etc) for action function names.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -358,14 +375,33 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
...
     }
-    const filingResponse = await this.createNamedBusiness(filingBody)
+    return this.createNamedBusiness(filingBody, business.businessIdentifier)
+  }
 
-    if (filingResponse?.errorMsg) {
-      this.$emit('add-unknown-error')
-      return ''
-    } else {
-      return filingResponse.data.filing.business.identifier
```
cameron-freshworks, line 367: Will this still handle a failed creation? 
`createNamedBusiness` will only return an error message it looks like. Looks like it was handled on here previously.

cameron-freshworks: I see, you moved it up 1 level.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -358,14 +375,33 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
...
+  private async createBusinessRegistration (business: Business): Promise<any> {
+    const filingBody: RegistrationRequest = {
+      filing: {
+        header: {
+          name: FilingTypes.REGISTRATION,
+          accountId: this.currentOrganization.id
+        },
+        business: {
+          legalType: business.nameRequest.legalType,
+          natureOfBusiness: business.nameRequest.natureOfBusiness
```
severinbeauvais, line 390: We may or may not need this, right?

I believe this is user-entered text, whereas in grooming today we talked about users entering a NAICS code and we'd look up the NOB. (It would still be text but we could potentially do a reverse lookup on it, whereas a reverse lookup on user-entered text would probably fail.)

I'm just wondering if we need to pre-populate this in the draft. Probably can't hurt though...

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -341,6 +342,22 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
 
   /** Create a business record in Lear */
   private async createBusinessRecord (business: Business): Promise<string> {
+    let filingResponse = null
+    if ([LegalTypes.SP, LegalTypes.GP].includes(business.nameRequest.legalType as LegalTypes)) {
```
cameron-freshworks, line 346: Never a bad idea to optionally chain the deconstruction of an object in the event it's looking for the values before the object itself is available or something.

ie `business.nameRequest?.legalType`

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -358,14 +375,33 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
...
+        },
+        registration: {
+          nameRequest: {
+            legalType: business.nameRequest.legalType,
+            nrNumber: business.businessIdentifier
+          }
+        }
+      }
+    }
+    if (business.nameRequest.legalType === LegalTypes.SP) {
```
severinbeauvais, line 400: Add a comment to what you're doing here.

Should this use a generic method (helper / mixin ?) to return Entity Type Code when given the Legal Type? (Is this used elsewhere? Are we expecting to use this again soon?)

auth-web/src/services/business.services.ts
```diff
@@ -25,11 +25,7 @@ export default class BusinessService {
     return axios.get(`${ConfigHelper.getLegalAPIUrl()}/businesses/${businessIdentifier}`)
   }
 
-  static async createNamedBusiness (filingBody: BusinessRequest): Promise<AxiosResponse<any>> {
-    return axios.post(`${ConfigHelper.getLegalAPIUrl()}/businesses?draft=true`, filingBody)
-  }
-
-  static async createNumberedBusiness (filingBody: BusinessRequest): Promise<AxiosResponse<any>> {
+  static async createDraftFiling (filingBody: any): Promise<AxiosResponse<any>> {
```
cameron-freshworks, line 28: Does the `BusinessRequest` no longer apply? 
If the structure has changed i would create/modify the interface to suite both filings or create new ones.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -358,14 +375,33 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
...
+      filingBody.filing.registration.businessType =
+                    (business.nameRequest.entityTypeCd === 'FR' ? 'SP' : business.nameRequest.entityTypeCd)
```
severinbeauvais, line 402: Should use an enum for Entity Type Code.

See https://github.com/bcgov/namerequest/blob/main/src/enums/entity-type.ts

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -316,11 +316,15 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
...
     this.isLoading = 1 // truthy
     this.$route.query.isNumberedCompanyRequest && await this.createNumberedBusiness(this.currentAccountSettings.id)
-    await this.syncBusinesses()
+    await this.syncBusinesses(this.enableSpGpDba)
     this.lastSyncBusinesses = Date.now()
     this.isLoading = 0 // falsy
   }
 
+  private get enableSpGpDba (): boolean {
+    return LaunchDarklyService.getFlag(LDFlags.EnableSpGpDba) || false
```
severinbeauvais, line 325: 👍

auth-web/src/store/modules/business.ts
```diff
@@ -51,10 +52,19 @@ export default class BusinessModule extends VuexModule {
...
 
+    /** Returns target conditionally. */
+    const getTarget = (nr): NrTargetTypes => {
+      if ([LegalTypes.SP, LegalTypes.GP].includes(nr.legalType) && !enableSpGpDba) {
+        // Fallback to onestop if SpGpDba is not enabled
+        return NrTargetTypes.ONESTOP
+      }
+      return nr.target
+    }
+
```
cameron-freshworks, line 67: Isn't the target be handled by the API? 
Or is this just a temporary means of feature flagging our redirect?

auth-web/src/store/modules/business.ts
```diff
@@ -36,7 +37,7 @@ export default class BusinessModule extends VuexModule {
   }
 
   @Action({ rawError: true })
-  public async syncBusinesses (): Promise<void> {
+  public async syncBusinesses (enableSpGpDba: boolean): Promise<void> {
```
cameron-freshworks, line 40: Is this parameter something we want to require for every call of `syncBusinesses`?
If not, you can set a default here, `syncBusinesses (enableSpGpDba: boolean = false)..` so that is isn't required everywhere.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -316,11 +316,15 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
...
+  }
+
```
cameron-freshworks, line 327: Makes sense that this is a GETTER but if this is temporary means of flagging i think it might be easier to just retrieve this value in the 1 spot you are using it. 
ie `getTarget` method. 
I'm not sure the purpose of passing it as a parameter in the `syncBusinesses` method.
Maybe i'm not seeing something though.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -316,11 +316,15 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
 
     this.isLoading = 1 // truthy
     this.$route.query.isNumberedCompanyRequest && await this.createNumberedBusiness(this.currentAccountSettings.id)
```
severinbeauvais, line 318: Does `createNumberedBusiness()` still exist, or replaced by `createDraftFiling()` ?

See also line 262.

auth-web/src/store/modules/business.ts
```diff
@@ -182,16 +194,15 @@ export default class BusinessModule extends VuexModule {
   }
 
   @Action({ rawError: true })
-  public async createNamedBusiness (filingBody: BusinessRequest) {
+  public async createNamedBusiness (filingBody: any, nrNumber: string) {
```
cameron-freshworks, line 197: We should try to keep these request parameters interfaced.

severinbeauvais (commented):
Are there any unit tests?

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: On the other hand, if you know for sure the object exists (eg, `business` as in Cam's example) then it doesn't need opt chaining. (Good example Cam.)

What I'm saying is we shouldn't use opt chaining on everything just because :)

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: I agree with this -- if it's only used in `syncBusinesses()` then just declare a local variable in there.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
vysakh-menon-aot: To support multiple type of filing. For now we have BusinessRequest and RegistrationRequest. Initially I thought of using `BusinessRequest | RegistrationRequest`. Then I changed it to any to support any future filing since it is only used to post a draft filing

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
vysakh-menon-aot: True. I included this since it is mentioned in Schema.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
vysakh-menon-aot: It may not be required always

auth-web/src/components/auth/manage-business/EntityManagement.vue
vysakh-menon-aot: `createNumberedBusiness` still exist in store. I replaced it in service

auth-web/src/services/business.services.ts
vysakh-menon-aot: It still applicable for IA. filingBody can be BusinessRequest and RegistrationRequest and maybe more in future. Changed it to `any` to support future filing since it is only used to post a draft filing

auth-web/src/store/modules/business.ts
vysakh-menon-aot: It is handled in API. It is temporary until we migrate to lear

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
cameron-freshworks: 🤔  Perhaps we should modify the interfaces then to require the parts common to all filings, with an interchangeable subset for each unique filing?
I think changing the type to `any` is more or less removing typing, more so than supporting something.

auth-web/src/services/business.services.ts
cameron-freshworks: Same comment as above, we should modify or add to our interfaces to include some typing. 
Even if it is just a draft that is not validated. 
Even more reason to include an interface i think since we are not validating, since errors by a developer here wouldn't be caught until the point of file and pay by a client, when the validators do run.

auth-web/src/store/modules/business.ts
cameron-freshworks: Gotcha, thanks!

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: Just thinking ahead a bit... What will our Create UI code do with this pre-populated text? Will the UI prompt for a NAICS code if this text is pre-populated? Will we store the NAICS code or just the NOB?

It may be that leaving this empty is preferable, but I'm not really sure.

If we want the initial Registration draft filing to have an empty NOB, who will clear it? The Filings UI? Now that would be complex (Auth Web creates draft, Filings UI clear a field, Create UI resumes it).

Can you shed any light on this?

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: OK, leave it for now.

severinbeauvais: ... But please add that comment :)

auth-web/src/services/business.services.ts
severinbeauvais: In our framework (TS), type checks are not done at runtime, only compile time. So yes, an interface (or partial or [union types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)) can avoid some errors. However, I'm going to leave it to the author to determine the best approach here at this time.

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: Fixed.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
vysakh-menon-aot: Yes NOB will be populated from the NR if available but we are not using it for now

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
cameron-freshworks: After offline discussion, we determined an acceptable path for the interfaces, will resolve!

auth-web/src/services/business.services.ts
cameron-freshworks: We determined that for the time being we will enforce the base structure of the filing with the subset of the filing type optional. Will resolve!

severinbeauvais (approved):

cameron-freshworks (approved):

shabeeb-aot (approved):

PR #1884, opened 2022-01-20
10554 - handle multiple account for single idir

karthik-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/10554
*Description of changes:*

- Update current account session variable to handle multiple accounts

### DIFF
--- a/auth-web/src/views/auth/AcceptInviteView.vue
+++ b/auth-web/src/views/auth/AcceptInviteView.vue
@@ -38,10 +38,6 @@ export default class AcceptInviteView extends Mixins(NextPageMixin) {
   private readonly getUserProfile!: (identifier: string) => Promise<User>
   private readonly setCurrentOrganization!: (organization: Organization) => void
   private readonly setCurrentMembership!: (membership: Member) => void
-
-  protected readonly userContact!: Contact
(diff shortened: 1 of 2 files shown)


auth-web/src/views/auth/AcceptInviteView.vue
```diff
@@ -97,6 +92,9 @@ export default class AcceptInviteView extends Mixins(NextPageMixin) {
             user: null
           }
           this.setCurrentMembership(membership)
+          // Store the invitation org as acuurentaccount in session for handling multiple accounts in a single IDIR.
+          // Preserves the invitation org while page refresh.
+          ConfigHelper.addToSession(SessionStorageKeys.CurrentAccount, JSON.stringify({ id: invitation.membership[0].org.id, label: invitation.membership[0].org.name }))
```
shabeeb-aot, line 97: If user moves out of the flow, what will happen?

karthik-aot: Removed this approach, now we add it as a variable in child and append it on save

shabeeb-aot (approved):

sumesh-aot (approved):

PR #1893, opened 2022-01-26
Updated Navigations for Edit Profile and staff

cameron-freshworks:
*Issue #:*
https://github.com/bcgov/entity/issues/10751

*Description of changes:*
* Updated redirect on 'Cancel' action from 'Edit Profile' page
* Updated home route when on the MyBusinessRegistry table to be dynamic based on role

### DIFF
--- a/auth-web/src/components/auth/create-account/UserProfileForm.vue
+++ b/auth-web/src/components/auth/create-account/UserProfileForm.vue
@@ -495,7 +495,7 @@ export default class UserProfileForm extends Mixins(NextPageMixin, Steppable) {
 
     private navigateBack (): void {
       if (this.currentOrganization) {
-        this.$router.push(`/account/${this.currentOrganization.id}`)
+        window.location.assign(configHelper.getBcrosDashboardURL())
       } else {
         this.$router.push('/home')
       }
--- a/auth-web/src/routes/router.ts
+++ b/auth-web/src/routes/router.ts
@@ -35,6 +35,7 @@ import GovmAccountCreationSuccessView from '@/views/auth/create-account/GovmAcco
 import GovmAccountSetupView from '@/views/auth/create-account/GovmAccountSetupView.vue'
 import HomeView from '@/views/auth/home/HomeView.vue'
 import IncorpOrRegisterView from '@/views/auth/home/IncorpOrRegisterView.vue'
+import KeyCloakService from 'sbc-common-components/src/services/keycloak.services'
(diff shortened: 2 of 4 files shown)


auth-web/src/routes/router.ts
```diff
@@ -111,6 +112,11 @@ function mapPendingDetails (route: any) {
   }
 }
 
+function isStaff () {
+  const kcUserProfile = KeyCloakService.getUserInfo()
+  return kcUserProfile.roles.includes(Role.Staff)
```
severinbeauvais, line 117: Do these statements always execute correctly or is there a need for a try-catch or conditional chaining?

severinbeauvais (approved):

auth-web/src/routes/router.ts
cameron-freshworks: I haven't encountered an issue yet since this method is called as part of an authenticated 'only' route but the optional chaining would be wise, good call!
Will update

severinbeauvais: If the called code is safe then nothing else is needed. Maybe just have a look that it can't throw an error or return a non-array.

auth-web/src/components/auth/create-account/UserProfileForm.vue
```diff
@@ -495,7 +495,7 @@ export default class UserProfileForm extends Mixins(NextPageMixin, Steppable) {
 
     private navigateBack (): void {
       if (this.currentOrganization) {
-        this.$router.push(`/account/${this.currentOrganization.id}`)
+        window.location.assign(configHelper.getBcrosDashboardURL())
```
karthik-aot, line 498: wouldn't this refresh the page?. Or is it expected to be that way :)

karthik-aot (approved):

auth-web/src/routes/router.ts
cameron-freshworks: Definitely, will do.
With the added optional chaining in the event of a failed request on `KeyCloakService` the route would resolve to the default Dashboard Breadcrumb.

auth-web/src/components/auth/create-account/UserProfileForm.vue
cameron-freshworks: It would yes, since we are navigating back to our Uber Dashboard (higher level page) and not a local route.
Is there an alternative and/or better approach when redirecting?

PR #1909, opened 2022-02-11
11134 moving business object into registration schema

vysakh-menon-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/11134

*Description of changes:*

### DIFF
--- a/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
+++ b/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
@@ -398,13 +398,15 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
           accountId: this.currentOrganization.id
         },
         business: {
-          legalType: business.nameRequest.legalType,
-          natureOfBusiness: business.nameRequest.natureOfBusiness
+          legalType: business.nameRequest.legalType
         },
         registration: {
           nameRequest: {
             legalType: business.nameRequest.legalType,
             nrNumber: business.businessIdentifier
+          },
+          business: {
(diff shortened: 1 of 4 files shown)


auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -397,14 +397,13 @@ export default class AffiliatedEntityTable extends Mixins(DateMixin) {
           name: FilingTypes.REGISTRATION,
           accountId: this.currentOrganization.id
         },
-        business: {
-          legalType: business.nameRequest.legalType,
```
severinbeauvais, line 401: Don't we want to keep legalType here? I think some of our other UIs look for legal type here.

auth-web/src/models/business.ts
```diff
@@ -94,16 +94,18 @@ export interface BusinessRequest {
             name: string,
             accountId: number
         },
-        business: {
+        business?: {
```
severinbeauvais, line 97: When is this optional?

vysakh-menon-aot: Made it optional to support IA filing (which still use business object) for now.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: > As per thor we should not use business object here since the business is not created yet

Make sense. Thanks.

auth-web/src/models/business.ts
severinbeauvais: Please add a comment saying just that.

severinbeauvais (approved):

PR #1931, opened 2022-03-14
10012_nr_affliation_emailcase

ozamani9:
*Issue #:*
https://github.com/bcgov/entity/issues/10012

*Description of changes:*

add casefold function to both the applicants email and the verification email in business affiliation logic.

### DIFF
--- a/auth-api/src/auth_api/version.py
+++ b/auth-api/src/auth_api/version.py
@@ -22,4 +22,4 @@ Post-release segment: .postN
 Development release segment: .devN
 """
 
-__version__ = '0.1.0a0.dev'  # pylint: disable=invalid-name
+__version__ = '0.2.0a0'  # pylint: disable=invalid-name
--- a/auth-api/src/auth_api/version.py
+++ b/auth-api/src/auth_api/version.py
@@ -22,4 +22,4 @@ Post-release segment: .postN
 Development release segment: .devN
 """
 
-__version__ = '0.2.0a0'  # pylint: disable=invalid-name
+__version__ = '0.1.0a0.dev'  # pylint: disable=invalid-name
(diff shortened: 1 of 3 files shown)


auth-api/src/auth_api/version.py
```diff
@@ -22,4 +22,4 @@
 Development release segment: .devN
 """
 
-__version__ = '0.1.0a0.dev'  # pylint: disable=invalid-name
+__version__ = '0.2.0a0'  # pylint: disable=invalid-name
```
saravanpa-aot, line 25: I am not sure if we need to do this.. any particular reason for doing this?

auth-api/tests/unit/services/test_affiliation.py
```diff
@@ -344,7 +344,7 @@ def test_create_new_business(session, auth_mock, nr_mock):  # pylint:disable=unu
     business_identifier = 'NR 1234567'
 
     affiliation = AffiliationService.create_new_business_affiliation(org_id, business_identifier=business_identifier,
-                                                                     email='test@test.com', phone='1112223333')
+                                                                     email='TEST@test.com', phone='1112223333')
```
saravanpa-aot, line 347: if you want , you could a different test case with both small case and capital.. that will be more clear and helpful

ozamani9: ok i will create two tests.

auth-api/src/auth_api/version.py
saravanpa-aot: Yeah...I am aware of that..
We release a bit different from entities ;mostly big release we do with bunch of tickets together..We don't generally rollout one change into prod quickly..mostly big releases...So we don't have this practice so far..but i am okay to bring in this practice ...May be we can discuss with @pwei1018 and agree on it..

ozamani9: fixed version and a lint error and repushing now

pwei1018 (approved):

PR #1954, opened 2022-05-10
11833 - Updated auth-web major release number

severinbeauvais:
*Issue #:* https://github.com/bcgov/entity/issues/11883

*Description of changes:*
- updated app major version since Node version (and packag-lock format) was just updated


severinbeauvais:
As we did with the other UIs where Node was updated, I bumped the app main version number (as this is a significant change). This should have been done in the previous PR but it was merged before I could comment on it.

seeker25 (approved):

PR #2041, opened 2022-07-20
12954 Resubmit BN request

vysakh-menon-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/12954

*Description of changes:*


seeker25:
Does this need to go out with the release tomorrow?

auth-web/src/store/modules/business.ts
```diff
@@ -218,20 +218,45 @@ export default class BusinessModule extends VuexModule {
 
   // Following searchBusiness will search data from legal-api.
   @Action({ rawError: true })
-  public async searchBusiness (businessIdentifier: string): Promise<any> {
+  public async searchBusiness (businessIdentifier: string): Promise<LearBusiness> {
     const response = await BusinessService.searchBusiness(businessIdentifier)
```
severinbeauvais, line 222: `searchBusiness` will throw an error if something went wrong. You should either add a `.catch()` call to it or wrap this in a try-catch block.

You could also do this but it doesn't really handle the error:
```
const response = await BusinessService.searchBusiness(businessIdentifier).catch(() => {})
```

severinbeauvais: Same for the other business service calls (lines 238, 247 and 253).

At the moment, all these calls will crash the JS thread if there's a network error. The exception must be caught.

severinbeauvais: You could also catch these errors in BusinessService and return null. Then THIS code might be OK.

auth-web/src/components/auth/staff/admin/AdministrativeBN.vue
```diff
@@ -0,0 +1,292 @@
+<template>
+  <v-card class="pa-8" flat style="height: 100%;">
```
severinbeauvais, line 2: I recommend giving this component an id (right here at the top level).

auth-web/src/components/auth/staff/admin/AdministrativeBN.vue
```diff
@@ -0,0 +1,292 @@
...
+
+const BusinessModule = namespace('business')
+
+@Component({
+  components: {
+    BNRequestManager
+  },
+  directives: { mask }
+})
+export default class AdministrativeBN extends Vue {
```
severinbeauvais, line 153: Is this component exactly what you took out of AdminDashboardView.vue ? (I'm hoping to not compare it line by line.)

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
```diff
@@ -0,0 +1,143 @@
...
+  protected requestDetails: RequestTracker = null
+  protected resubmitRequestDialog = false
+
+  @Watch('businessIdentifier', { deep: true, immediate: true })
+  async businessIdentifierChange () {
+    this.bnRequests = await this.getBNRequests(this.businessIdentifier)
+  }
+
+  private async showResubmitRequestDialog (item: RequestTracker): Promise<void> {
+    this.requestDetails = await this.getRequestTracker(item.id)
```
severinbeauvais, line 98: As in the other component, `getRequestTracker()` should either return null if it encounters an error, or you need to handle it here.

I noticed that line 5 de-references "requestDetails" so it should be updated to handle null.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
```diff
@@ -0,0 +1,143 @@
...
+    this.resubmitRequestDialog = true
+  }
+
+  private async hideResubmitRequestDialog (): Promise<void> {
+    this.resubmitRequestDialog = false
+    this.requestDetails = null
+  }
+
+  private async resubmitRequest (xmlData): Promise<void> {
+    const queued = await this.resubmitBNRequest({
```
severinbeauvais, line 108: As above, may throw an exception.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
```diff
@@ -0,0 +1,143 @@
...
+}
+
+.action-cell {
+  max-width: 0;
+  max-height: 30px !important;
+  text-align: end !important;
+  padding-right: 5;
+}
+
+::v-deep {
```
severinbeauvais, line 138: Please add a comment saying what/why this CSS is here.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/scss/theme.scss';
+
+.table-header {
+  display: flex;
```
severinbeauvais, line 126: If you prefer, you can use class `d-flex` in the element instead.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
```diff
@@ -0,0 +1,143 @@
...
+
+  @BusinessModule.Action('getRequestTracker')
+  private readonly getRequestTracker!: (requestTrackerId: number) => Promise<RequestTracker>
+
+  @BusinessModule.Action('resubmitBNRequest')
+  private readonly resubmitBNRequest!: (resubmitBNRequest: ResubmitBNRequest) => Promise<boolean>
+
+  @Prop({ default: undefined }) businessIdentifier: string
+
+  protected headers = [
```
severinbeauvais, line 81: I recommend using readonly for static data like this.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
```diff
@@ -0,0 +1,143 @@
...
+    { text: 'Actions', align: 'end', value: 'action', sortable: false, show: true }
+  ]
+
+  protected bnRequests: RequestTracker[] = []
+  protected requestDetails: RequestTracker = null
+  protected resubmitRequestDialog = false
+
+  @Watch('businessIdentifier', { deep: true, immediate: true })
+  async businessIdentifierChange () {
+    this.bnRequests = await this.getBNRequests(this.businessIdentifier)
```
severinbeauvais, line 94: May throw exception.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
```diff
@@ -0,0 +1,143 @@
...
+  private async resubmitRequest (xmlData): Promise<void> {
+    const queued = await this.resubmitBNRequest({
+      businessIdentifier: this.businessIdentifier,
+      requestType: this.requestDetails.requestType,
+      request: xmlData
+    })
+    if (queued) {
+      this.resubmitRequestDialog = false
+      this.requestDetails = null
+      this.businessIdentifierChange()
```
severinbeauvais, line 116: As the method is async, you should await it here.

auth-web/src/components/auth/staff/admin/ResubmitRequestDialog.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <v-dialog
+    v-model="dialog"
+    persistent
+    :attach="attach"
+    content-class="resubmit-request-dialog"
+  >
+    <v-card flat id="resubmit-request-container">
```
severinbeauvais, line 8: This id is not needed as the dialog itself can be selected with `.resubmit-request-dialog`.

auth-web/src/components/auth/staff/admin/ResubmitRequestDialog.vue
```diff
@@ -0,0 +1,86 @@
...
+  @Prop() readonly attach: string
+
+  @Prop({ default: '' })
+  readonly xmlData: string
+
+  @Prop({ default: false })
+  readonly submitActive: boolean
+
+  private save (): void {
+    this.resubmit(this.xmlData)
```
severinbeauvais, line 67: I recommend combining `save()` and `resubmit()` below.

```
@Emit('resubmit')
private resubmit (): string {
  return this.xmlData
}
```

auth-web/src/components/auth/staff/admin/AdministrativeBN.vue
vysakh-menon-aot: Yes

auth-web/src/store/modules/business.ts
vysakh-menon-aot: Will do

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
vysakh-menon-aot: I have already included `v-if="requestDetails"` in line 3

severinbeauvais: OK, this seems to work because you initialize requestDetails to null and you also reset it when hiding the dialog.

It's good practise to handle exceptions from network calls anyway. Since you don't, here, line 98 will cause the JS thread to terminate, so line 99 won't get run -- the dialog won't display if there was a network error. This is a bug.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
severinbeauvais: `d-flex` is a Vuetify class, not a directive/property.

auth-web/src/components/auth/staff/admin/BNRequestManager.vue
severinbeauvais: Perfect!

auth-web/src/store/modules/business.ts
```diff
@@ -218,20 +218,45 @@ export default class BusinessModule extends VuexModule {
...
+    if (response && response?.status === 200) {
+      return response.data.requestTrackers
+    }
+    return []
+  }
+
+  @Action({ rawError: true })
+  public async resubmitBNRequest (resubmitRequest: ResubmitBNRequest): Promise<any> {
+    const response = await BusinessService.resubmitBNRequest(resubmitRequest).catch(() => null)
+    return response && (response?.status === 200 || response?.status === 201)
```
severinbeauvais, line 248: This can be simplified with conditional chaining.

Does the caller handle the different return types (could be True, False or Null) ?

auth-web/src/store/modules/business.ts
```diff
@@ -218,20 +218,45 @@ export default class BusinessModule extends VuexModule {
...
+  }
+
+  @Action({ rawError: true })
+  public async getRequestTracker (requestTrackerId: number): Promise<RequestTracker> {
+    const response = await BusinessService.getRequestTracker(requestTrackerId).catch(() => null)
+    if (response && (response?.status === 200)) {
```
severinbeauvais, line 254: This can be simplified with conditional chaining.

auth-web/src/store/modules/business.ts
```diff
@@ -218,20 +218,45 @@ export default class BusinessModule extends VuexModule {
...
 
   @Action({ rawError: true })
   public async createBNRequest (request: BNRequest): Promise<any> {
     return BusinessService.createBNRequest(request)
   }
 
+  @Action({ rawError: true })
+  public async getBNRequests (businessIdentifier: string): Promise<RequestTracker[]> {
+    const response = await BusinessService.getBNRequests(businessIdentifier).catch(() => null)
+    if (response && response?.status === 200) {
```
severinbeauvais, line 239: This can be simplified with conditional chaining.

auth-web/src/store/modules/business.ts
```diff
@@ -218,20 +218,45 @@ export default class BusinessModule extends VuexModule {
 
   // Following searchBusiness will search data from legal-api.
   @Action({ rawError: true })
-  public async searchBusiness (businessIdentifier: string): Promise<any> {
-    const response = await BusinessService.searchBusiness(businessIdentifier)
-    if (response && response.data && response.status === 200) {
+  public async searchBusiness (businessIdentifier: string): Promise<LearBusiness> {
+    const response = await BusinessService.searchBusiness(businessIdentifier).catch(() => null)
+    if (response && response?.status === 200 && response?.data?.business?.legalName) {
```
severinbeauvais, line 223: This can be simplified with conditional chaining.

severinbeauvais (approved):
Looks good.

auth-web/src/components/auth/staff/admin/AdministrativeBN.vue
```diff
@@ -0,0 +1,292 @@
+<template>
+  <v-card id="administrative-bn" class="pa-8" flat style="height: 100%;">
```
shabeeb-aot, line 2: please move please move inline style to class

shabeeb-aot (approved):

PR #2058, opened 2022-08-16
13154 Updated sbc-common-components import

severinbeauvais:
*Issue #:* bcgov/entity#13154

*Description of changes:*
- updated sbc-common-components import to 3.0.1
- app version = 1.1.0

### DIFF
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "1.0.19",
+  "version": "1.1.0",
   "private": true,
   "scripts": {
     "serve": "vue-cli-service serve",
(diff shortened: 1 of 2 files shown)


auth-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "1.0.19",
+  "version": "1.1.0",
```
severinbeauvais, line 3: Since the sbc-common-components is a major version jump (even though its changes are minimal), and since "19" is a pretty large patch number, I increased the  minor version here.

severinbeauvais:
Please review and merge if OK. Thanks!

seeker25 (approved):

auth-web/package.json
seeker25: Yeah I didn't want to bump sbc-common-components a major version. Faiz wanted to be consistent with other projects I believe.

PR #2056, opened 2022-08-15
13253 Auth Web - Update Marketing Page Buttons

lewischenstudio:
*Issue #:*
https://github.com/bcgov/entity/13253

*Description of changes:*
- update auth web buttons according to the [new design UI](https://projects.invisionapp.com/share/5C12X9U7RUJT#/screens/468822174)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the sbc-auth license (Apache 2.0).

PR #2059, opened 2022-08-17
13251 Auth Web - Move buttons

lambert-alex:
*Issue #:* /bcgov/entity#13251

*Description of changes:*
- Move incorporate numbered company from business marketing page, btn from HomeView in [13253](https://github.com/bcgov/sbc-auth/pull/2056)
- Add tooltips for "incorporate numbered company" and "Add an Existing Business or NR" btns
- typo correction
- unit tests

### DIFF
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "1.0.19",
+  "version": "1.0.20",
   "private": true,
   "scripts": {
     "serve": "vue-cli-service serve",
--- a/auth-web/src/components/auth/manage-business/EntityManagement.vue
+++ b/auth-web/src/components/auth/manage-business/EntityManagement.vue
@@ -6,7 +6,7 @@
       </div>
     </v-fade-transition>
 
-    <v-container class="view-container">
+    <v-container justify class="view-container">
       <div class="view-header align-center">
         <h1 class="view-header__title">My Business Registry<br>
           <span class="subtitle">{{ $t('myBusinessDashSubtitle') }}</span>
@@ -26,7 +26,7 @@
       </div>
 
       <v-row no-gutters id="dashboard-actions" class="mb-n3">
-        <v-flex shrink class="mr-4">
+        <v-col cols="auto">
           <!-- Add Existing Name Request or Business -->
           <v-menu
             v-model="addAffiliationDropdown"
@@ -36,7 +36,7 @@
                 <template v-slot:activator="{ on: onTooltip }">
                   <v-btn
                     id="add-existing-btn"
-                    class="mt-2"
+                    class="mt-2 mr-4"
                     color="primary"
                     dark large
                     v-on="{ ...onMenu, ...onTooltip }"
@@ -61,13 +61,13 @@
               <v-list-item class="add-existing-item" @click="showAddNRModal()">Name Request</v-list-item>
             </v-list>
           </v-menu>
-        </v-flex>
-        <v-flex shrink>
+        </v-col>
+        <v-col cols="auto">
           <v-tooltip top max-width="361px" content-class="top-tooltip">
             <template v-slot:activator="{ on, attrs }">
               <v-btn
                 id="incorporate-numbered-btn"
-                class="mt-2"
+                class="mt-2 mr-4"
                 color="primary"
                 outlined dark large
                 v-bind="attrs"
@@ -80,8 +80,8 @@
             </template>
             <span>Start an incorporation application for a numbered benefit company in B.C.</span>
           </v-tooltip>
-        </v-flex>
-        <v-flex>
+        </v-col>
+        <v-col>
           <v-select
             dense filled multiple
             class="column-selector"
@@ -92,7 +92,7 @@
           >
             <template v-slot:selection></template>
           </v-select>
-        </v-flex>
+        </v-col>
       </v-row>
 
       <AffiliatedEntityTable
@@ -337,7 +337,7 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
   }
 
   // create a numbered company
-  async startNumberedCompany () {
+  protected async startNumberedCompany () {
     await this.createNumberedBusiness(this.currentAccountSettings.id)
     await this.syncBusinesses()
   }
--- a/auth-web/src/components/auth/manage-business/EntityManagement.vue
+++ b/auth-web/src/components/auth/manage-business/EntityManagement.vue
@@ -26,24 +26,29 @@
       </div>
 
       <v-row no-gutters id="dashboard-actions" class="mb-n3">
-        <v-col cols="9">
+        <v-flex shrink class="mr-4">
           <!-- Add Existing Name Request or Business -->
           <v-menu
             v-model="addAffiliationDropdown"
           >
-            <template v-slot:activator="{ on }">
-              <v-btn
-                id="add-existing-btn"
-                class="mt-2"
-                color="primary"
-                dark
-                large
-                v-on="on"
-              >
-                <v-icon>mdi-plus</v-icon>
-                <span><strong>Add an Existing Business or Name Request</strong></span>
-                <v-icon class="ml-2 mr-n2">{{ addAffiliationDropdown ? 'mdi-menu-up' : 'mdi-menu-down' }}</v-icon>
-              </v-btn>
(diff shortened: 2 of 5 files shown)


auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -26,24 +26,29 @@
       </div>
 
       <v-row no-gutters id="dashboard-actions" class="mb-n3">
-        <v-col cols="9">
+        <v-flex shrink class="mr-4">
```
lambert-alex, line 29: The v-flex made for much nicer dynamic page layout when the page size gets changed

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -312,6 +336,12 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
     window.location.href = appendAccountId(ConfigHelper.getNameRequestUrl())
   }
 
+  // create a numbered company
+  async startNumberedCompany () {
+    await this.createNumberedBusiness(this.currentAccountSettings.id)
```
lambert-alex, line 341: The source of the 10 numbered ben drafts

auth-web/tests/unit/components/EntityManagement.spec.ts
```diff
@@ -114,4 +114,19 @@ describe('Entity Management Component', () => {
...
+  it('all buttons, tooltips and v-menu selections exist', async () => {
+    // All buttons exist
+    expect(wrapper.find('#add-existing-btn').exists()).toBe(true)
+    expect(wrapper.find('#add-name-request-btn').exists()).toBe(true)
+    expect(wrapper.find('#incorporate-numbered-btn').exists()).toBe(true)
+
+    // Existing Business or NameRequest menu selections
+    wrapper.find('#add-existing-btn').trigger('click')
+    await Vue.nextTick()
+    expect(wrapper.findAll('.add-existing-item').length).toBe(2)
```
lambert-alex, line 127: findAll is depreciating but only when used on components it is still good to use on classes

auth-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "1.0.19",
+  "version": "1.0.20",
```
severinbeauvais, line 3: This project is up to 1.1.0 already (and there's another PR taking it to 1.1.1).

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: Don't use v-flex. [It's deprecated.](https://vuetifyjs.com/en/components/grids/#caveats)).

You can specify responsive  props for v-cols to make them look nice when the page size gets changed. [example](https://github.com/bcgov/business-create-ui/blob/7d29ed3ded310ac4a47e7de08ca53b9bacb69b6f/src/App.vue#L128)

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -26,24 +26,29 @@
...
-              </v-btn>
+            <template v-slot:activator="{ on: onMenu }">
+              <v-tooltip top content-class="top-tooltip">
+                <template v-slot:activator="{ on: onTooltip }">
+                  <v-btn
+                    id="add-existing-btn"
+                    class="mt-2"
+                    color="primary"
+                    dark large
+                    v-on="{ ...onMenu, ...onTooltip }"
```
severinbeauvais, line 42: I think you're the first dev to use v-on like this. What does it do?

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: This should be protected unless it's desirable and safe to call from outside this class.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -492,6 +522,37 @@ export default class EntityManagement extends Mixins(AccountChangeMixin, NextPag
...
+
+.top-tooltip:after {
+  top: 100% !important;
+  left: 45% !important;
+  margin-top: 0 !important;
+  border-right: 10px solid transparent !important;
+  border-left: 10px solid transparent !important;
+  border-top: 8px solid RGBA(73, 80, 87, 0.95) !important;
+}
+
```
severinbeauvais, line 555: Let's keep an eye on all this CSS. We might want to make it global (by adding to base.scss) if it applies to all tooltips, which it probably should but we may not be ready for that (needs UI/UX input).

severinbeauvais (commented):
I don't see where you "moved" a button; only where you added one. Should this PR also remove it from its original location?

auth-web/src/components/auth/manage-business/EntityManagement.vue
lambert-alex: Done!

auth-web/src/components/auth/manage-business/EntityManagement.vue
lambert-alex: Updated I don't think there is any reason why it would need to be called from outside this class

severinbeauvais: My comment is resolved (but I didn't start this conversation so I won't resolve it).

auth-web/src/components/auth/manage-business/EntityManagement.vue
lambert-alex: The multiple activator slots working on v-btn cause a conflict on v-on to apply the tooltip event handler and the menu event handler. So on line 34 and line 36 I assigned them different names  and then ```{ ...onMenu, ...onTooltip }``` stiches them together so they both apply to the btn. [Stack post](https://stackoverflow.com/questions/58075400/using-v-tooltip-inside-v-menu-activator-in-vuetify-2-0) where I got it. It has a good explination.

auth-web/src/components/auth/manage-business/EntityManagement.vue
lambert-alex: I agree, I think there are 6 separate instances where this or similar code is being used. I think the biggest problem is the arrows and if the tooltip appears on the side or the top ect

lambert-alex:
> I don't see where you "moved" a button; only where you added one. Should this PR also remove it from its original location?

It was removed by the PR that Lewis did. I updated the PR description to indicate this.

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: Well, some tooltips are "top-tooltip" and others are "bottom-tooltip" so they'll have different classes.

severinbeauvais: Could the CSS be made global now (easily)?

severinbeauvais (approved):
LGTM

Up to you whether to globalize the tooltip CSS now. If not, please create a tech debt ticket for that.

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais: So the tooltip displays on both the dropdown menu icon and the Add button?

lewischenstudio (approved):

auth-web/src/components/auth/manage-business/EntityManagement.vue
lambert-alex: I don't think so there isn't consistency across the css, this maybe should be made into a tech debt ticket.

seeker25:
Looks great (from test):
image

seeker25:
One issue clicking on the button: (I'm logged in as staff)

image

seeker25:
@lewischenstudio Tagging you

lambert-alex:
> One issue clicking on the button Incorporate a Number Benefit Company: (I'm logged in as staff)
> 
> image
> 
> This is probably related to the fact that Staff doesn't have an account id? We'll be fixing this in the near future hopefully.
> 
> Not sure what you'd like to do for Staff now though.

The issue has now been fixed in [13370](https://github.com/bcgov/sbc-auth/pull/2060) . Just commenting for posterity.

PR #2076, opened 2022-09-15
Convert from options-api / vue-class-components -> composition api

severinbeauvais:
*Issue #:*

*Description of changes:*

> Hey guys, 
> I found this tool:
> https://vue-composition-converter.vercel.app/
> Managed to look at the source.. 
> Got it working in node:
> ...
> It’s still missing the template code.. I could figure out a way to read this from the file and insert it using node
> The code formatting gets turfed.. (used prettier here) – Not sure if we can do much about this
> It isn’t script setup.. probably would need to manually convert it, if we wanted to take that route
> I haven’t tested it.. 
> I don’t know if it will compile..
> Not sure if it’s worth looking at this path?
> Thanks,
> Travis


seeker25:
Just to make sure this is all in one spot:

Potential solutions:
https://github.com/yoyo930021/vc2c
https://github.com/HearTao/class-component-converter

seeker25:
Closing, can reopen in the future

PR #2091, opened 2022-09-27
13612 - added LaunchDarkly to sbc-auth

MatthewCai2002:
*Issue #: 13612*
https://github.com/bcgov/entity/issues/13612

*Description of changes:*
    - added LDclient to auth-api with initialization and setup code for flag service
    - added unit tests to test initialization of flag service
    - added launchdarkly-server-sdk to requirements.txt 


auth-api/tests/unit/services/test_flags.py
```diff
@@ -0,0 +1,94 @@
+# Copyright © 2019 Province of British Columbia
```
seeker25, line 1: 2019 haha

auth-api/src/auth_api/services/flags.py
```diff
@@ -0,0 +1,146 @@
+# Copyright © 2019 Province of British Columbia
```
seeker25, line 1: 2019

auth-api/flags.json
```diff
@@ -0,0 +1,7 @@
+{
+    "flagValues": {
```
seeker25, line 2: is this needed?

seeker25 (changes requested):
Nice work, we need this added to the side projects as well, specifically the queue_services. 

@pwei1018 Any projects in `AUTH` you'd like LD added to as well? `status-api`? `notify-api`?

pwei1018:
> Nice work, we need this added to the side projects as well, specifically the `queue_services`.
> 
> @pwei1018 Any projects in `AUTH` you'd like LD added to as well? `status-api`? `notify-api`?

Don't need to add the LD to status-api and notify-api for now.

auth-api/src/auth_api/config.py
```diff
@@ -59,6 +59,7 @@ class _Config:  # pylint: disable=too-few-public-methods
     PROJECT_ROOT = os.path.abspath(os.path.dirname(__file__))
 
     SECRET_KEY = 'a secret'
+    LD_SDK_KEY = os.getenv('LD_SDK_KEY', None)
```
pwei1018, line 62: Please change the environment variable name LD_SDK_KEY to AUTH_LD_SDK_KEY. We have multiple keys for different applications.

pwei1018 (changes requested):

auth-api/flags.json
MatthewCai2002: I think in LEAR It's mainly used for loading in some dummy flags from a local file when we're initializing the flag service without an SDK key while app.env isn't production.

I thought it might be useful in auth too so I pulled it over, Although it's not doing anything right now.

thorwolpert: It can be used for testing purposes.
If you want to run unit or other tests, and as a dev you don't have access to change the LD environment, you can use the local file to exercise the flag values.

thorwolpert (commented):
Just a few comments & suggestions to consider.

auth-api/requirements/prod.txt
```diff
@@ -28,3 +28,4 @@ asyncio-nats-streaming
 sqlalchemy<1.4
 markupsafe==2.0.1
 protobuf~=3.19.0
+launchdarkly-server-sdk==7.5.0
```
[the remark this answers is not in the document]
MatthewCai2002: what about the other fields with version numbers attached? Should I also consider removing those?

auth-api/src/auth_api/__init__.py
```diff
@@ -57,6 +58,7 @@ def create_app(run_mode=os.getenv('FLASK_ENV', 'production')):
     db.init_app(app)
     ma.init_app(app)
     mail.init_app(app)
+    flags.init_app(app)
```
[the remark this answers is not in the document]
MatthewCai2002: Yup can do

auth-api/src/auth_api/config.py
MatthewCai2002: Yup made the change

auth-api/requirements/prod.txt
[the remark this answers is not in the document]
seeker25: no they're hard stuck for a reason

MatthewCai2002:
removed exception handling for is_on and value because linter kept flagging it with a "too broad exception" and I couldn't find documentation on whether an exception was thrown at all with the client.variation() function nor what kind of exception was being thrown.

auth-api/src/auth_api/services/flags.py
```diff
@@ -0,0 +1,139 @@
...
+from ldclient.config import Config  # noqa: I005
+from ldclient.integrations import Files
+
+from auth_api.models import User
+
+# from msilib.schema import File
+# from ldclient.impl.integrations.files.file_data_source import _FileDataSource
+# from ldclient.interfaces import UpdateProcessor
+
+# class FileDataSource(UpdateProcessor):
```
seeker25, line 26: Code smell, can you remove this code?

MatthewCai2002: yup

auth-api/tests/unit/services/test_flags.py
```diff
@@ -0,0 +1,93 @@
...
+    app.env = 'production'
+    app.config['AUTH_LD_SDK_KEY'] = None
+
+    with app.app_context():
+        flags = Flags()
+        flags.init_app(app)
+        with pytest.raises(KeyError):
+            client = app.extensions['featureflags']
+            assert not client
+    assert flags
```
kialj876, line 93: Could you add in some tests here to assert the flag values?

In https://github.com/bcgov/registries-search we also have some integration test examples (you need the LD flag set for them) which would be great to have. They can be pretty useful when experimenting with new flags especially when adding segments for specific users, etc.

kialj876: Was just looking at them and they don't test much haha, but they're a good stepping stone for flag tests once real flags are added to the api and targeting specific users

MatthewCai2002: yup added

seeker25 (approved):

seeker25:
Need to update `auth-api\vaults.json` before merging

seeker25:
@pwei1018 Would this go under vault `relationship` or vault `api`?

kialj876 (approved):
nice work!

PR #2102, opened 2022-10-06
13248 - Auth ORG changes

seeker25:
- Revert service level blocking
- Implement model level blocking (thanks Kial)
- Allow Staff and SBC Staff to add products
- Transaction and Statements needed a touchup in Auth-web
- Payment type check fix

*Issue #:*
https://github.com/bcgov/entity/issues/13248

### DIFF
--- a/auth-api/src/auth_api/models/org.py
+++ b/auth-api/src/auth_api/models/org.py
@@ -16,12 +16,16 @@
 Basic users will have an internal Org that is not created explicitly, but implicitly upon User account creation.
 """
 from flask import current_app
-from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Integer, String, and_, cast, func
+from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Integer, String, and_, cast, event, func
 from sqlalchemy.orm import contains_eager, relationship
+
+from auth_api.exceptions import BusinessException
+from auth_api.exceptions.errors import Error
 from auth_api.models.affiliation import Affiliation
 from auth_api.models.dataclass import OrgSearch
 from auth_api.utils.enums import AccessType, InvitationStatus, InvitationType
 from auth_api.utils.enums import OrgStatus as OrgStatusEnum
+from auth_api.utils.enums import OrgType as OrgTypeEnum
 from auth_api.utils.roles import EXCLUDED_FIELDS, VALID_STATUSES
 
 from .base_model import VersionedModel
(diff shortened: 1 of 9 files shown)


auth-api/src/auth_api/models/org.py
```diff
@@ -240,3 +244,25 @@ def reset(self):
...
+    """Rejects invalid type_codes on insert."""
+    org = target
+    if org.type_code in (OrgTypeEnum.SBC_STAFF.value, OrgTypeEnum.STAFF.value):
+        raise BusinessException(
+            Error.INVALID_INPUT,
+            None
+        )
+
+
+@event.listens_for(Org, 'before_update', raw=True)
```
kialj876, line 260: you can stack this with the 'before_insert' so that you don't have duplicate code unless there's a reason you want to separate these

auth-api/tests/unit/services/test_org.py
```diff
@@ -403,6 +405,54 @@ def test_create_product_multiple_subscription(session, keycloak_mock, monkeypatc
...
+                                                               TestOrgProductsInfo.org_products2,
+                                                               skip_auth=True)
+
+    assert next(prod for prod in subscriptions
+                if prod.get('code') == TestOrgProductsInfo.org_products2['subscriptions'][0]['productCode'])
+    assert next(prod for prod in subscriptions
+                if prod.get('code') == TestOrgProductsInfo.org_products2['subscriptions'][1]['productCode'])
+
+
+def test_create_product_subscription_sbc_staff(session, keycloak_mock, monkeypatch):
```
kialj876, line 432: to prevent the duplicate code in here can you add this test case to the above one and pass in the org type as a param?

kialj876 (changes requested):
Hey Travis this looks great just a couple minor things I noticed. Can you also add a test case to verify that updating the org type to sbc staff / staff fails?

auth-api/src/auth_api/models/org.py
seeker25: `raw=True` is the reason why they're separate - changes the object "state"

auth-api/tests/unit/services/test_org.py
seeker25: Yup

auth-api/src/auth_api/models/org.py
seeker25: They have to be seperate

auth-api/tests/unit/services/test_org.py
```diff
@@ -403,6 +405,54 @@ def test_create_product_multiple_subscription(session, keycloak_mock, monkeypatc
...
+    """Assert that updating product subscription works for sbc staff."""
+    user = factory_user_model(TestUserInfo.user_test)
+    patch_token_info({'sub': user.keycloak_guid}, monkeypatch)
+    org = OrgService.create_org(TestOrgInfo.org1, user_id=user.id)
+
+    # Clearing the event listeners here, because we can't change the type_code.
+    clear_event_listeners(OrgModel)
+    org_db = OrgModel.find_by_id(org._model.id)
+    org_db.type_code = OrgType.SBC_STAFF.value
```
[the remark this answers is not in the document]
seeker25: Because there are previous tests that already test `create_org`

kialj876 (approved):
looks good!

PR #2105, opened 2022-10-11
13861 - Update SBC-Auth to SBC-common-components 3.0.3 to support BC Reg Staff.

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/13861

*Description of changes:*
Update sbc-common-components

### DIFF
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -30,7 +30,7 @@
     "moment": "^2.29.4",
     "regenerator-runtime": "^0.13.3",
     "register-service-worker": "^1.6.2",
-    "sbc-common-components": "3.0.1",
+    "sbc-common-components": "3.0.3",
     "vue": "2.6.12",
     "vue-i18n": "^8.0.0",
     "vue-plugin-helper-decorator": "0.0.12",
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "1.1.9",
+  "version": "1.1.10",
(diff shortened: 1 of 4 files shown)


seeker25:
image

auth-web/package.json
```diff
@@ -30,7 +30,7 @@
     "moment": "^2.29.4",
     "regenerator-runtime": "^0.13.3",
     "register-service-worker": "^1.6.2",
-    "sbc-common-components": "3.0.1",
+    "sbc-common-components": "3.0.3",
```
severinbeauvais, line 33: Maybe bump up line 3 as well (even though that version is not exposed yet).

auth-web/src/App.vue
```diff
@@ -3,6 +3,7 @@
     <div class="header-group" ref="headerGroup">
       <sbc-loader :show="showLoading" />
       <sbc-header
+        class="flex-direction-column"
```
severinbeauvais, line 6: Would one of these work instead?

https://vuetifyjs.com/en/styles/flex/#flex-direction

severinbeauvais (approved):
Whatever you decide in OK by me.

auth-web/src/App.vue
seeker25: .flex-column

seeker25: let me bookmark this page -^

auth-web/package.json
seeker25: should be good

PR #2110, opened 2022-10-13
imported LD flag service into queue_services

MatthewCai2002:
*Issue #: 13846*
https://github.com/bcgov/entity/issues/13846

*Description of changes:*
   - imported LD flag service from auth_api.services into account-mailer, activity-log-listener, business-events-listener, and 
     events listener worker files in queue_services
   - updated requirements, configs, and vaults for each respective queue service

### DIFF
--- a/queue_services/events-listener/src/events_listener/worker.py
+++ b/queue_services/events-listener/src/events_listener/worker.py
@@ -32,6 +32,7 @@ from datetime import datetime
 import nats
 from auth_api.models import Org as OrgModel
 from auth_api.models import db
+from auth_api.services import Flags
 from auth_api.services.queue_publisher import publish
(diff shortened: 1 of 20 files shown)


seeker25 (approved):

kialj876 (approved):
looks good 👍

queue_services/events-listener/src/events_listener/worker.py
```diff
@@ -32,6 +32,7 @@
 import nats
 from auth_api.models import Org as OrgModel
 from auth_api.models import db
+from auth_api.services import Flags
```
kialj876, line 35: @seeker25 in the future we might wanna put this Flags service in the shared common BE repo or even the entity_queue_common so that other things can use it like this without coupling with one of the apis

PR #2125, opened 2022-10-25
13962 - manage if an api-gw account is accessing a resource

thorwolpert:
Part of:
*Issue #:* https://github.com/bcgov/entity/issues/13962

*Description of changes:*
- allow api-gw access to resources from the account level

### DIFF
--- a/auth-api/src/auth_api/models/views/authorization.py
+++ b/auth-api/src/auth_api/models/views/authorization.py
@@ -55,6 +55,9 @@ class Authorization(db.Model):
             auth = cls.query.filter_by(keycloak_guid=keycloak_guid,
                                        business_identifier=business_identifier,
                                        org_id=org_id).one_or_none()
+        elif business_identifier and org_id:
+            auth = cls.query.filter_by(business_identifier=business_identifier,
+                                       org_id=org_id).first()
         elif keycloak_guid and business_identifier:
             auth = cls.query.filter_by(keycloak_guid=keycloak_guid,
                                        business_identifier=business_identifier).one_or_none()
--- a/auth-api/src/auth_api/services/authorization.py
+++ b/auth-api/src/auth_api/services/authorization.py
@@ -21,7 +21,7 @@ from flask import abort, current_app
 
 from auth_api.models.views.authorization import Authorization as AuthorizationView
 from auth_api.services.permissions import Permissions as PermissionsService
-from auth_api.utils.enums import ProductTypeCode as ProductTypeCodeEnum
+from auth_api.utils.enums import LoginSource, ProductTypeCode as ProductTypeCodeEnum
 from auth_api.utils.roles import STAFF, Role
 from auth_api.utils.user_context import UserContext, user_context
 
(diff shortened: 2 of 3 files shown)


auth-api/src/auth_api/services/authorization.py
```diff
@@ -109,18 +109,27 @@ def get_user_authorizations_for_entity(business_identifier: str, expanded: bool
...
-                    org_id=user_from_context.account_id
-                )
+            if business_identifier and (org_id := user_from_context.account_id):
 
-            if auth:
-                permissions = PermissionsService.get_permissions_for_membership(auth.status_code, auth.org_membership)
-                auth_response = Authorization(auth).as_dict(expanded)
-                auth_response['roles'] = permissions
+                # if this is an API GW account, check if the account has access to the resource
+                if user_from_context.login_source == LoginSource.API_GW.value:
```
seeker25, line 115: I think this should happen in legal-api, not auth-api.. the function you're changing is `get_user_authorizations_for_entity`, when there is already a function for `get_account_authorizations_for_org` on line 72

auth-api/src/auth_api/services/authorization.py
```diff
@@ -109,18 +109,27 @@ def get_user_authorizations_for_entity(business_identifier: str, expanded: bool
...
+                    auth = AuthorizationView.find_user_authorization_by_business_number(
+                        business_identifier=business_identifier,
+                        org_id=org_id
+                    )
+
+                # Check if the user has access to the resource
+                if keycloak_guid := user_from_context.sub:
```
seeker25, line 122: Is it possible to have no sub? I don't think so

auth-api/src/auth_api/models/views/authorization.py
```diff
@@ -55,6 +55,9 @@ def find_user_authorization_by_business_number(cls, business_identifier: str, ke
             auth = cls.query.filter_by(keycloak_guid=keycloak_guid,
                                        business_identifier=business_identifier,
                                        org_id=org_id).one_or_none()
+        elif business_identifier and org_id:
```
seeker25, line 58: I'd move this last?

auth-api/src/auth_api/services/authorization.py
```diff
@@ -109,18 +109,27 @@ def get_user_authorizations_for_entity(business_identifier: str, expanded: bool
...
                     auth_response['roles'] = permissions
         else:
-            keycloak_guid = user_from_context.sub
-            if business_identifier and keycloak_guid:
-                auth = AuthorizationView.find_user_authorization_by_business_number(
-                    business_identifier=business_identifier,
-                    keycloak_guid=keycloak_guid,
-                    org_id=user_from_context.account_id
-                )
+            if business_identifier and (org_id := user_from_context.account_id):
```
seeker25, line 112: is it possible `org_id` could have been None previously?

auth-api/src/auth_api/services/authorization.py
kialj876: I agree this should be an elif (api_gw users will have the keycloak_guid and then get nothing). I think the org_id is correct though since user_context is populated from the jwt not the header in the decorator right? I don't see anything in there that tries to get it from the header

auth-api/src/auth_api/services/authorization.py
kialj876: yeah and if they are a system user it should go through the other 'if' so this should be good

auth-api/src/auth_api/services/authorization.py
seeker25: Might be fine, but still risky.. I'd look at the `account_id_claim` instead of the `account_id`

image

thorwolpert: I left the regular user to the account_id, and the gw user to use account_id_claim
I don't think regular users have their account in the JWT as they can change it between calls.

seeker25 (approved):

PR #2134, opened 2022-10-31
14025 -Passcode prompt presented when it shouldn't be - Firms do not have passcodes

ketaki-deodhar:
*Issue #:*
https://github.com/bcgov/entity/issues/14025

*Description of changes:*

Added a new modal for removing SP/GP. The wording is different from Passcode modal.


ketaki-deodhar:
Hi all, I cannot add reviewers here yet. Do not have write permissions yet.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -12,13 +12,8 @@
           <span class="subtitle">{{ $t('myBusinessDashSubtitle') }}</span>
         </h1>
         <div class="view-header__actions">
-          <v-btn
-            id="add-name-request-btn"
-            class="font-weight-regular"
-            color="primary"
-            outlined dark large
-            @click="goToNameRequest()"
-          >
```
severinbeauvais, line 21: This is an acceptable code style -- no need to change it.

Also, note that the `>` is in line with the `<` so that elements under this line explicitly look nested.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -28,26 +23,19 @@
...
+                  <v-btn id="add-existing-btn" class="mt-2 mr-4" color="primary" dark large
+                    v-on="{ ...onMenu, ...onTooltip }" @click="addAffiliationDropdown = !addAffiliationDropdown">
                     <v-icon>mdi-plus</v-icon>
                     <span><strong>Add an Existing Business or Name Request</strong></span>
                     <v-icon class="ml-2 mr-n2">{{ addAffiliationDropdown ? 'mdi-menu-up' : 'mdi-menu-down' }}</v-icon>
                   </v-btn>
                 </template>
-                <span>To view and manage existing businesses and Name Requests, you can manually add them to your table.</span>
+                <span>To view and manage existing businesses and Name Requests, you can manually add them to your
+                  table.</span>
```
severinbeauvais, line 38: Spans are inline elements (not block level) so this change injects a carriage return into the string. This may be a problem in unit tests that check for exact matches, or may result in a newline on screen depending how this span is used.

Also, our max line length is 120, although I generally try to keep them under 100 as you have done here.

I'm not saying you need to revert this, but you do need to aware. If this still displays OK then this is fine.

ChrisGabel23 (commented):
Looks like a lot of these formatting changes might have been done by an auto-formatter. Is there a standard used at Registries?

auth-web/src/components/auth/manage-business/EntityManagement.vue
severinbeauvais, line 21: This is an acceptable code style -- no need to change it.

Also, note that the `>` is in line with the `<` so that elements under this line explicitly look nested.

severinbeauvais: Please revert the other changes from multi-line to single-line.

We're kinda used to the original style so best not to change it at this time. We can discuss further if you feel strongly about this.

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -623,7 +572,8 @@ export default class EntityManagement extends Mixins(AccountMixin, AccountChange
     color: $app-blue !important;
   }
 
-  .v-list .v-list-item--active, .v-list .v-list-item--active {
+  .v-list .v-list-item--active,
+  .v-list .v-list-item--active {
```
severinbeauvais, line 576: I like this change -- much easier to see that it's separate selectors with the same styles.

severinbeauvais:
> Looks like a lot of these formatting changes might have been done by an auto-formatter. Is there a standard used at Registries?

I noticed that too. We do use ESLint, which in some cases fixes simple issues (like spaces around =).

We don't have a standard "prettier" program at the moment. Do want to suggest one? Check also with @kialj876 and @cameron-freshworks in case they already use one or have a preference.

cameron-freshworks:
> > Looks like a lot of these formatting changes might have been done by an auto-formatter. Is there a standard used at Registries?
> 
> I noticed that too. We do use ESLint, which in some cases fixes simple issues (like spaces around =).
> 
> We don't have a standard "prettier" program at the moment. Do want to suggest one? Check also with @kialj876 and @cameron-freshworks in case they already use one or have a preference.

We do have a prettier config in Assets UI: https://github.com/bcgov/ppr/blob/main/ppr-ui/prettier.config.js
Although not exactly standardized, just some preferences in current state 👍

ChrisGabel23:
> > > Looks like a lot of these formatting changes might have been done by an auto-formatter. Is there a standard used at Registries?
> > 
> > 
> > I noticed that too. We do use ESLint, which in some cases fixes simple issues (like spaces around =).
> > We don't have a standard "prettier" program at the moment. Do want to suggest one? Check also with @kialj876 and @cameron-freshworks in case they already use one or have a preference.
> 
> We do have a prettier config in Assets UI: https://github.com/bcgov/ppr/blob/main/ppr-ui/prettier.config.js Although not exactly standardized, just some preferences in current state 👍

I used one on TheQ project, but unfortunately I don't remember which one. I have no objections to Cameron's suggestion.

seeker25:
Looks way easier to figure out what changed now.

seeker25:
Looks way easier to figure out what changed now.

seeker25 (approved):

ketaki-deodhar:
> Looks way easier to figure out what changed now.

There was some formatting/linting setting which caused the issue. Sorted it and now looks way cleaner.

severinbeauvais (approved):
LGTM but please update and commit the app version (package files).

ketaki-deodhar:
> Looks way easier to figure out what changed now.

There was some formatting/linting setting which caused the issue. Sorted it and now looks way cleaner.

> 


> LGTM but please update and commit the app version (package files).

Updated.

PR #2163, opened 2022-11-23
14249 - refactored enums to LEAR style enums

MatthewCai2002:
*Issue #: 14249*
https://github.com/bcgov/entity/issues/14249

*Description of changes:*
    - added `base_enum.py` file into `auth_api/utils` as opposed to importing it from LEAR to avoid adding a dependency on LEAR in auth_api
    - refactored `auth_api/utils/enums.py` and ` account_mailer/enums.py` to use LEAR style enums

### DIFF
--- a/auth-web/package-lock.json
+++ b/auth-web/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "auth-web",
-      "version": "1.1.21",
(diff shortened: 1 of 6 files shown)


seeker25:
@MatthewCai2002 There is a database migration component to this as well.

PR #2183, opened 2022-12-12
14172 - Update "How does it work?"

ketaki-deodhar:
*Issue #:*
https://github.com/bcgov/entity/issues/14172

*Description of changes:*
Update "How does it work?" text

### DIFF
--- a/auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
+++ b/auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
@@ -1,24 +1,40 @@
 <template>
   <v-tooltip top max-width="450px" light content-class="tooltip">
     <template v-slot:activator="{ on }">
-      <span v-on="on" class="tooltip-text">Numbered Benefit Company</span>
+      <span v-on="on" class="tooltip-text" v-if="enableBcCccUlc">Numbered Company</span>
+      <span v-on="on" class="tooltip-text" v-else>Numbered Benefit Company</span>
     </template>
     <v-card class="tooltip-content">
-      <h3 class="mb-3">Numbered Benefit Company</h3>
-      <span>A Benefit Company can choose to use as its name the incorporation number of the company followed by “B.C.
-        Ltd.” The incorporation number is assigned by the Business Registry after the Incorporation Application is filed
-        and the company is incorporated.</span>
+      <div v-if="enableBcCccUlc">
+        <h3 class="mb-3">Numbered Company</h3>
+        <span>A Company can choose to use as its name the incorporation number of the company followed by “B.C. Ltd.”,
+          "B.C. Unlimited Liability Company", or "B.C. Community Contribution Company." The incorporation number is
+          assigned by the Business Registry after the Incorporation Application is filed and the company is
+          incorporated.</span>
+      </div>
+      <div v-else>
+        <h3 class="mb-3">Numbered Benefit Company</h3>
+        <span>A Benefit Company can choose to use as its name the incorporation number of the company followed by “B.C.
+          Ltd.” The incorporation number is assigned by the Business Registry after the Incorporation Application is
+          filed and the company is incorporated.</span>
+      </div>
     </v-card>
   </v-tooltip>
 </template>
 
 <script lang="ts">
 import { Component } from 'vue-property-decorator'
+import { LDFlags } from '@/util/constants'
+import LaunchDarklyService from 'sbc-common-components/src/services/launchdarkly.services'
 import Vue from 'vue'
 
 @Component({})
 export default class NumberedCompanyTooltip extends Vue {
+  private get enableBcCccUlc (): boolean {
+    return LaunchDarklyService.getFlag(LDFlags.EnableBcCccUlc) || false
+  }
 }
+
 </script>
 <style lang="scss" scoped>
   @import '$assets/scss/theme.scss';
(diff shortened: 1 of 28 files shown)


kialj876:
Do these changes need to wrapped in a feature flag? I don't think we would want to release this until the different company types are available in the business app.

It would probably be beneficial to have the text come from the feature flag unless all the new types are getting launched at the same time

ketaki-deodhar:
> Do these changes need to be wrapped in a feature flag? I don't think we would want to release this until the different company types are available in the business app.
> 
> You might need different variations of this if the types aren't being released at the same time

Yes. You are right. Missed it. Will fix it.

ketaki-deodhar:
> Do these changes need to be wrapped in a feature flag? I don't think we would want to release this until the different company types are available in the business app.
> 
> You might need different variations of this if the types aren't being released at the same time

Added feature flag and fixed tests.

seeker25 (approved):

severinbeauvais:
Please update app version. Re: https://github.com/bcgov/sbc-auth/blob/main/auth-web/package.json

seeker25:
Also: Are these flags setup in DEV/TEST/PROD?

severinbeauvais:
> Also: Are these flags setup in DEV/TEST/PROD?

Yes, this is an existing flag in LD sbc-auth project.
Dev: true
Test: true
Production: false

severinbeauvais (approved):
LGTM.

Please create a new Tech Debt ticket to remove this flag and conditional code once this feature is deployed to Prod.

auth-web/src/views/auth/home/IncorpOrRegisterView.vue
```diff
@@ -14,41 +14,80 @@
...
               </v-list-item-subtitle>
-              <v-list-item class="list-item" v-for="(item, index) in bulletPoints" :key="index">
-                <v-icon size="8" class="list-item-bullet mt-5">mdi-square</v-icon>
-                <v-list-item-content>
-                  <v-list-item-subtitle class="list-item-text">
-                    {{item.text}}
-                  </v-list-item-subtitle>
-                </v-list-item-content>
-              </v-list-item>
+              <div v-if="enableBcCccUlc">
```
kialj876, line 17: can you please add a computed value for `bulletPoints` that changes based on the `enableBcCccUlc` flag instead of having the v-if/else clause?

auth-web/src/views/auth/home/IncorpOrRegisterView.vue
```diff
@@ -14,41 +14,80 @@
...
-                          <v-list-item-subtitle class="list-item-text py-3">
-                            {{subItem.text}}
-                          </v-list-item-subtitle>
-                        </v-list-item-content>
-                      </v-list-item>
-                    </v-list-item-content>
-                </v-expansion-panel-content>
-              </v-expansion-panel>
-            </v-expansion-panels>
+            <div v-if="enableBcCccUlc">
```
kialj876, line 41: same thing here for `expansionPanels`

auth-web/src/views/auth/home/MaintainBusinessView.vue
```diff
@@ -5,22 +5,43 @@
...
-                <v-icon size="8" class="list-item-bullet mt-5">mdi-square</v-icon>
-                <v-list-item-content>
-                  <v-list-item-subtitle class="list-item-text">
-                    {{item.text}}
-                  </v-list-item-subtitle>
-                </v-list-item-content>
-              </v-list-item>
-            </v-list-item-content>
-          </v-list-item>
+          <div v-if="enableBcCccUlc">
```
kialj876, line 8: same idea here

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
```diff
@@ -1,24 +1,40 @@
 <template>
   <v-tooltip top max-width="450px" light content-class="tooltip">
     <template v-slot:activator="{ on }">
-      <span v-on="on" class="tooltip-text">Numbered Benefit Company</span>
+      <span v-on="on" class="tooltip-text" v-if="enableBcCccUlc">Numbered Company</span>
```
kialj876, line 4: pls rmv v-if/else and do a computed value for the text here

auth-web/src/components/auth/common/NumberedCompanyTooltip.vue
```diff
@@ -1,24 +1,40 @@
...
+      <span v-on="on" class="tooltip-text" v-else>Numbered Benefit Company</span>
     </template>
     <v-card class="tooltip-content">
-      <h3 class="mb-3">Numbered Benefit Company</h3>
-      <span>A Benefit Company can choose to use as its name the incorporation number of the company followed by “B.C.
-        Ltd.” The incorporation number is assigned by the Business Registry after the Incorporation Application is filed
-        and the company is incorporated.</span>
+      <div v-if="enableBcCccUlc">
```
kialj876, line 8: same thing in here

kialj876 (changes requested):
Hey Ketaki,

You're PR introduces a lot of duplicate code that can be avoided by using computed values for your text instead of v-if/else blocks, can you please update it based on my comments?
It will make it easier in the future to read / refactor / do styling updates etc.

Other than that I think it is looking good 👍

PR #2189, opened 2022-12-16
14066 Auth Web: Affiliation certifier should be editable by SBC Staff and Registries Staff

lewischenstudio:
*Issue #:*
https://github.com/bcgov/entity/issues/14066

*Description of changes:*
- updated UI to allow text entry of certifier (when staff only)
- updated UI to save certifier in call to API
- updated unit tests

### DIFF
--- a/auth-web/package-lock.json
+++ b/auth-web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "auth-web",
-  "version": "1.1.30",
+  "version": "1.1.29",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "auth-web",
-      "version": "1.1.30",
+      "version": "1.1.29",
       "dependencies": {
         "@bcrs-shared-components/bread-crumb": "^1.0.2",
         "@bcrs-shared-components/corp-type-module": "1.0.9",
--- a/auth-web/package-lock.json
+++ b/auth-web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "auth-web",
-  "version": "1.1.28",
+  "version": "1.1.30",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "auth-web",
-      "version": "1.1.28",
+      "version": "1.1.30",
       "dependencies": {
         "@bcrs-shared-components/bread-crumb": "^1.0.2",
         "@bcrs-shared-components/corp-type-module": "1.0.9",
@@ -47,7 +47,7 @@
         "@types/vuelidate": "^0.7.13",
         "@typescript-eslint/eslint-plugin": "^5.23.0",
         "@typescript-eslint/parser": "^5.23.0",
-        "@volar-plugins/vetur": "*",
+        "@volar-plugins/vetur": "latest",
         "@vue/cli-plugin-babel": "^4.5.17",
         "@vue/cli-plugin-e2e-nightwatch": "^4.5.17",
         "@vue/cli-plugin-eslint": "^4.5.17",
--- a/auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
+++ b/auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
@@ -72,7 +72,7 @@
             <!-- Certified By Name -->
             <v-expand-transition>
               <v-text-field
-                v-if="isBusinessIdentifierValid && isFirm && isStaffAccount"
+                v-if="isBusinessIdentifierValid && showCertifier"
                 filled
                 persistent-hint
                 :label="certifiedByNameLabel"
@@ -156,13 +156,8 @@ import { FolioNumberload, LoginPayload } from '@/models/business'
 import Certify from './Certify.vue'
 import CommonUtils from '@/util/common-util'
 import HelpDialog from '@/components/auth/common/HelpDialog.vue'
-import { KCUserProfile } from 'sbc-common-components/src/models/KCUserProfile'
-import { Role } from '@/util/constants'
 import { StatusCodes } from 'http-status-codes'
 import { mapActions } from 'vuex'
-import { namespace } from 'vuex-class'
-
-const UserModule = namespace('user')
 
 @Component({
   components: {
@@ -183,9 +178,13 @@ export default class AddBusinessDialog extends Vue {
     helpDialog: HelpDialog
   }
 
-  @UserModule.State('currentUser') readonly currentUser!: KCUserProfile
-
   @Prop({ default: false }) readonly dialog: boolean
+
+  @Prop({ default: false }) readonly isGovStaffAccount: boolean
+
+  @Prop({ default: '' }) readonly userFirstName: string
+
+  @Prop({ default: '' }) readonly userLastName: string
 
   private readonly addBusiness!: (loginPayload: LoginPayload) => any
   private readonly updateBusinessName!: (businessNumber: string) => any
@@ -199,6 +198,7 @@ export default class AddBusinessDialog extends Vue {
   protected isCertified = false // firms only
   protected businessIdentifierRules = []
   protected certifiedByName = ''
+  protected validCertifier = true
 
   readonly certifyClause = 'Note: It is an offence to make or assist in making a false or misleading ' +
     'statement in a record filed under the Partnership Act. A person who commits this offence is ' +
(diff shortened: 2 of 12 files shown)


lewischenstudio:
Staff User View:
image

lewischenstudio:
Client User View:
BCREG0020

image

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
```diff
@@ -191,6 +218,22 @@ export default class AddBusinessDialog extends Vue {
     return CommonUtils.isFirmNumber(this.businessIdentifier)
   }
 
+  get isStaffAccount (): boolean {
+    return this.currentUser.roles.includes(Role.Staff) || false
```
seeker25, line 222: This only works for BC Reg staff? You'll have to check to see their ORG if they're in the SBC_STAFF.

seeker25 (changes requested):

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
lewischenstudio: Here I've changed the code to include `GOVMAccountUser` (sbc staff) for this part.

auth-web/package-lock.json
```diff
@@ -47,7 +47,7 @@
         "@types/vuelidate": "^0.7.13",
         "@typescript-eslint/eslint-plugin": "^5.23.0",
         "@typescript-eslint/parser": "^5.23.0",
-        "@volar-plugins/vetur": "*",
+        "@volar-plugins/vetur": "latest",
```
severinbeauvais, line 50: If you rebase then you should not see this change.

Also, why the jump from 1.1.28 to 1.1.30?

auth-web/src/models/business.ts
```diff
@@ -6,6 +6,7 @@ export interface LoginPayload {
     passCode?: string
     phone?: string
     email?: string
+    certifiedByName?: string
```
severinbeauvais, line 9: In which case is this not present?

lewischenstudio: I'm not sure in which case this is not present, but having "?" will not break anything if this is not present in any case.

auth-web/package-lock.json
lewischenstudio: I changed 1.1.30 to 1.1.29. This "latest" shows up when I do "npm i".

auth-web/src/models/business.ts
severinbeauvais: It won't break anything but it might be incorrect. This is an anti-pattern.

I suggest you remove the ? and see if anything breaks.

auth-web/package-lock.json
severinbeauvais: OK. Looks like that package was changed in package.json but not package lock (until now).

severinbeauvais:
Looks like there's something wrong with the setup for the broken unit test in AddBusinessDialog.

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
```diff
@@ -191,6 +218,22 @@ export default class AddBusinessDialog extends Vue {
     return CommonUtils.isFirmNumber(this.businessIdentifier)
   }
 
+  get isStaffAccount (): boolean {
+    return this.currentUser.roles.some((role) => (role === Role.Staff || role === Role.GOVMAccountUser))
```
seeker25, line 222: Role GOVM account User? I don't think that's right

seeker25 (changes requested):
SBC STAFF identification is still incorrect

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
seeker25: You'll have to look at the org's `type_code`

severinbeauvais: Travis, do you have sample code?

seeker25:
```
  protected get isSbcStaffAccount (): boolean {
    return this.currentOrganization?.orgType === Account.SBC_STAFF
  }
  ```
  if we're dealing with Affiliations to SBC_STAFF org.. if we're not that will need to be changed

File: auth-web/src/models/business.ts (hunk 3 above)
Comment (lewischenstudio) [reply auth-web/src/models/business.ts]: It's now removed, and nothing breaks.

File: auth-web/src/components/auth/manage-business/AddBusinessDialog.vue (hunk 4 above)
Comment (lewischenstudio) [reply auth-web/src/components/auth/manage-business/AddBusinessDialog.vue]: I pass in isSbcStaffAccount from EntityManagement to AddBusinessDialog. This prevents the need to fetch staff account in AddBusinessDialog and passes the unit test files.

File: auth-web/src/components/auth/manage-business/AddBusinessDialog.vue (hunk 1 above)
Comment (lewischenstudio) [reply auth-web/src/components/auth/manage-business/AddBusinessDialog.vue]: This code is now removed. isStaffAccount is passed from EntityManagement to AddBusinessDialog instead of fetching it here.

File: auth-web/src/components/auth/manage-business/AddBusinessDialog.vue [hunk 5]
```diff
@@ -69,11 +69,28 @@
...
 
+            <!-- Certified By Name -->
+            <v-expand-transition>
+              <v-text-field
+                v-if="isBusinessIdentifierValid && showCertifier"
+                filled
+                persistent-hint
+                :label="certifiedByNameLabel"
+                :rules="certifiedByNameRules"
+                :maxlength="passcodeMaxLength"
```
severinbeauvais, line 80: This doesn't look like the right max length.

What is the max length for the Certified By Name? Or what is a reasonable max number?

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
```diff
@@ -191,6 +218,22 @@ export default class AddBusinessDialog extends Vue {
...
+
+  get currentUserName (): string {
+    if (this.isGovStaffAccount) return this.certifiedByName
+    return `${this.userLastName}, ${this.userFirstName}`
+  }
+
+  get certifiedByNameRules (): any[] {
+    return [
+      v => !!v || 'Certification is required',
+      v => v.length <= 150 || 'Maximum 150 characters'
```
severinbeauvais, line 233: I see this is code copied from the passcode. However, I think you don't need this line since maxLength is already being used.

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
lewischenstudio: I am using the same max length for passcode if it is for firm. 150 is a pretty good max number.

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
lewischenstudio: The text is different in passcodeRules, so I have to use another rule.

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
severinbeauvais: You're overloading the use of passcodeMaxLength. This is an anti-pattern!

Did you confirm that 150 is a good number with an auth API developer (who can check the db field size) and the UI designers (who can tell you what makes sense from a business point of view)?

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
severinbeauvais: I was not saying to reuse the passcode rules! I was saying that you don't need line 233.

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
lewischenstudio: Sure. I can ask the UI designers.

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
lewischenstudio: Sure. I can remove line 233.

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
lewischenstudio: I have just verified with Ethan that 150 is the max length for the certifier name.

severinbeauvais (approved):

ChrisGabel23 (approved):

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
seeker25: ` return this.currentUser.roles.some((role) => (role === Role.Staff || role === Role.GOVMAccountUser))`

->
` return this.currentUser.roles.some((role) => (role === Role.Staff))`

seeker25 (approved):

seeker25:
@kialj876 Can you take a look?

kialj876 (approved):

seeker25:
Note this may require some subsequent UI tweaks.

severinbeauvais:
> Note this may require some subsequent UI tweaks.

How will we know if it does?

PR #2190, opened 2022-12-19
Revert "14172 - Update "How does it work?""

seeker25:
Reverts bcgov/sbc-auth#2183

So we can push the build out. Ketaki is on vacation.


kialj876 (approved):

severinbeauvais:
Did PR 2183 break something?

seeker25:
@severinbeauvais no Ketaki is on vacation, and we have some changes that need to be pushed out now. Rather just revert the ticket and get it in place later.. then UX can look at it QA etc.

severinbeauvais (approved):

PR #2193, opened 2022-12-20
14507 add column of BCOL USER ID and 'FIRST' to Admin Email&Phone

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/14507

*Description of changes:*
added column of BCOL USER ID and word 'FIRST' to Admin Email&Phone column

### DIFF
--- a/docs/docs/database/bca_users_accounts.sql
+++ b/docs/docs/database/bca_users_accounts.sql
@@ -7,6 +7,7 @@ select
 	o.name as "ACCOUNT NAME",
 	o.branch_name as "BRANCH NAME",
 	m.membership_type_code as "USER TYPE",
+	o.bcol_user_id as "BCOL USER ID",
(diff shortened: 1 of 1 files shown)


docs/docs/database/bca_users_accounts.sql
```diff
@@ -31,6 +32,6 @@ LEFT JOIN contacts c2 on c2.id = cl2.contact_id
 where 
 ps.product_code = 'BCA' 
 and membership_type_code = 'ADMIN'
-and o.id in ('') -- GRAB FROM PAY query.
```
seeker25, line 34: You'll probably need to get a list of users from PAY and use them here. ORGS I don't think make sense, because they can be fairly large.

seeker25 (approved):

seeker25:
Note there is a Juypter notebook for this now. @Jxio maybe you could add a readme or a note linking to it?

PR #2194, opened 2022-12-23
14796 - Fix @vue/composition-api import so it takes from node_modules, instea…

seeker25:
https://github.com/bcgov/entity/issues/14796

### DIFF
--- a/auth-web/package-lock.json
+++ b/auth-web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "auth-web",
-  "version": "1.1.30",
+  "version": "1.1.31",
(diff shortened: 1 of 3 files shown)


Jxio (approved):

thorwolpert (approved):
why the explicit pathing?
I see it is working around a conflict. Is this something to share across teams?

seeker25:
Probably, `@vue/composition-api` and `vue` need to be external when being used as a plugin.

Otherwise having two `@vue/composition-api` instances gets messy and cause issues. Auth-web recently added `@vue/composition-api`.

PR #2195, opened 2022-12-23
14796 - Fix reference to fas-ui.

seeker25:
https://github.com/bcgov/entity/issues/14796


thorwolpert (approved):
i like the ref change!

PR #2198, opened 2023-01-04
Update Vuetify, so v-select works correctly like fas-ui

seeker25:


seeker25:
This seems to have broken alot of CSS and other issues. Closing for now. Probably easier to wait until March until Vuetify 3 with data tables comes out and upgrade then.

PR #2202, opened 2023-01-06
14249 - enum refactor

MatthewCai2002:
*Issue #:14249*
https://github.com/bcgov/entity/issues/14249

*Description of changes:*
    - refactored enums in auth-api and account-mailer to LEAR style enums based on https://github.com/thorwolpert/flask-ld-logger/blob/main/flask_ld_logger/enums.py

### DIFF
--- a/auth-api/src/auth_api/utils/base_enum.py
+++ b/auth-api/src/auth_api/utils/base_enum.py
@@ -1,0 +1,41 @@
+# Copyright © 2021 Province of British Columbia
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


auth-api/src/auth_api/utils/base_enum.py
```diff
@@ -0,0 +1,41 @@
...
+        """Return True if 'in' the Enum."""
+        try:
+            self(other)  # pylint: disable=no-value-for-parameter
+        except ValueError:
+            return False
+        else:
+            return True
+
+
+class BaseEnum(Enum, metaclass=BaseMeta):
```
MatthewCai2002, line 32: removed `str` here because it was converting everything to strings and breaking a bunch of tests

seeker25:
I don't see any migrations? were they not needed?

MatthewCai2002:
These enums don't remove the need for .value to access enum member values. 

Can define `__str__` but then all enum.MEMBER calls are strings -> type errors or can override `__getattribute__` but then enum.MEMBER loses access to enum member objects like .name and .value so some calls to .name break

is there another way around this?

PR #2205, opened 2023-01-09
12753 - Navigating by URL on staff account causes the wrong org to be loaded 

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/12753

*Description of changes:*
Read from route instead of session storage.

### DIFF
--- a/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
+++ b/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
@@ -397,7 +397,7 @@ export default class AccountInfo extends Mixins(
 
   private async mounted () {
     const accountSettings = this.getAccountFromSession()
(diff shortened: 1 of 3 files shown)


auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
```diff
@@ -396,8 +396,7 @@ export default class AccountInfo extends Mixins(
   ]
 
   private async mounted () {
-    const accountSettings = this.getAccountFromSession()
-    await this.syncOrganization(accountSettings.id)
+    await this.syncOrganization(parseInt(this.$route.params['orgId']))
```
kialj876, line 399: Are you sure orgId is always passed as a param?

seeker25: Yes

PR #2206, opened 2023-01-09
12753 - Navigating by URL on staff account causes the wrong org to be loaded - Part2

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/12753

*Description of changes:*
Needed to change router, so it rerenders on path change

### DIFF
--- a/auth-web/src/App.vue
+++ b/auth-web/src/App.vue
@@ -38,7 +38,7 @@
       <pay-system-alert />
     </div>
     <div class="app-body">
-      <router-view  />
+      <router-view :key="$route.path" />
     </div>
     <sbc-footer></sbc-footer>
   </v-app>
--- a/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
+++ b/auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
@@ -397,7 +397,7 @@ export default class AccountInfo extends Mixins(
 
   private async mounted () {
     const accountSettings = this.getAccountFromSession()
-    await this.syncOrganization(accountSettings.id)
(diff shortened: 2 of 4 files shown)


auth-web/src/App.vue
```diff
@@ -38,7 +38,7 @@
       <pay-system-alert />
     </div>
     <div class="app-body">
-      <router-view  />
+      <router-view :key="$route.path" />
```
kialj876, line 41: what is this for?

kialj876 (approved):

auth-web/src/App.vue
seeker25: Basically if we don't have this and you navigate to the same URL it wont re-render the components. 

EX. navigate to:
https://test.bcregistry.ca/business/auth/account/1999/settings/account-info

but then you go into 

image

It wont rerender the component at all, even though the route has changed.

auth-web/src/components/auth/account-settings/account-info/AccountInfo.vue
```diff
@@ -396,7 +396,7 @@ export default class AccountInfo extends Mixins(
   ]
 
   private async mounted () {
-    await this.syncOrganization(parseInt(this.$route.params['orgId']))
+    await this.syncOrganization(parseInt(this.$route.params.orgId))
```
seeker25, line 399: better than dict access I think

auth-web/src/App.vue
seeker25: If I navigated repeatedly using the URL, it would work correctly.. but the button press above.. wouldn't rerender the component.

PR #2209, opened 2023-01-12
refactored padInfoForm comp

kialj876:
Signed-off-by: Kial Jinnah <kialjinnah@gmail.com>

*Issue #:*
https://github.com/bcgov/entity/issues/12322

*Description of changes:*


auth-web/src/components/auth/common/PADInfoForm.vue
```diff
@@ -126,169 +123,155 @@
...
-  private transitNumber: string = ''
-  private institutionNumber: string = ''
-  private accountNumber: string = ''
-  private isTOSAccepted: boolean = false
-  private isAcknowledged: boolean = false
-  private isTouched: boolean = false
-  private isStartedEditing: boolean = false
-  private bankInfoDialog: boolean = false
+// FUTURE: remove this in vue 3
+interface PADInfoFormState {
```
seeker25, line 134: Should we put this in a separate interface folder?

auth-web/src/components/auth/common/PADInfoForm.vue
```diff
@@ -126,169 +123,155 @@
...
+        return (state.showPremiumPADInfo)
+          ? 'I understand that services will continue to be billed to the linked BC Online account until the mandatory (3) day confirmation period has ended.'
+          : 'I understand that this account will not be able to perform any transactions until the mandatory (3) day confirmation period for pre-authorized debit has ended.'
+      }),
+      padInfoSubtitle: computed((): string => {
+        return (state.showPremiumPADInfo)
+          ? 'Services will continue to be billed to the linked BC Online account until the mandatory (3) day confirmation period has ended.'
+          : 'This account will not be able to perform any transactions until the mandatory (3) day confirmation period has ended.'
+      })
+    }) as unknown) as PADInfoFormState
```
seeker25, line 205: There might be a better way of doing this? Not sure why you have to cast as unknown

auth-web/tests/unit/components/PADInfoForm.spec.ts
```diff
@@ -62,20 +75,73 @@ describe('PADInfoForm.vue', () => {
...
 
+  afterEach(() => {
+    wrapper.destroy()
+    sandbox.restore()
+  })
+
   it('is a Vue instance', () => {
     expect(wrapper.isVueInstance()).toBeTruthy()
   })
+  it('renders with base values', () => {
```
seeker25, line 99: Thank you for this Kial

seeker25 (approved):
LGTM

auth-web/src/components/auth/common/PADInfoForm.vue
kialj876: I tried a lot of things here. Nothing worked. It works fine without this in ppr / any other composition api I've written but in auth-web the build fails without this

auth-web/src/components/auth/common/PADInfoForm.vue
kialj876: I was thinking we would keep them inside since it'll be deleted asap rather than clutter the interfaces folder. It should be inferring the types without the interface (every other project I've tried you don't need this). It's a symptom to do with your comment below

auth-web/src/components/auth/common/PADInfoForm.vue
kialj876: I think I'll have another go at this later, but I need to let it simmer / sleep on it a bit before trying again

seeker25: Sounds good, in the mean time.. lets merge it

PR #2216, opened 2023-01-18
14886 - converted deactivateCard.vue to composition-api

MatthewCai2002:
*Issue #: 14886*
https://github.com/bcgov/entity/issues/14886

*Description of changes:*
    - Converted deactivateCard.vue into composition-api
    - Updated VueI18n plugin to work with composition-api
    - Updated deactivateCard tests with VueI18n plugin

### DIFF
--- a/auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
+++ b/auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
@@ -13,7 +13,7 @@
         </div>
         <div class="ml-3 mt-1">
           <h4 class="font-weight-bold">{{ t(item.title) }}</h4>
-          <p>{{ t(item.description) }}</p>
+          <p v-if="item.description">{{ t(item.description) }}</p>
         </div>
       </div>
     </v-card-text>
@@ -21,7 +21,7 @@
 </template>
 
 <script lang="ts">
-import { computed, defineComponent } from '@vue/composition-api'
+import { PropType, computed, defineComponent } from '@vue/composition-api'
 import { Account } from '@/util/constants'
 import Vue from 'vue'
 import { useI18n } from 'vue-i18n-bridge'
@@ -29,7 +29,7 @@ import { useI18n } from 'vue-i18n-bridge'
 export default defineComponent({
   name: 'DeactivateCard',
   props: {
-    type: () => Account
+    type: String as PropType<Account>
   },
   setup (props) {
     const { t } = useI18n()
@@ -37,7 +37,7 @@ export default defineComponent({
       { title: 'deactivateMemberRemovalTitle', description: 'deactivateMemberRemovalDesc' },
       { title: 'businessRemovalTitle', description: 'businessRemovalDesc' },
       { title: 'padRemovalTitle', type: Account.PREMIUM }
-    ]
+    ] as { title: string, description?: string, type?: Account }[]
 
     const info = computed(() => {
       return infoArray.filter(obj => !obj.type || obj.type === props.type)
--- a/auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
+++ b/auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
@@ -21,35 +21,34 @@
 </template>
 
 <script lang="ts">
-import { Component, Prop } from 'vue-property-decorator'
+import { computed, defineComponent, ref } from '@vue/composition-api'
 import Vue from 'vue'
+import { useI18n } from 'vue-i18n-bridge'
 
-@Component({
-})
-export default class DeactivateCard extends Vue {
-  @Prop() private type: string
+export default defineComponent({
+  name: 'DeactivateCard',
+  props: {
+    type: String
+  },
+  setup (props) {
+    const { t } = useI18n()
+    const infoArray = ref([
+      { title: 'deactivateMemberRemovalTitle', description: 'deactivateMemberRemovalDesc' },
+      { title: 'businessRemovalTitle', description: 'businessRemovalDesc' },
+      { title: 'padRemovalTitle', type: 'PREMIUM' }
+    ] as any)
 
-  // no type matches means show it for all types
-  private infoArray: { title: string, description?: string, type?: string }[] = [
-    {
-      title: 'deactivateMemberRemovalTitle',
-      description: 'deactivateMemberRemovalDesc'
-    },
-    {
-      title: 'businessRemovalTitle',
-      description: 'businessRemovalDesc'
-    },
-    {
-      title: 'padRemovalTitle',
-      type: 'PREMIUM'
+    const info = computed(() => {
+      return infoArray.value.filter(obj => !obj.type || obj.type === props.type)
+    })
+
+    return {
+      infoArray,
+      t,
+      info
     }
-
-  ]
-
-  private get info () {
-    return this.infoArray.filter(obj => !obj.type || obj.type === this.type)
   }
-}
+})
 </script>
 
 <style lang="scss" scoped>
--- a/auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
+++ b/auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
@@ -12,8 +12,8 @@
           >
         </div>
         <div class="ml-3 mt-1">
-          <h4 class="font-weight-bold">{{ $t(item.title) }}</h4>
-          <p>{{ $t(item.description) }}</p>
+          <h4 class="font-weight-bold">{{ t(item.title) }}</h4>
+          <p>{{ t(item.description) }}</p>
         </div>
       </div>
     </v-card-text>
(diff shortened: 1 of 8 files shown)


seeker25:
Unit tests need to pass

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
```diff
@@ -21,35 +21,34 @@
...
-      description: 'businessRemovalDesc'
-    },
-    {
-      title: 'padRemovalTitle',
-      type: 'PREMIUM'
-    }
+export default defineComponent({
+  name: 'DeactivateCard',
+  props: {
+    type: String
```
kialj876, line 31: Can you change this to be the type of the enum? Looks like its probably meant to be account type?

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
```diff
@@ -21,35 +21,34 @@
...
+  },
+  setup (props) {
+    const { t } = useI18n()
+    const infoArray = ref([
```
kialj876, line 35: I don't think you need a ref for this since this array is static right?

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
```diff
@@ -21,35 +21,34 @@
...
+      { title: 'deactivateMemberRemovalTitle', description: 'deactivateMemberRemovalDesc' },
+      { title: 'businessRemovalTitle', description: 'businessRemovalDesc' },
+      { title: 'padRemovalTitle', type: 'PREMIUM' }
+    ] as any)
```
kialj876, line 39: generally we only use 'any' as a last resort. I think you can keep the typing of this (replacing type: string with type: AccountType).

NOTE: you can import Ref from @vue/composition-api to type refs. In vue 3 its all inferred which is way nicer but in vue 2 you need to explicitly cast it a lot unfortunately

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
```diff
@@ -21,35 +21,34 @@
...
-  ]
+    const info = computed(() => {
+      return infoArray.value.filter(obj => !obj.type || obj.type === props.type)
+    })
 
-  private get info () {
-    return this.infoArray.filter(obj => !obj.type || obj.type === this.type)
+    return {
+      infoArray,
+      t,
```
kialj876, line 47: you're returning `t` here, but you're using `$t` in the html. You'll need to change that if you want anything to happen. Did you test this locally? The way you have it currently looks like it would break the description

auth-web/tests/unit/components/DeactivateCard.spec.ts
```diff
@@ -4,6 +4,7 @@ import Vue from 'vue'
 import VueRouter from 'vue-router'
 import Vuetify from 'vuetify'
 import Vuex from 'vuex'
+import i18n from '@/plugins/i18n'
```
kialj876, line 8: you need to update all these tests to mock the correct function or even better would be to reinitialize i18n to go from a test file instead of mocking it out. Right now they only pass because you're mocking the `$t` which is actually incorrect since that function is never initialized in the new code (its `t` now)

kialj876 (changes requested):
Nice work! I've got some comments for ya

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
MatthewCai2002: done

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
MatthewCai2002: made the change

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
MatthewCai2002: made the change

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
MatthewCai2002: updated the change in the template

auth-web/tests/unit/components/DeactivateCard.spec.ts
MatthewCai2002: I made a mockI18n.ts file that reinitialized vuei18n but still had to install it onto a localVue instance is that the right way of doing this?

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
```diff
@@ -12,44 +12,44 @@
...
-      description: 'businessRemovalDesc'
-    },
-    {
-      title: 'padRemovalTitle',
-      type: 'PREMIUM'
-    }
+export default defineComponent({
+  name: 'DeactivateCard',
+  props: {
+    type: () => Account
```
kialj876, line 32: bah sorry I know I suggested this but this is actually how you want to type this:
`type: String as PropType<Account>`

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
```diff
@@ -12,44 +12,44 @@
           >
         </div>
         <div class="ml-3 mt-1">
-          <h4 class="font-weight-bold">{{ $t(item.title) }}</h4>
-          <p>{{ $t(item.description) }}</p>
+          <h4 class="font-weight-bold">{{ t(item.title) }}</h4>
+          <p>{{ t(item.description) }}</p>
```
kialj876, line 16: probably want add v-if="item.description" since padRemovalTitle doesn't have an item.description

auth-web/src/components/auth/account-deactivate/DeactivateCard.vue
```diff
@@ -12,44 +12,44 @@
...
+  props: {
+    type: () => Account
+  },
+  setup (props) {
+    const { t } = useI18n()
+    const infoArray = [
+      { title: 'deactivateMemberRemovalTitle', description: 'deactivateMemberRemovalDesc' },
+      { title: 'businessRemovalTitle', description: 'businessRemovalDesc' },
+      { title: 'padRemovalTitle', type: Account.PREMIUM }
+    ]
```
kialj876, line 40: you'll need to add `as { title: string, description?: string, type?: Account }[]` to maintain type inference

auth-web/tests/unit/components/DeactivateCard.spec.ts
```diff
@@ -21,55 +24,73 @@ describe('Deactivated card.vue', () => {
...
+  localVue.use(i18n)
 
   it('Truthy and basic test', () => {
     wrapper = mount(DeactivateCard, {
       vuetify,
       localVue,
       router,
       mocks: {
-        $t: (mock) => mock
+        t: (mock) => mock
```
kialj876, line 35: for this and all the other tests: you've added a mock for i18n already so you don't need to mock out t. It's a better test if t is not mocked otherwise you can commit bugs like before with the test passing

auth-web/tests/unit/test-utils/test-data/MockI18n.ts
```diff
@@ -0,0 +1,22 @@
+import { castToVueI18n, createI18n } from 'vue-i18n-bridge'
+import Vue from 'vue'
+import VueI18n from 'vue-i18n'
+
+export default {
+  mock () {
```
kialj876, line 6: add a param here to add in a custom en object

auth-web/tests/unit/test-utils/test-data/MockI18n.ts
```diff
@@ -0,0 +1,22 @@
...
+    Vue.use(VueI18n, { bridge: true })
+
+    const i18n = castToVueI18n(createI18n({
+      legacy: false,
+      locale: 'en',
+      fallbackLocale: 'en',
+      messages: {
+        en: {
```
kialj876, line 14: set this to the given param (this will allow other test suites to use this with custom data)

auth-web/tests/unit/components/DeactivateCard.spec.ts
```diff
@@ -9,6 +11,7 @@ Vue.use(VueRouter)
 Vue.use(Vuetify)
 const vuetify = new Vuetify({})
 const router = new VueRouter()
+const i18n = MockI18n.mock()
```
kialj876, line 14: if you add in the 'en' param to the MockI18n.mock() you can set the custom en values for the test suite here
i.e.
```
const en = {
  businessRemovalDesc: 'i8n businessRemovalDesc',
  businessRemovalTitle: 'i8n businessRemovalTitle',
  deactivateMemberRemovalDesc: 'i8n deactivateMemberRemovalDesc',
  deactivateMemberRemovalTitle: 'i8n deactivateMemberRemovalTitle',
  padRemovalTitle: 'i8n padRemovalTitle'
}
const i18n = MockI18n.mock(en)
```

kialj876 (changes requested):
Looks really good, just have a couple more small things

kialj876 (approved):
nice work!

PR #2226, opened 2023-01-26
15136 - Rename business-events-listener to namex-event-listener

seeker25:
https://github.com/bcgov/entity/issues/15136

*Description of changes:*
The old `business-events-listener` was listening from namex. It's been renamed to `names-events-listener`. 


seeker25:
Test are failing.. because there are no tests.

seeker25:
Ideally if we re-do the queues.. we can combine the projects and watch multiple queues at the same time hopefully. It will save a good 2000 lines of mostly duplicated code and maintenance.

kialj876 (approved):
I just did a pretty quick scan -- looks good to me. Can you add the devops folder into business-events-listener? I don't think Jia or Matthew will know to put that in

seeker25:
I've removed the business-events-listener.  We wont be using it if we're calling directly into LEAR / NAMEX.

PR #2229, opened 2023-01-30
15134 freeze unfreeze alert

MatthewCai2002:
*Issue #:*

*Description of changes:*
    - added an alert icon and tooltip to tell a user whether or not a business has any important details they need to know about ie: frozen, bad standing, dissolution, and liquidation

*Caveats*
    - right now all of these details are put in a mock array just to see the look and feel of the alert
    - right now alerts are shown just for RITVIK SEP26, want this to only show on all frozen orgs.


auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
```diff
@@ -0,0 +1,112 @@
+<!-- TODO:
```
kialj876, line 1: Can you use the IconTooltip component in here? You will need to add some more props to it, but that way we only have the one template for icon tooltips. You could add a slot to the content as well to make it easy to add your list in there

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
```diff
@@ -0,0 +1,112 @@
...
+
+<script lang="ts">
+import { computed, defineComponent, nextTick, ref } from '@vue/composition-api'
+import { EntityDetailTypes } from '@/util/constants'
+import { PropType } from 'vue'
+
+export default defineComponent({
+  name: 'EntityDetailsAlert',
+  props: {
+    statuses: Array as PropType<Array<String>>
```
kialj876, line 38: this would be an array of EntityDetailTypes right?

seeker25:
Please update package.json as well

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
MatthewCai2002: Yeah I'm assuming that we're going to pass an array of details (are they strings?) to the component. 
forgot to update the typing there

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
```diff
@@ -0,0 +1,112 @@
...
+      if (m1.priority < m2.priority) {
+        return -1
+      }
+      if (m1.priority > m2.priority) {
+        return 1
+      }
+      return 0
+    }
+
+    const detailMessages = computed(() => {
```
kialj876, line 71: I think you only need to run this once right? It will never change dynamically so I don't think you need to make this a computed value

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
```diff
@@ -0,0 +1,112 @@
...
+      let temp = []
+      for (let detail of mockArray) {
+        temp.push(generateMessage(detail))
+      }
+      temp.sort(compareMessages)
+      return temp
+    })
+
+    const alertMessage = computed(() => {
```
kialj876, line 80: same thing here

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
```diff
@@ -0,0 +1,112 @@
...
+      return detailMessages.value.length > 1 ? 'Alerts' : 'Alert'
+    })
+
+    const iconColour = computed(() => {
```
kialj876, line 84: and here

auth-web/src/util/constants.ts
```diff
@@ -435,3 +435,10 @@ export enum RequestTrackerType {
     ChangeStatus = 'CHANGE_STATUS',
     ChangeParty = 'CHANGE_PARTY',
 }
+
+export enum EntityDetailTypes {
```
kialj876, line 439: I would rename this EntityAlertType or something like that so its more clear what this is for

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
MatthewCai2002: I think I would also need to change some of the CSS too since the common tooltip comes out of the bottom and points upward while we want this tooltip to come out of the top and point downward. 

can I do that through props and stuff?

MatthewCai2002: image

vs

image

kialj876: yep just add the css you need for the top variant and set the class/html variant via a prop. It will be really good to have that option for future things

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
kialj876: Or are you just passing an array of enums (EntityDetailsAlert - enum) and this component determines the string based on that?

MatthewCai2002: right now I'm thinking that it's just oging to be an array of strings where each string is a status

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
MatthewCai2002: yeah made 2 classes to toggle between top and bottom tooltips

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
MatthewCai2002: Yeah I don't think any user can mutate that, was just worried if the status of a business changed while a user was on the page. 

But the status would only change when a new call to LEAR or NAMEX is made on page reload so this only ran once

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
MatthewCai2002: made the change

auth-web/src/components/auth/manage-business/EntityDetailsAlert.vue
MatthewCai2002: made the change

auth-web/src/util/constants.ts
MatthewCai2002: Yup renamed

kialj876 (commented):
Great work! You just need to rebase. I think your pckg lock changes were already made in a previous merge

kialj876 (approved):

PR #2234, opened 2023-02-02
Hide hidden products in account-info product page.

seeker25:

### DIFF
--- a/auth-web/src/components/auth/common/Product.vue
+++ b/auth-web/src/components/auth/common/Product.vue
@@ -8,6 +8,7 @@
         class="product-card py-8 px-5 mb-4 elevation-1"
         :class="[ {'processing-card' : isSelected}]"
         :data-test="`div-product-${productDetails.code}`"
+        v-if="!productDetails.hidden"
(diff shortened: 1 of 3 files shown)


auth-web/src/components/auth/common/Product.vue
```diff
@@ -8,6 +8,7 @@
         class="product-card py-8 px-5 mb-4 elevation-1"
         :class="[ {'processing-card' : isSelected}]"
         :data-test="`div-product-${productDetails.code}`"
+        v-if="!productDetails.hidden"
```
kialj876, line 11: not sure how other people feel about these, but I find it really helpful to have v-if/for's at the beginning

kialj876 (approved):

kialj876:
tests fails are my fault, sorry --- fixed now with that last merge

auth-web/src/components/auth/common/Product.vue
seeker25: sounds good I'll move it back

PR #2238, opened 2023-02-04
15100 mhr change product tile text

ChasBoucher:
*Issue #:*https://github.com/bcgov/entity/issues/15100
*Description of changes:*

- Add MHR product tile label and description for MHR product tile


auth-web/src/components/auth/common/Product.vue
```diff
@@ -180,23 +184,26 @@ export default class Product extends Vue {
...
+  }
   get productLabel () {
     // this is mapping product code with lang file.
     // lang file have subtitle and description with product code prefix.
     // eg: pprCodeSubtitle, pprCodeDescription
     // Also, returns check box icon and color if the product has been reviewed.
     const { code } = this.productDetails
-    let subTitle = `${code && code.toLowerCase()}CodeSubtitle` || ''
-    let details = `${code && code.toLowerCase()}CodeDescription` || ''
+    let mhrNotStaff = !this.isStaff && code.toLowerCase() === MhrProductCodeConstants.MHR.toLowerCase() ? MhrProductCodeConstants.NOT_STAFF : ''
```
dimak1, line 196: Is there a way to simplify this? From what I see here, we just want to show subtitle and details for non-staff roles. Maybe we could check for that condition and avoid all the default empty strings...

ChasBoucher: Subtitle and details are shown for all products.  

This logic builds a string for selecting from the lang file.  it is only applicable for MHR non staff.

auth-web/src/components/auth/common/Product.vue
```diff
@@ -180,23 +184,26 @@ export default class Product extends Vue {
   get showProductFee () {
     return this.canManageProductFee && this.orgProduct && this.orgProduct.product
   }
-
+  private get isStaff (): boolean {
+    return this.currentUser.roles.includes(Role.Staff)
```
seeker25, line 188: This is only for BC Reg staff? not SBC Staff?

auth-web/src/components/auth/common/Product.vue
kialj876: I think this would be a lot simpler if you set `code` to `mhrNotStaff` here and remove the MhrProductCodeConstants and all the code changes below. That way if we decide to refactor it later or add other 'staff' versions of descriptions its just one line of code to add to / alter

I'd also alter the default (mhrCodeDescription) to be the non staff version, and make this version mhrStaffCodeDescription (simpler I think when you're reading it the first time)

auth-web/src/components/auth/common/Product.vue
kialj876: also should use the account composable (or mixin which we will alter to composable when we update it to compostion-api) here instead of checking roles

auth-web/src/components/auth/common/Product.vue
kialj876: i.e. a future implementation to have other staff descriptions could be something like [...].includes(code) then code+= 'Staff'

ChasBoucher: Good suggestions.  I have updated code.

auth-web/src/components/auth/common/Product.vue
ChasBoucher: It is only for BC Reg Staff.
I have updated code to use Mixins

seeker25:
Two code smells to fix before merging please:

image

kialj876 (approved):
Nice work 👍

ChasBoucher:
@kialj876 @seeker25 Is there any action required by me here?  I noticed that the test failed base on, what looks like some timeout issue..

PR #2239, opened 2023-02-06
15245 - Change path regex, so it includes full urls. Remove redirectLoginFail…

seeker25:
…, it would be difficult to figure out which part of the URL is the redirectLoginFail based on the route.

*Issue #:*
https://github.com/bcgov/entity/issues/15245

*Description of changes:*

### DIFF
--- a/auth-web/src/util/common-util.ts
+++ b/auth-web/src/util/common-util.ts
@@ -8,7 +8,7 @@ import moment from 'moment'
 export default class CommonUtils {
   // checking url matches the regex
   static isUrl (value:string):boolean {
-    const URL_MATCHER = new RegExp('^(?:\\w+:)?\\/([^\\s\\.]+\\.\\S{2}|localhost[\\:?\\d]*)\\S*$')
+    const URL_MATCHER = new RegExp('^(?:\\w+:)?\\/([^\\s\\.]+\\.\\S{2}|[\\/]?localhost[\\:?\\d]*)\\S*$')
     return URL_MATCHER.test(value)
   }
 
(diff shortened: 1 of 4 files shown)


auth-web/src/util/common-util.ts
```diff
@@ -8,7 +8,7 @@ import moment from 'moment'
 export default class CommonUtils {
   // checking url matches the regex
   static isUrl (value:string):boolean {
-    const URL_MATCHER = new RegExp('^(?:\\w+:)?\\/\\/([^\\s\\.]+\\.\\S{2}|localhost[\\:?\\d]*)\\S*$')
+    const URL_MATCHER = new RegExp('^(?:\\w+:)?\\/([^\\s\\.]+\\.\\S{2}|localhost[\\:?\\d]*)\\S*$')
```
seeker25, line 11: so this works with http:/ and https:/ as well.

auth-web/src/util/common-util.ts
```diff
@@ -8,7 +8,7 @@ import moment from 'moment'
 export default class CommonUtils {
   // checking url matches the regex
   static isUrl (value:string):boolean {
-    const URL_MATCHER = new RegExp('^(?:\\w+:)?\\/\\/([^\\s\\.]+\\.\\S{2}|localhost[\\:?\\d]*)\\S*$')
+    const URL_MATCHER = new RegExp('^(?:\\w+:)?\\/([^\\s\\.]+\\.\\S{2}|[\\/]?localhost[\\:?\\d]*)\\S*$')
```
seeker25, line 11: image

Jxio (approved):

auth-web/src/util/common-util.ts
seeker25: Fails on:
`/confirmtoken/eyJpZCI6MTc1MSwidHlwZSI6IlNUQU5EQVJEIn0.Y-qrVg.jqfdfdfdfd`

PR #2277, opened 2023-03-13
Update nginx.conf - add in caching for OPTIONS

seeker25:
Update nginx.conf - add in caching for OPTIONS


seeker25:
Better way to fix this, just fix the code in the API that includes the header instead of just including it in the nginx config.

PR #2281, opened 2023-03-17
15688 - Add in invoice references to transactions screen + Fix UTC date, make it display PDT/PST. 

seeker25:
From customer feedback. 

https://github.com/bcgov/entity/issues/15688


seeker25:
Still need a fix for the date, doing that now

Jxio (approved):

auth-web/src/components/auth/account-settings/transaction/TransactionsDataTable.vue
```diff
@@ -194,7 +195,10 @@ export default defineComponent({
       }
     })
 
-    const displayDate = (val: Date) => CommonUtils.formatDisplayDate(val, 'MMMM DD, YYYY')
+    const displayDate = (val: Date) => {
+      const targetDate = val ? moment.utc(val).toDate() : val
```
kialj876, line 199: I think you need to check if 'val' has a timezone already not if it exists or not here? Now if I used this fn with a date that contained a non utc timezone it would give me the wrong time I think

auth-web/tests/unit/components/TransactionsDataTable.spec.ts
```diff
@@ -97,9 +98,10 @@ describe('TransactionsDataTable tests', () => {
     expect(row1Cells.at(3).text()).toContain('January 24, 20233:09 PM')
```
kialj876, line 98: can you alter the test setup data to not include the timezone in the date string? That way the test will secure this bug

auth-web/src/components/auth/account-settings/transaction/TransactionsDataTable.vue
seeker25: Yup, just being lazy i'll fix it

seeker25: I just used the `moment-timezone` library, makes managing dates and times way easier

auth-web/tests/unit/components/TransactionsDataTable.spec.ts
seeker25: sounds good

seeker25:
1 sec, something doesn't look right in the GUI even though the unit tests pass

auth-web/tests/unit/components/TransactionsDataTable.spec.ts
```diff
@@ -94,12 +95,13 @@ describe('TransactionsDataTable tests', () => {
     expect(row1Cells.at(0).find('span').text()).toBe('label1 value1')
     expect(row1Cells.at(1).text()).toBe('ab12')
     expect(row1Cells.at(2).text()).toBe('tester 123')
-    expect(row1Cells.at(3).text()).toContain('January 24, 20233:09 PM')
+    expect(row1Cells.at(3).text()).toContain('January 24, 20237:00 PM')
```
kialj876, line 98: you'll want to set this to 6:00 PM based on the setup value (its in january so the difference will be 8 hours)

seeker25: https://www.worldtimebuddy.com/

kialj876: yeah the time you are testing is in january not March so it is an 8 hour difference

kialj876: you shouldn't be commenting out that check -- it is what tests the time conversion

auth-web/tests/unit/test-utils/test-data/transaction.ts
```diff
@@ -6,7 +6,7 @@ export const transactionResponse: TransactionListResponse = {
     {
       businessIdentifier: '123',
       createdName: 'tester 123',
-      createdOn: new Date('2023-01-24T23:09:46+00:00'),
+      createdOn: new Date('2023-01-24T14:00:00'), // If we don't specify a timezone, it converts to PST.
```
kialj876, line 9: I think you should update this comment to: `The api does not provide the timezone information so the date display will set it to a UTC timezone`

auth-web/tests/unit/components/TransactionsDataTable.spec.ts
kialj876: sorry haha -- I see you updated it

kialj876 (approved):
nice work 👍

seeker25:
Pain in the neck, you'd think when you have a type of Date, and it returns from the API it would convert it somehow using new Date()

auth-web/tests/unit/components/TransactionsDataTable.spec.ts
```diff
@@ -140,7 +143,8 @@ describe('TransactionsDataTable tests', () => {
     expect(row1Cells.at(4).text()).toBe('123')
     expect(row1Cells.at(5).text()).toBe('ab12')
     expect(row1Cells.at(6).text()).toBe('tester 123')
-    expect(row1Cells.at(7).text()).toContain('January 24, 20233:09 PM')
+    // Input:createdOn: new Date('2023-01-24T14:00:00')
+    expect(row1Cells.at(7).text()).toContain('January 24, 20232:00 PM')
```
kialj876, line 147: Oh wait a minute... Shouldn't this read January 24, 20236:00 AM ? Since the input is 2PM UTC?

seeker25: No, because newDate converts it into PST

PR #2297, opened 2023-03-24
15593 - DRAFT PR - First initial changes for adding in groups for products.

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/15593

*Description of changes:*
Various changes to support adding keycloak roles for products.


seeker25:
WIP, will finish this off before Monday

auth-api/src/auth_api/services/membership.py
```diff
@@ -67,6 +67,10 @@ def as_dict(self):
         obj = membership_schema.dump(self._model, many=False)
         return obj
 
+    def get_owner_count(self, org):
```
seeker25, line 70: Moved this out so we could get rid of the dependency of `membership -> org`

auth-api/src/auth_api/services/org.py
```diff
@@ -688,9 +695,6 @@ def __delete_contact(org):
                 return contact
         return None
 
-    def get_owner_count(self):
```
seeker25, line 691: Moved this out so we could get rid of the dependency of membership -> org

seeker25:
Still need to test the migration - the flags might need some tweaking. 
Will give this a go on DEV.

auth-api/migrations/alembic.ini
```diff
@@ -45,6 +45,11 @@ args = (sys.stderr,)
 level = NOTSET
 formatter = generic
 
+[logger_auth_api]
+level = DEBUG
```
seeker25, line 49: We want `auth_api` logging in the migration logs, it's much easier and cleaner than using `print`

auth-api/src/auth_api/models/membership.py
```diff
@@ -149,14 +149,24 @@ def find_membership_by_user_and_org(cls, user_id, org_id) -> Membership:
 
     @classmethod
     def find_membership_by_userid(cls, user_id) -> Membership:
-        """Get the membership for the specified user and org."""
```
seeker25, line 152: I've created a bug ticket for this, it's used in a few spots - I don't think it's correct as it doesn't consider the org at all. 

https://github.com/bcgov/entity/issues/15756

auth-api/src/auth_api/services/products.py
```diff
@@ -259,3 +263,90 @@ def send_approved_product_subscription_notification(receipt_admin_emails, produc
...
+                UserModel.id,
+                UserModel.keycloak_guid,
+                ProductCodeModel.code,
+                ProductCodeModel.keycloak_group,
+                func.sum(active_subscription_case).label('active_subscription_count')
+        ).all()  # pylint: disable=comparison-with-callable
+
+        keycloak_group_subscriptions = []
+        for ups in user_subscriptions:
+            action = KeycloakGroupActions.ADD_TO_GROUP.value \
```
seeker25, line 331: There might have been a way to create a case to make this unnecessary, but I'm not entirely sure how to do nested cases in SQLAlchemy

auth-api/tests/docker/setup/demo-realm.json
```diff
@@ -368,6 +368,36 @@
       ],
       "clientRoles": {},
       "subGroups": []
+    },
+    {
+      "name": "ppr",
```
seeker25, line 373: Added these, so we could test against the docker keycloak.

auth-api/migrations/alembic.ini
```diff
@@ -45,6 +45,12 @@ args = (sys.stderr,)
 level = NOTSET
 formatter = generic
 
+[logger_auth_api]
```
seeker25, line 48: Didn't want to use print in migrations, would rather see code execute.

PR #2301, opened 2023-03-28
Fix get group id.

seeker25:


seeker25:
Time to goto bed

PR #2339, opened 2023-05-02
15241 minio upgrade fix

piahmed:
*Issue #:*
https://github.com/bcgov/entity/issues/15241

*Description of changes:*
auth-web defaults to sending an `Authorization` header for any request it makes - newer version of Minio does not like this header, and throws 400 errors. This change will not auto-add the header for urls with `minio` in them.

### DIFF
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.1.3",
+  "version": "2.1.4",
   "appName": "Auth Web",
   "sbcName": "SBC Common Components",
   "private": true,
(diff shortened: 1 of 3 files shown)


auth-web/src/util/http-util.ts
```diff
@@ -7,6 +7,11 @@ const axios = Axios.create()
 
 axios.interceptors.request.use(
   config => {
```
seeker25, line 9: I'd maybe recommend renaming this to `request` it's really not `config`

auth-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.1.3",
```
seeker25, line 3: run npm i after updating this, it will update the package-lock.json as well

seeker25 (approved):

pwei1018:
Temporary Url for review: https://bcregistry-account-dev--pr-2339-bvez1aqb.web.app

PR #2340, opened 2023-05-08
14438 update validation on NR number

eve-git:
*Issue #:*
https://github.com/bcgov/entity/issues/14438

*Description of changes:*
Re-write the NR number validation to make it consistent with other apps, such as business-edit-ui

### DIFF
--- a/auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
+++ b/auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
@@ -9,8 +9,8 @@
           label="Enter a Name Request Number"
           hint="Example: NR 1234567"
           :rules="nrNumberRules"
-          v-model="nrNumber"
-          @keyup="updateNrNumber()"
+          :value="nrNumber"
+          @input="nrNumber = updateNrNumber($event)"
           data-test="nr-number"
           autofocus
         />
@@ -94,8 +94,8 @@ import { StatusCodes } from 'http-status-codes'
   }
 })
 export default class AddNameRequestForm extends Vue {
-  private readonly currentOrganization!: Organization
-  private readonly addNameRequest!: (requestBody: CreateNRAffiliationRequestBody) => any
+  readonly currentOrganization!: Organization
+  readonly addNameRequest!: (requestBody: CreateNRAffiliationRequestBody) => any
 
   readonly helpDialogBlurb = 'If you have lost your receipt and name results email and ' +
     'need assistance finding your Name Request (NR) Number, please contact use at:'
@@ -115,18 +115,19 @@ export default class AddNameRequestForm extends Vue {
     v => this.isValidateEmail(v) || 'Email is Invalid'
   ]
 
-  private VALID_NR_FORMAT = new RegExp(/^(NR)?\s*(\d{7})$/)
+  VALID_NR_FORMAT = new RegExp(/^(NR)?\s*(\d{7})$/)
 
-  private nrNumber = ''
-  private applicantPhoneNumber = ''
-  private applicantEmail = ''
-  private isLoading = false
+  nrNumber = ''
+  applicantPhoneNumber = ''
+  applicantEmail = ''
+  isLoading = false
 
-  private updateNrNumber (): void {
-    this.nrNumber = this.nrNumber?.toUpperCase()
-    if (this.VALID_NR_FORMAT.test(this.nrNumber)) {
-      this.nrNumber = 'NR ' + this.VALID_NR_FORMAT.exec(this.nrNumber)[2]
+  updateNrNumber (value): string {
+    let nrNumber = value?.toUpperCase()
+    if (this.VALID_NR_FORMAT.test(value)) {
+      nrNumber = 'NR ' + this.VALID_NR_FORMAT.exec(value)[2]
     }
+    return nrNumber
   }
 
   $refs: {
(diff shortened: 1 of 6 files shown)


seeker25 (dismissed):

auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
```diff
@@ -143,6 +143,19 @@ export default class AddNameRequestForm extends Vue {
...
   }
 
+  private uppercase (prop: string): void {
+    this[prop] = this[prop]?.toUpperCase()
+  }
+
+  private isValidNrNumber (value: any): boolean {
+    const VALID_FORMAT = new RegExp(/^(NR)?\s*(\d{7})$/)
+    if (VALID_FORMAT.test(value)) {
+      this.nrNumber = 'NR ' + value.match(VALID_FORMAT)[2]
```
seeker25, line 153: I don't think this is a good idea on a validation function

auth-web/tests/unit/components/AddNameRequestForm.spec.ts
```diff
@@ -1,17 +1,33 @@
...
 const vuetify = new Vuetify({})
 
 // Prevent the warning "[Vuetify] Unable to locate target [data-app]"
 document.body.setAttribute('data-app', 'true')
 
 describe('Add Name Request Form', () => {
+  let wrapperFactory: any
+
+  beforeEach(() => {
+    wrapperFactory = (props: any) => {
```
seeker25, line 18: Why not set wrapper = wrapperFactory? 

instead of doing `const wrapper = wrapperFactory()` 6x

auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
```diff
@@ -5,12 +5,12 @@
...
         <v-text-field
           ref="nrNumber"
-          filled persistent-hint validate-on-blur
+          filled persistent-hint
           label="Enter a Name Request Number"
           hint="Example: NR 1234567"
           :rules="nrNumberRules"
           v-model="nrNumber"
-          @blur="formatNrNumber()"
+          @keyup="uppercase('nrNumber')"
```
seeker25, line 13: Softly typed strings are fairly prone to errors

seeker25 (changes requested):
Validation function shouldn't be setting state

auth-web/tests/unit/components/AddNameRequestForm.spec.ts
eve-git: updated

auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
eve-git: It is the same requirement as in business-edit-ui, which is ticket 14438 for.

auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
```diff
@@ -129,7 +138,7 @@ export default class AddNameRequestForm extends Vue {
   private isFormValid (): boolean {
     return !!this.nrNumber &&
       (!!this.applicantPhoneNumber || !!this.applicantEmail) &&
-      CommonUtils.validateNameRequestNumber(this.nrNumber)
+      this.isValidNrNumber(this.nrNumber)
```
seeker25, line 141: Much better, this way you don't have any side effects if you want to check the validation of a NR

seeker25 (approved):

auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
```diff
@@ -115,11 +114,21 @@ export default class AddNameRequestForm extends Vue {
...
   private nrNumber = ''
   private applicantPhoneNumber = ''
   private applicantEmail = ''
   private isLoading = false
 
+  @Watch('nrNumber', { immediate: true })
+  private updateNrNumber (): void {
+    this.nrNumber = this.nrNumber?.toUpperCase()
+    if (this.VALID_NR_FORMAT.test(this.nrNumber)) {
+      this.nrNumber = 'NR ' + this.nrNumber.match(this.VALID_NR_FORMAT)[2]
```
seeker25, line 128: `Use the "RegExp.exec()" method instead.` saw this from code cov

eve-git:
@seeker25 have to replace @Watch with @keyup. In the case the input box has a string, and then paste a new nr number, such as 'nr1234567', @watch won't invoke 'updateNrNumber' and update the nr number to normal format.

seeker25:
Probably because your REF and v-model have the same name?

seeker25:
If watch doesn't work

```
  :value="searchText"
  @input="searchText = $event.target.value"
```

image

Should work? I'll try it myself

auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
```diff
@@ -173,7 +187,7 @@ export default class AddNameRequestForm extends Vue {
     }
   }
 
-  private resetForm (event: string): void {
+  resetForm (event: string): void {
```
seeker25, line 190: We don't need the private public protected accessors within a vue component. It's just extra words that clutter the code.

auth-web/src/components/auth/manage-business/AddNameRequestForm.vue
```diff
@@ -5,12 +5,12 @@
...
           ref="nrNumber"
-          filled persistent-hint validate-on-blur
+          filled persistent-hint
           label="Enter a Name Request Number"
           hint="Example: NR 1234567"
           :rules="nrNumberRules"
-          v-model="nrNumber"
-          @blur="formatNrNumber()"
+          :value="nrNumber"
+          @input="nrNumber = updateNrNumber($event)"
```
seeker25, line 13: This will work on any input, versus keyup / keydown. I tested by pasting in a NR without touching the keyboard.

seeker25:
Should be good after the CI goes through

PR #2360, opened 2023-06-01
auth-web: move to business v2 end point

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/16611
*Description of changes:*
move to business v2 end point

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the sbc-auth license (Apache 2.0).

PR #2362, opened 2023-06-06
Revert "auth-web: move to business v2 end point"

seeker25:
Reverts bcgov/sbc-auth#2360


seeker25:
Reverting this back until https://github.com/bcgov/entity/issues/16658 is done

Jxio (approved):

PR #2365, opened 2023-06-19
16621 added business search instead of business input

severinbeauvais:
*Issue #:*
https://github.com/bcgov/entity/issues/16621

*Description of changes:*
- app version = 2.2.0
- added bcrs shared enums package
- added vue-debounce-decorator package
- copied Business Lookup component from shared repo
- copied business lookup services from Create UI
- fixed incorrect imports (removed file extensions)
- added Registries Search env variables
- added business search component
- added business name and number
- updated visibility of some properties and methods
- customized Business Search component per UI design
- added unit tests
- prevent event propagation (Enter key form submit bug)
- added FF for business search (else fall back to old code)
- changed icon when searching (in addition to existing "loading" behaviour)


auth-web/package.json
```diff
@@ -18,6 +18,7 @@
   "dependencies": {
     "@bcrs-shared-components/bread-crumb": "^1.0.2",
     "@bcrs-shared-components/corp-type-module": "1.0.9",
+    "@bcrs-shared-components/enums": "1.0.39",
```
severinbeauvais, line 21: For CorpTypeCd and EntityStates.

auth-web/package.json
```diff
@@ -35,6 +36,7 @@
     "register-service-worker": "^1.6.2",
     "sbc-common-components": "3.0.8",
     "vue": "2.6.12",
+    "vue-debounce-decorator": "^1.0.1",
```
severinbeauvais, line 39: To debounce the search function (to limit the number of calls).

auth-web/src/assets/scss/theme.scss
```diff
@@ -68,6 +68,7 @@ $app-blue: #1669bb; // same as the Vuetify theme primary
 $app-lt-blue: #E2E8EE; // LOW OPACITY PRIMARY BLUE
 $app-background-blue: #E4EDF7; // Hover Background Blue
 $app-dk-blue: #38598a;
+$app-green: #2E8540;
```
severinbeauvais, line 71: For Added label, ie.

image

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
```diff
@@ -10,7 +10,7 @@
       v-model="dialog"
       persistent
       scrollable
-      max-width="675"
+      max-width="50rem"
```
severinbeauvais, line 13: To allow business id column to fit without wrapping, ie.

image

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
```diff
@@ -431,6 +429,16 @@ export default class AddBusinessDialog extends Vue {
...
 
+dl {
+  line-height: 2rem;
+}
+
+// pair up terms and definitions
+dt {
+  float: left;
+  clear: left;
+}
```
severinbeauvais, line 440: Ie,

image

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,197 @@
...
+      :search-input.sync="searchField"
+    >
+      <!-- NB: use empty selection slot to stop re-triggering selectedBusiness watcher -->
+      <template v-slot:selection="" />
+
+      <template v-slot:prepend-item>
+        <p class="pt-2 pl-5 font-size-12">
+          Active B.C. Businesses
+        </p>
+      </template>
```
severinbeauvais, line 28: Ie,

image

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,197 @@
...
+
+      <template v-slot:no-data>
+        <p class="pl-5 font-weight-bold">
+          No active B.C. business found
+        </p>
+        <p class="pl-5">
+          Ensure you have entered the correction business name or numbers.
+        </p>
+      </template>
```
severinbeauvais, line 37: Ie,

image

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,197 @@
...
+enum States {
+  INITIAL = 'initial',
+  SEARCHING = 'searching',
+  SHOW_RESULTS = 'show results',
+  NO_RESULTS = 'no results',
+  SUMMARY = 'summary'
+}
+
+@Component({})
+export default class BusinessLookup extends Vue {
```
severinbeauvais, line 72: This is almost the same as the shared component, but with some UI differences. It would be a medium-size job to combine them, so I didn't.

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,197 @@
...
+    // safety check
+    if (this.searchField && this.searchField.length > 2) {
+      this.state = States.SEARCHING
+      const searchStatus = null // search all (ACTIVE + HISTORICAL)
+      this.searchResults = await BusinessLookupServices.search(this.searchField, searchStatus).catch(() => [])
+
+      // enable or disable items according to whether they have already been added
+      for (let result of this.searchResults) {
+        result.disabled = await this.isAffiliated(result.identifier)
+      }
```
severinbeauvais, line 116: Ie,

image

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,197 @@
...
+.result-identifier,
+.result-name {
+  font-size: $px-16;
+
+  // limit col width and show an ellipsis for long names:
+  max-width: 485px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
```
severinbeauvais, line 183: Ie,

image

auth-web/src/services/business-lookup.services.ts
```diff
@@ -0,0 +1,61 @@
+import { BusinessLookupResultIF } from '@/models'
+import ConfigHelper from '@/util/config-helper'
+import { SessionStorageKeys } from '@/util/constants'
+import { axios } from '@/util/http-util'
+
+/**
+ * Class that provides integration with the BusinessLookup API.
+ */
+export default class BusinessLookupServices {
```
severinbeauvais, line 9: There is no shared component that provides this service. This file is almost exactly the same as the one in Create UI and Edit UI (except for how the URL is fetched).

severinbeauvais:
NB: if you see the following error in your dev console then it's one of your browser extensions. (In my case, it was LastPass. Disabling it got rid of the error.)

seeker25:
High level looks good to me - lets try it?

/gcbrun

severinbeauvais:
This PR is ready for review. See ticket for uxpin. Note that I added the Business Lookup component in the "Add an Existing Business" dialog in place of the business id input. (Dave, Sarah et al are aware.) This is an interim measure until the other tickets in the story are implemented, and then it will look like Janis' designs.

Travis, one thing I noticed and don't love is that the Business Search API key (and the Canada Post key, LaunchDarkly key, etc) are in clear-text in the network traffic. (The other UIs do the same thing.) But I think it does expose us to "key borrowing" (at least for a user with the right KC token). Thoughts on future improvements? Not needed?

auth-web/package.json
```diff
@@ -17,8 +17,9 @@
   },
   "dependencies": {
     "@bcrs-shared-components/bread-crumb": "^1.0.2",
-    "@bcrs-shared-components/corp-type-module": "1.0.9",
-    "@bcrs-shared-components/interfaces": "1.0.28",
+    "@bcrs-shared-components/corp-type-module": "1.0.12",
+    "@bcrs-shared-components/enums": "1.0.39",
+    "@bcrs-shared-components/interfaces": "1.0.63",
```
severinbeauvais, line 22: For Business Lookup IFs.

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,179 @@
+<template>
+  <!-- once in Summary state, need to re-mount to reuse this component -->
```
severinbeauvais, line 2: The alternative would be to pass in a prop that gets toggled to reset the state, etc. Or a public method that the parent calls.

At the moment, there's no need to reset the state except between instances of the Add Business dialog, so I didn't complicate this.

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,179 @@
...
+    // event data is automatically returned
+  }
+}
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/theme.scss';
+
+.font-size-12 {
+  font-size: 12px;
```
severinbeauvais, line 133: This is the only class of its kind so I didn't put it in base.scss. If we get more of these then we should move them.

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,179 @@
...
+  .added {
+    color: $app-green;
+  }
+}
+
+// prevent Magnify icon from being rotated when list is displayed
+::v-deep .v-input__icon .mdi-magnify {
+  -webkit-transform: none !important;
+  transform: none !important;
+}
```
severinbeauvais, line 178: Normally, the icon is flipped when the v-menu (list) is shown, which makes sense when the icon is an up triangle or down triangle. It doesn't make sense when the icon is a magnifying glass 🙄

pwei1018:
Temporary Url for review: https://bcregistry-account-dev--pr-2365-4bteghcw.web.app

seeker25:
"Travis, one thing I noticed and don't love is that the Business Search API key (and the Canada Post key, LaunchDarkly key, etc) are in clear-text in the network traffic. (The other UIs do the same thing.) But I think it does expose us to "key borrowing" (at least for a user with the right KC token). Thoughts on future improvements? Not needed?"

I think the Canada Post key could be proxied so it isn't exposed. What happens if you try to use that key from a different domain? Does it still work? 

Canada Post people could use the key to flood requests for looking up addresses?

Launch darkly - I can't see why someone would use the key for anything? 

VUE_APP_REGISTRIES_SEARCH_API_KEY - Maybe it's possible to encrypt this?

auth-web/package-lock.json
severinbeauvais, line 1: There's a lot of changes on here because I need a `npm ci` (clean install). Also I update the browserslist.

seeker25: This may break alot of stuff

pwei1018:
Temporary Url for review: https://bcregistry-account-dev--pr-2365-4bteghcw.web.app

auth-web/package-lock.json
severinbeauvais: Would you prefer I revert those changes? Or drag this UI kicking and screaming into the present?

seeker25: Probably better to just update what you need - unless you want to let Ashna or Riyaz know to retest all of auth-web

auth-web/src/services/business-lookup.services.ts
seeker25: Should we make them the same? so we can refactor into bcrs-shared-components (add it to the pile haha)

auth-web/package-lock.json
seeker25: You shouldn't need to do a clean install for the deps you installed.

severinbeauvais:
> I think the Canada Post key could be proxied so it isn't exposed. What happens if you try to use that key from a different domain? Does it still work?

Yes, it could be proxied through the Auth or Legal API, but a logged-in user (or API user?) could still query things using our (paid) key.

> VUE_APP_REGISTRIES_SEARCH_API_KEY - Maybe it's possible to encrypt this?

@pwei1018 Any thoughts on  key security?

auth-web/package-lock.json
severinbeauvais: On it...

pwei1018:
Temporary Url for review: https://bcregistry-account-dev--pr-2365-4bteghcw.web.app

auth-web/tests/unit/components/AddBusinessDialog.spec.ts
```diff
@@ -115,17 +115,30 @@ tests.forEach(test => {
...
 
     it(test.desc, async () => {
-      wrapper.setData({ businessIdentifier: test.businessIdentifier })
+      wrapper.setData({
+        businessIdentifier: test.businessIdentifier,
+        businessName: 'My Business Inc'
+      })
       await flushPromises()
 
+      // console.log(wrapper.html())
```
seeker25, line 124: Remove this please

auth-web/tests/unit/services/business-lookup.service.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it('does not return a result when the business is not found', async () => {
+    // mock unsuccesssful search
+    const args = `${BUSINESS_SEARCH_URL}businesses/search/facets?start=0&rows=20&categories=legalType` +
+      `:BC,A,ULC,C,S,XP,GP,LP,CUL,XS,LLC,LL,BEN,CP,CC,XL,FI,XCP,PA&query=value:FM1000003`
+    sinon.stub(axios, 'get').withArgs(args).returns(
+      Promise.resolve({ data: { searchResults: { results: [] } } })
+    )
+
+    // search and look at results
```
seeker25, line 41: Maybe redundant comment?

auth-web/src/services/business-lookup.services.ts
severinbeauvais: It would require some refactoring to pass in the URL, API key and account id, but it could be done. (Any thoughts on the params?)

How about if I look into that after this PR?

seeker25:
Enter key issues, redirects you to a new page - I think it should not redirect? I notice when you have a valid business highlighted and press enter, doesn't redirect

auth-web/package-lock.json
severinbeauvais: Done but I still see some library changes in there. I tried it a couple of different ways so 🤷‍♂️

auth-web/src/services/business-lookup.services.ts
kialj876: If we want it to be the same code across business search / ppr / auth / entities then we should use the search composable (it's the only one that will support the functionality for all the UIs)

kialj876: for vue 3 implementation you would only have to import it, set the env var, set the specific filters (i.e. legal type etc.), use the results array it provides, and trigger the search function so it would slim down the code a lot too

kialj876:
> "Travis, one thing I noticed and don't love is that the Business Search API key (and the Canada Post key, LaunchDarkly key, etc) are in clear-text in the network traffic. (The other UIs do the same thing.) But I think it does expose us to "key borrowing" (at least for a user with the right KC token). Thoughts on future improvements? Not needed?"
> 
> I think the Canada Post key could be proxied so it isn't exposed. What happens if you try to use that key from a different domain? Does it still work?
> 
> Canada Post people could use the key to flood requests for looking up addresses?
> 
> Launch darkly - I can't see why someone would use the key for anything?
> 
> VUE_APP_REGISTRIES_SEARCH_API_KEY - Maybe it's possible to encrypt this?

The business search api key is fine. It's a public key (there's 0 auth around what it gives access to)

auth-web/src/components/auth/manage-business/BusinessLookup.vue
severinbeauvais: Hey, class isn't bad :)

Yes, a single code source would be better, but I didn't want to get into that here (and I don't want to refactor Create UI or Edit UI either).

How about if I create a tech debt ticket?

auth-web/src/services/business-lookup.services.ts
severinbeauvais: This UI isn't at Vue 3 yet, so could we use a common implementation? Or should we wait?

auth-web/tests/unit/components/AddBusinessDialog.spec.ts
severinbeauvais: Done.

severinbeauvais:
> Enter key issues, redirects you to a new page - I think it should not redirect? I notice when you have a valid business highlighted and press enter, doesn't redirect

Fixed.

seeker25 (approved):

auth-web/src/components/auth/manage-business/BusinessLookup.vue
seeker25: It's ok, we have a bunch of those already hahaha

auth-web/src/services/business-lookup.services.ts
kialj876: I think we should wait and then sort out the common one after/during the upgrade

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
```diff
@@ -40,19 +40,23 @@
...
-              v-model="businessIdentifier"
-              @blur="formatBusinessIdentifier()"
-              class="business-identifier mb-n2"
-              aria-label="Incorporation Number and Password or Passcode"
-              autofocus
+            <!-- Business Identifier and Name -->
+            <!-- NB: use v-if to re-mount component between instances -->
+            <BusinessLookup
+              v-if="dialog"
+              @business="businessName = $event.name; businessIdentifier = $event.identifier"
```
JazzarKarim, line 47: Can we set the event to one variable called `business` and then we can call the name with `business.name` for example? What do you think Sev?

pwei1018:
Temporary Url for review: https://bcregistry-account-dev--pr-2365-4bteghcw.web.app

auth-web/src/components/auth/manage-business/BusinessLookup.vue
kialj876: yeah that's fine with me 👍

auth-web/package-lock.json
seeker25: It's ok, worth a shot

JazzarKarim (approved):
Lots of learning for me here 😄 Very helpful since I might also be picking up a ticket in Auth Web in the near future. Awesome stuff Sev!

seeker25:
Merge when you're ready

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
severinbeauvais: Yes, that would surely work, but I wanted to change as little as possible in this component (since I may have to FF this) and `businessIdentifier` is used in a lot of places in here.

auth-web/src/components/auth/manage-business/BusinessLookup.vue
severinbeauvais: New tech debt ticket is https://github.com/bcgov/entity/issues/16821.

auth-web/src/services/business-lookup.services.ts
severinbeauvais: New tech debt ticket is https://github.com/bcgov/entity/issues/16821.

auth-web/src/services/business-lookup.services.ts
```diff
@@ -0,0 +1,62 @@
...
+   * @returns a promise to return the search results
+   */
+  static async search (query: string, status = ''): Promise<BusinessLookupResultIF[]> {
+    const legalType = 'BC,A,ULC,C,S,XP,GP,LP,CUL,XS,LLC,LL,BEN,CP,CC,XL,FI,XCP,PA'
+
+    let url = this.registriesSearchApiUrl + 'businesses/search/facets?start=0&rows=20'
+    url += `&categories=legalType:${legalType}${status ? '::status:' + status : ''}`
+    url += `&query=value:${encodeURIComponent(query)}`
+
+    return axios.get(url, {
```
severinbeauvais, line 44: @seeker25 @kialj876 @argush3 @vysakh-menon-aot 

I'm wondering about the near future, where we need to implement a NR Search service and API ([ticket 16714](https://app.zenhub.com/workspaces/entities-team-space-6143567664fb320019b81f39/issues/gh/bcgov/entity/16714)), which will likely be implemented in Namex API...

And I'm wondering about the need to have config keys for the URL and API key...

Do we want some consistency/symmetry between the Business Search calls and the NR Search calls? Would it make sense to proxy one or both via Auth API?

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -0,0 +1,180 @@
...
+      :items="searchResults"
+      :loading="state == States.SEARCHING"
+      :name="Math.random()"
+      :search-input.sync="searchField"
+      append-icon="mdi-magnify"
+      autocomplete="chrome-off"
+      autofocus
+      class="mt-5 mb-n2"
+      filled
+      hint="For example: &quot;Joe's Plumbing Inc.&quot;, &quot;BC1234567&quot;, &quot;FM1234567&quot;"
```
severinbeauvais, line 15: This is how you get double quotes in a string in the template (without resorting to binding a variable in the script).

auth-web/src/services/business-lookup.services.ts
severinbeauvais: Travis, Kial, OK if I resolve this conversation?

auth-web/src/components/auth/manage-business/AddBusinessDialog.vue
severinbeauvais: Karim, OK if I resolve this conversation?

JazzarKarim: Yes Sev! I was trying to resolve it myself but I have no access to do so 🤣

Jxio (approved):
LGTM!

auth-web/src/services/business-lookup.services.ts
kialj876: I don't think it makes sense to do add an endpoint in auth that just relays data everyone can already access directly. Both calls are fully public so there's no reason to verify the token either (if we needed that then we would have to do it in the search-api/namex-api as well anyways).

severinbeauvais: OK, thanks.

Do you think we'll need to verify the token (or key) for a NR search via SOLR?

auth-web/src/util/constants.ts
```diff
@@ -291,26 +291,27 @@ export enum Permission {
 
 export enum LDFlags {
```
severinbeauvais, line 292: I sorted this list.

pwei1018:
Temporary Url for review: https://bcregistry-account-dev--pr-2365-4bteghcw.web.app

bennypc (approved):
LGTM

auth-web/tests/unit/components/AddBusinessDialog.spec.ts
```diff
@@ -115,17 +115,28 @@ tests.forEach(test => {
...
+      wrapper.setData({
+        businessIdentifier: test.businessIdentifier,
+        businessName: 'My Business Inc'
+      })
       await flushPromises()
 
       // verify components
       expect(wrapper.attributes('id')).toBe('add-business-dialog')
       expect(wrapper.find('#add-business-dialog').isVisible()).toBe(true)
+      expect(wrapper.find('businesslookup-stub').exists()).toBe(true)
```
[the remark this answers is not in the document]
seeker25: `business-lookup-stub` :P

auth-web/src/services/business-lookup.services.ts
severinbeauvais: I think this conversation is complete (but feel free to add anything pertinent).

A follow-up conversation is continuing in the [NR search API ticket](https://github.com/bcgov/entity/issues/16714).

PR #2369, opened 2023-06-22
16527 - check token cache endpoint before generating new service account token

bolyachevets:
*Issue #: https://github.com/bcgov/entity/issues/16527

*Description of changes: added a service account token caching service in gcp as a POC: https://github.com/bolyachevets/token_cache

Looking for some feedback.

### DIFF
--- a/auth-api/src/auth_api/services/rest_service.py
+++ b/auth-api/src/auth_api/services/rest_service.py
@@ -164,7 +164,7 @@ class RestService:
         return response
 
     @staticmethod
-    @cache.cached()
+    @cache.cached(query_string=True)
     def get_service_account_token(config_id='KEYCLOAK_SERVICE_ACCOUNT_ID',
                                   config_secret='KEYCLOAK_SERVICE_ACCOUNT_SECRET') -> str:
         """Generate a service account token."""
--- a/auth-api/src/auth_api/services/rest_service.py
+++ b/auth-api/src/auth_api/services/rest_service.py
@@ -30,6 +30,7 @@ from urllib3.util.retry import Retry
 
 from auth_api.exceptions import ServiceUnavailableException
 from auth_api.utils.enums import AuthHeaderType, ContentType
+from auth_api.utils.cache import cache
 
 RETRY_ADAPTER = HTTPAdapter(max_retries=Retry(total=5, backoff_factor=1, status_forcelist=[404]))
 
@@ -163,11 +164,13 @@ class RestService:
         return response
 
     @staticmethod
+    @cache.cached()
     def get_service_account_token(config_id='KEYCLOAK_SERVICE_ACCOUNT_ID',
                                   config_secret='KEYCLOAK_SERVICE_ACCOUNT_SECRET') -> str:
         """Generate a service account token."""
         kc_service_id = current_app.config.get(config_id)
         kc_secret = current_app.config.get(config_secret)
+
         issuer_url = current_app.config.get('JWT_OIDC_ISSUER')
         token_url = issuer_url + '/protocol/openid-connect/token'
         auth_response = requests.post(token_url, auth=(kc_service_id, kc_secret), headers={
--- a/auth-api/src/auth_api/utils/cache.py
+++ b/auth-api/src/auth_api/utils/cache.py
@@ -18,17 +18,17 @@ from flask_caching import Cache
 cache_servers = os.environ.get('CACHE_MEMCACHED_SERVERS')
 
 if cache_servers:
-    cache = Cache(config={  'CACHE_TYPE': 'MemcachedCache',
-                            'CACHE_MEMCACHED_SERVERS': cache_servers.split(',') })
+    cache = Cache(config={'CACHE_TYPE': 'MemcachedCache',
+                          'CACHE_MEMCACHED_SERVERS': cache_servers.split(',')})
 else:
 
     redis_host = os.environ.get('CACHE_REDIS_HOST')
     redis_port = os.environ.get('CACHE_REDIS_PORT')
 
     if redis_host and redis_port:
-        cache = Cache(config={  'CACHE_TYPE': 'RedisCache',
-                                'CACHE_REDIS_HOST': redis_host,
-                                'CACHE_REDIS_PORT': redis_port })
+        cache = Cache(config={'CACHE_TYPE': 'RedisCache',
+                              'CACHE_REDIS_HOST': redis_host,
+                              'CACHE_REDIS_PORT': redis_port})
 
     else:
         cache = Cache(config={'CACHE_TYPE': 'simple'})  # pylint: disable=invalid-name
--- a/auth-api/src/auth_api/utils/cache.py
+++ b/auth-api/src/auth_api/utils/cache.py
@@ -12,8 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Bring in the common cache."""
+import os
 from flask_caching import Cache
 
+cache_servers = os.environ.get('CACHE_MEMCACHED_SERVERS')
 
-# lower case name as used by convention in most Flask apps
-cache = Cache(config={'CACHE_TYPE': 'simple'})  # pylint: disable=invalid-name
+if cache_servers:
+    cache = Cache(config={  'CACHE_TYPE': 'MemcachedCache',
+                            'CACHE_MEMCACHED_SERVERS': cache_servers.split(',') })
+else:
+
+    redis_host = os.environ.get('CACHE_REDIS_HOST')
+    redis_port = os.environ.get('CACHE_REDIS_PORT')
+
+    if redis_host and redis_port:
+        cache = Cache(config={  'CACHE_TYPE': 'RedisCache',
+                                'CACHE_REDIS_HOST': redis_host,
+                                'CACHE_REDIS_PORT': redis_port })
+
+    else:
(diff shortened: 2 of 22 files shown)


seeker25:
Interesting - why not just do a memory cache and save yourself the RTT?

seeker25 (approved):

seeker25 (commented):
Please fix the unit tests / lints before merging

seeker25:
Also, maybe a better idea to repoint keycloak token fetching to just use your caching service instead? Thoughts?

bolyachevets:
> Maybe a suggestion? Repoint keycloak token fetching to just use your caching service instead (like wrapping keycloak)? Maybe through the apigee gateway? @thorwolpert -^ maybe you have some ideas?

right now it is just an 'in memory' cache, since i only had in mind caching a few service account tokens

seeker25:
Cool whichever way you want to do it, you have my support

pwei1018 (approved):

thorwolpert:
Is this for service accounts, or token caching in GCP?
if it's for caching in GCP land, why wouldn't we use memorystore (redis or memchache)?
does it have any similarities to the POC with IDIM?
https://github.com/bcgov/entity/blob/main/rfcs/rfc-3rd-party-otp.md

bolyachevets:
> Is this for service accounts, or token caching in GCP? if it's for caching in GCP land, why wouldn't we use memorystore (redis or memchache)? does it have any similarities to the POC with IDIM? https://github.com/bcgov/entity/blob/main/rfcs/rfc-3rd-party-otp.md

the immediate usage is to cache keycloak client service account tokens used on the openshift side, e.g. in queues

thorwolpert:
Those a re long running services, could you just use the local cache service?
like
[api/namex/services/cache.py](https://github.com/bcgov/namex/blob/main/api/namex/services/cache.py)
which uses the pallets cache project?
https://github.com/pallets-eco/cachelib

bolyachevets:
> Those a re long running services, could you just use the local cache service? like [api/namex/services/cache.py](https://github.com/bcgov/namex/blob/main/api/namex/services/cache.py) https://github.com/bcgov/namex/blob/main/api/namex/resources/statistics/wait_time_statistics.py#L34
> 
> which uses the pallets cache project? https://github.com/pallets-eco/cachelib
> 
> Not sure if it meets your needs, but has a nice uage: `@cache.cached(timeout=14400) # cached for 4 hours`
> 
> I use cachelib to cache the JWKS, so we're not pulling the key and cert from keycloak all the time when validating the incoming JWT. Seems like it would work for caching service-acct JWTs in a similar fashion

- One of the requirements was to cache tokens outside of app context
- These are today's counts for number of times we called 'protocol/openid-connect/token' endpoint for each client.

-------------06/22/2023-------------
api-key-account-6923 : 60
api-key-account-4637-sandbox : 14
business-search-service-account : 786
cso-service-account : 1367
entity-notebook-service-account : 188
name-request-service-account : 24447
lear_sa : 40930
api-key-account-3963 : 3
name-request : 55
terraform : 8
api-key-account-4637 : 50
api-key-account-14355 : 36
ppr-service-account : 2837
api-key-account-7140 : 53
api-key-account-5401 : 4
api-key-account-13301-sandbox : 36
api-key-account-6873 : 1
api-key-account-5797-sandbox : 3
entity-service-account : 42951
ppr-document-delivery-service-account : 1043
sbc-auth-admin : 1149
api-key-account-5841 : 39
api-key-account-6833 : 10
api-key-account-5797 : 102
api-key-account-5838 : 54
namex-solr-service-account : 1963
api-key-account-6502 : 31

thorwolpert:
which cachelib does, as it's used here to not keep generating the stat.
It's what the library and cache were designed to do.
image

thorwolpert:
It caches across calls (contexts), https://flask-caching.readthedocs.io/en/latest/
as stated before, that's how we cache the JWKS. If you look it's not getting JWKS for every JWT across every user/session/context.

thorwolpert:
i meant the service itself should add in cachelib or flask-caching.
you are building a backing store, and for the long running processes they don't need it.
for shortlived things, like knative containers, the backing store could be redis/mamcache, but still use cachelib/flask-cache in the service and doesn't need to know more, as it becomes a deployment config/decision

bolyachevets:
> i meant the service itself should add in cachelib or flask-caching. you are building a backing store, and for the long running processes they don't need it. for shortlived things, like knative containers, the backing store could be redis/memcache, but still use cachelib/flask-cache in the service and doesn't need to know more, as it becomes a deployment config/decision

Just to clarify, I thought you meant that hooking up a caching library directly to RestService token fetching method would cache token values across calls (as in the most recent submission), or do you mean that I should use a caching library for the token fetching route running in gcp

thorwolpert:
If I were adding caching to PPR or Lear, I would add cachelib or flask-caching to those services.
If they are running in OCP, I'm probably done.
If they are running in GCP, I'd change the config to point the caching store to redis or memcache.

bolyachevets:
> memcache

I tried setting up memorystore in gcp with redis and memcached options. redis estimate is > 100$ a month, memcached is > 40$ a month.

I guess we would be using for more than just caching a couple of keycloak tokens. I did a check in cache.py to see if redis/memcached memorystores are configured.

thorwolpert:
I was not suggesting you do that.
It should be done in the client, not as a newly created service.

bolyachevets:
> I was not suggesting you do that. It should be done in the client, not as a newly created service.
what do you mean by doing it in the client?
e.g., to instantiate flask-caching client via memorycached backend you need to provide a list of server addresses. where would these servers come from, if not as gcp memorystore services?

thorwolpert:
Like it's done in NameX api, auth, JWKS, etc. the cachelib is in the API.  
When running ling services in OCP, long jobs in GCP, then the simple-cache is fine. For short lived services like CloudRun, then you'd change the config in that container service to use redis/memcache. 
Building a competitor to redis/memcache or creating a new module/library for the apps isn't something we want to do.

bolyachevets:
> cachelib

I may be not understanding this completely, but I don't think I created a competitor to redis/memcache. Looking at comments in cachelib's memcached.py you can see that it needs a memcached library to instantiate:     ```
The first argument can either be an object that resembles the API of a
    :class:`memcache.Client` or a tuple/list of server addresses. In the
    event that a tuple/list is passed, Werkzeug tries to import the best
    available memcache library.

    This cache looks into the following packages/modules to find bindings for
    memcached:

        - ``pylibmc``
        - ``google.appengine.api.memcached``
        - ``memcached``
        - ``libmc``
```
Similar for redis.

What I did is modify the cache loader file https://github.com/bcgov/sbc-auth/blob/1be82e530036052d47d6b1f2d2e93e94c951d4a7/auth-api/src/auth_api/utils/cache.py to check whether memcache/redis relevant parameters have been set up in the config. If they are, then appropriate backend would be loaded, otherwise if would stick with simple cache (which has been used in the repos you mentioned, and what is fine for the purpose of caching service account tokens as well). When I spoke with @pwei1018 , I was under the impression that we want to make sure this will work if we decide to use memcache/redis in gcp. I tried using memcache/redis in gcp, and the backend would fail to load unless proper library is available (i.e.  for memcache these would be    - ``pylibmc``
        - ``google.appengine.api.memcached``
        - ``memcached``
        - ``libmc`` ) In summary, nothing changed in the behavior of cache - it is still using simple cache, but is also robust to use mecache or redis.

**Comment by seeker25** [comment]
After closer inspection, I don't think we're reinventing the wheel here, looks like the same pathway of Namex. 

Platform folks (SSO team) are seeing a spike of keycloak activity, have notified Andriy about it - I think we should tweak this PR a bit and get it rolling. 

We might also want to create service accounts specific for each application so we can trace these down in the future. EG. LEGAL-API uses the same service account as AUTH.

File: auth-api/src/auth_api/utils/cache.py [hunk 1]
```diff
@@ -12,8 +12,23 @@
...
 # limitations under the License.
 """Bring in the common cache."""
+import os
 from flask_caching import Cache
 
+cache_servers = os.environ.get('CACHE_MEMCACHED_SERVERS')
 
-# lower case name as used by convention in most Flask apps
-cache = Cache(config={'CACHE_TYPE': 'simple'})  # pylint: disable=invalid-name
+if not cache_servers is None:
```
seeker25, line 20: Double negative

auth-api/src/auth_api/services/rest_service.py
```diff
@@ -163,11 +164,13 @@ def get(endpoint, token=None,  # pylint: disable=too-many-arguments
         return response
 
     @staticmethod
+    @cache.cached()
```
seeker25, line 167: needs to be tied to `CACHE_DEFAULT_TIMEOUT`?

seeker25: nvm:

image

Jxio (approved):

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2369-httbbpif.web.app

seeker25:
Good stuff, not sure why you ran /gcbrun, only api changes

bolyachevets:
> Good stuff, not sure why you ran /gcbrun, only api changes

for some reason gcp check was listed as one of the conditions for the green checkmark

seeker25:
yeah we usually just ignore it haha

PR #2377, opened 2023-07-04
testing

severinbeauvais:
This works but needs a bit of cleanup.

Also it would be nice if the colour transitioned back to normal (instead of "flicking" off).


auth-web/src/assets/scss/base.scss
```diff
@@ -184,3 +184,8 @@ legend {
 .gray-9 {
   color: $gray9;
 }
+
+.background-green {
+  background: $app-background-green;
+  border-color: $app-background-green;
+}
```
severinbeauvais, line 191: I declared this here so BaseVDataTable can see it (even though we reference it in AffiliatedEntityTable).

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -467,10 +469,11 @@ export default class EntityManagement extends Mixins(AccountMixin, AccountChange
...
   async showAddSuccessModalNR () {
     this.$refs.addNRDialog.close()
-    this.dialogTitle = 'Name Request Added'
-    this.dialogText = 'You have successfully added a name request'
     await this.syncBusinesses()
-    this.$refs.successDialog.open()
+    this.highlightRowIndex = 1
+    setTimeout(() => {
+      this.highlightRowIndex = NaN
+    }, 4000)
```
severinbeauvais, line 476: So, THIS component keeps track of which row to highlight and for how long. (How to style it is AffiliatedEntityTable's job.) These are not the VDataTable's concerns.

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -20,6 +20,8 @@ const affiliations = (reactive({
 }) as unknown) as AffiliationState
 
 export const useAffiliations = () => {
+  const highlightRowIndex = NaN
```
severinbeauvais, line 23: Not needed, will remove.

PR #2383, opened 2023-07-10
16220 - Account Authorization request UI component

hfekete:
Issue #:16220
https://github.com/bcgov/entity/issues/16220

Adding a component for requesting access to account (account authorization request)


auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
```diff
@@ -0,0 +1,105 @@
...
+      type: String,
+      required: true
+    },
+    businessIdentifier: {
+      type: String,
+      required: true
+    }
+  },
+  setup () {
+    const accounts = ref([])
```
seeker25, line 69: Use reactive if possible here, gets rid of .value and is a bit cleaner

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
```diff
@@ -0,0 +1,105 @@
...
+    return {
+      selectedAccount,
+      requestAccessMessage,
+
+      accounts,
+      isLoading
+    }
+  },
+  mounted () {
+    OrgService.getOrganizationsNameAndUuidByAffiliation(this.businessIdentifier).then(
```
seeker25, line 83: I'd recommend using async await pattern, that way you don't run into callback hell
http://callbackhell.com/

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
```diff
@@ -0,0 +1,105 @@
...
+  emits: ['changeRequestAccessMessage', 'selectAccount'],
+  methods: {
+    emitSelected: function (uuid) {
+      const selectedAcc = this.accounts.find(acc => acc.uuid === uuid)
+      this.emitSelectedAccount(selectedAcc)
+    },
+    emitSelectedAccount: function (selectedAcc) {
+      this.$emit('selectAccount', selectedAcc)
+    }
+
```
seeker25, line 56: spacing

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
```diff
@@ -0,0 +1,105 @@
...
+    return {
+      selectedAccount,
+      requestAccessMessage,
+
+      accounts,
+      isLoading
+    }
+  },
+  mounted () {
```
seeker25, line 82: Please use the composition-api life cycle hooks:

https://vuejs.org/api/composition-api-lifecycle.html

`onMounted()`

seeker25: mounted() is for options api I believe

seeker25 (changes requested):
See comments above

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
hfekete: Is it OK if I stick with refs here ?

What I know reactive can only be used on objects, and here I am declaring additionally string and bool which are not objects for which reactive wont work.

I think mixing them might confuse the reader what is reactive and what not, as some will have '.value' (bool and string) while the objects wont.

If it is still recommended, I will update it for sure. 

Thanks!

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
hfekete: Will do, excellent point, thanks!

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
kialj876: similarly the methods should be defined/initialized in the setup function. This looks like a mix of options-api and composition-api to me. When this moves to vue3 all of the boiler will go away as we'll be using script setup so we want it all in there

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
```diff
@@ -0,0 +1,105 @@
...
+
+          if (this.accounts.length === 1) {
+            this.selectedAccount = this.accounts[0]
+            this.emitSelectedAccount(this.selectedAccount)
+          }
+        }
+      }
+    ).catch(
+      err => {
+        console.log(err)
```
kialj876, line 96: is there anything in the design for the error flow? This will need to show the user an indication that something failed and they need to retry

auth-web/tests/unit/components/AccountAuthorizationRequest.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+    expect(wrapper.find('#account-authorization-request-request-account-select').exists()).toBe(true)
+    expect(wrapper.find('#account-authorization-request-additional-message-textarea').exists()).toBe(true)
+    expect(wrapper.find('h3').exists()).toBe(false)
+
+    // verify that account is selected and selector disabled
+    expect(wrapper.find('#account-authorization-request-request-account-select').attributes().disabled).toBeDefined()
+  })
+
+  it('renders enabled select with no preselected item, when multiple affiliated accounts found', async () => {
+    const get = sandbox.stub(axios, 'get')
```
kialj876, line 105: can you make a function for this that returns the stub/wrapper for each test?

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
seeker25: Yeah use reactive, build an object

```
const person = reactive({
  name: 'Albert',
  age: 30,
  isNinja: true,
});
```

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
```diff
@@ -0,0 +1,105 @@
...
+
+<script lang='ts'>
+import { defineComponent, ref } from '@vue/composition-api'
+import OrgService from '@/services/org.services'
+
+export default defineComponent({
+  name: 'AccountAuthorizationRequest',
+  emits: ['changeRequestAccessMessage', 'selectAccount'],
+  methods: {
+    emitSelected: function (uuid) {
```
kialj876, line 49: move these into the setup function pls

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
seeker25: Anyways I'll leave it up to you

auth-web/tests/unit/components/AccountAuthorizationRequest.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import AccountAuthorizationRequest from '@/components/auth/manage-business/AccountAuthorizationRequest.vue'
+import Vuetify from 'vuetify'
+import Vuex from 'vuex'
+import { axios } from '@/util/http-util'
+import flushPromises from 'flush-promises'
+import sinon from 'sinon'
+
+Vue.use(Vuetify)
+Vue.use(Vuex)
```
seeker25, line 18: I'd recommend setting it up like this:
https://github.com/bcgov/business-edit-ui/blob/main/tests/unit/setup.ts
https://github.com/bcgov/business-edit-ui/blob/0007c737e64ce1ae1aee2db1341da922e039f089/jest.config.js#L11

So we don't have a bunch of imports and Vue use at the top of each unit test

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
kialj876: you can also return `{ ...toRefs(state) }` so that you don't need to reference 'state' in the html

kialj876: not a big deal though I think. In vue3 I generally just declare them as refs since you don't need to return each variable individually when using script setup

kialj876: can you type the accounts array if possible though?

hfekete: Updated to have single reactive object (componentState). Let me know if this is not what you had in mind. 

Just as a footnote, ref is actually recommended way to declare a reactive vars, if we want to revise this approach in the future. 
https://vuejs.org/guide/essentials/reactivity-fundamentals.html
I did not know that before today though :D 

Thanks!

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
```diff
@@ -0,0 +1,98 @@
...
+        }
+      } else {
+        console.error('OrgService.getOrganizationsNameAndUuidByAffiliation', response.status)
+      }
+      componentState.isLoading = false
+    })
+
+    return {
+      emitSelected,
+
```
seeker25, line 93: maybe remove this extra line spacing - eventually we're going to steal these and put them in auth-web 
https://github.com/bcgov/business-edit-ui/blob/0007c737e64ce1ae1aee2db1341da922e039f089/.eslintrc.js#L7

that way you can just npm run lint, and it should fix alot of code formatting problems like this

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
hfekete: Should I revert it back maybe ? I do not mind doing that as ref() looks nicer for me

hfekete: I'll leave it with reactive for now. Let me know if there are any changes around the componentState variable I added need to be changed/updated

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
hfekete: moved

auth-web/src/components/auth/manage-business/AccountAuthorizationRequest.vue
hfekete: No, but I discussed it with UX and they will provide some path for what/how to handle the errors. Lets add that in separate PR as I think that parent modal might need to handle the error, rather then component itself.

auth-web/tests/unit/components/AccountAuthorizationRequest.spec.ts
hfekete: done

auth-web/tests/unit/components/AccountAuthorizationRequest.spec.ts
hfekete: I tried to see if I can quickly update current setup, but run in some issues. 
So like you suggested, lets worry later

Also, maybe we can connect and see what approach I should do for the next UI parts I will need to add. I could possibly update this test file retroactively too.

seeker25 (approved):

seeker25:
I think it's looking pretty good, just the error handling to resolve, but you can resolve that when you use the component

PR #2389, opened 2023-07-18
16825 Display Name Request type

leodube-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/16825

*Description of changes:*
- Added Name Request type to the table
- Feature Flagged it with `enable-name-request-type`


auth-web/src/util/constants.ts
```diff
@@ -247,6 +247,48 @@ export enum NrEntityType {
     INFO = 'INFO', // special value for sub-menu
 }
 
+export enum NrRequestType {
```
leodube-aot, line 250: I had to guess a bit at which of the [enum NR request types](https://github.com/bcgov/bcrs-shared-components/blob/main/src/enums/name-request-codes.ts) corresponded with the request types in the [ux pin](https://preview.uxpin.com/a86dccdd20ac62828f965c5eea10f81b828a9853#/pages/163299967/simulate/sitemap). There are also a couple types I can across when testing that I am not sure about, so the code will default to not displaying a type (for now). These codes I am unsure about are: DBA, GP, XCP.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -51,7 +51,8 @@
 
       <!-- Type -->
       <template v-slot:item-slot-Type="{ item }">
-        <div class="gray-9 font-weight-bold">{{ type(item) }}</div>
+        <div class="gray-9 font-weight-bold type-text">{{ type(item) }}</div>
+        <div v-if="enableNameRequestType && isNameRequest(item)" class="gray-9 font-weight-bold type-text ml-1">{{ nameRequestType(item) }}</div>
         <div>{{ typeDescription(item) }}</div>
       </template>
```
leodube-aot, line 58: Adding the NR type in a separate `div` is a little weird here, but it's to avoid an issue with the Type filter. When the NR type is included in the same `div`, the filter treats each unique NR type as a new item to filter on, resulting in the screenshot below. Notice "Name Request - Incorporation" and "Name Request - Registration". Separating them into different `div`s is a (not great) workaround. If anyone has experience with `v-data-table` filters and knows a better fix, let me know.
 
Screenshot 2023-07-18 073104

severinbeauvais: How does this look?

Did you look at affiliations-factory.ts::updateFilter() ? Maybe you could tweak that to match all Name Requests regardless of entity type ...?

pwei1018:
Temporary Url for review: https://bcregistry-account-dev--pr-2389-vke4mdx1.web.app

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: I'm not sure what filtering looks at, either, but your solution works:

image

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -154,6 +154,16 @@ export const useAffiliations = () => {
     )
   }
 
+  const nameRequestType = (business: Business): string => {
+    if (isNameRequest(business)) {
+      const nrType: string = NrRequestType[business.nameRequest?.entityTypeCd]
+      if (nrType) {
+        return '\u2014 ' + nrType
```
severinbeauvais, line 161: Here's how Filings UI does the dash, fwiw (it's self-documenting):

https://github.com/bcgov/business-filings-ui/blob/fafe0c3af7da84fa91f3677f785cfe5c7f0adb8b/src/stores/rootStore.ts#L228

auth-web/src/util/constants.ts
severinbeauvais: Nice digging to find the shared enum!

Does this help?
https://github.com/bcgov/namerequest/blob/feature-way-of-navigating/src/enums/entity-type.ts

severinbeauvais: This app isn't consistent with enum naming, but it should be plural.

JazzarKarim:
Leo, there's a test that's failing in `AffiliatedEntityTable.spec.ts` on line 136 (https://github.com/leodube-aot/sbc-auth/blob/1fd72226e780f6b897bdd350949f3c5ee3b48442/auth-web/tests/unit/components/AffiliatedEntityTable.spec.ts#L136).

failing test

There's an extra space that's being received. This probably rose from your change to the `AffiliatedEntityTable.vue` file.

severinbeauvais:
> Leo, there's a test that's failing in `AffiliatedEntityTable.spec.ts` on line 136 (https://github.com/leodube-aot/sbc-auth/blob/1fd72226e780f6b897bdd350949f3c5ee3b48442/auth-web/tests/unit/components/AffiliatedEntityTable.spec.ts#L136).
> 
> _snipped_
> 
> There's an extra space that's being received. This probably rose from your change to the `AffiliatedEntityTable.vue` file.

I recommend separate `toContain` calls for the first part and the second part.

auth-web/src/composables/affiliations-factory.ts
leodube-aot: Ah that's much cleaner, I'll use that.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -435,6 +445,10 @@ export default defineComponent({
       .type-column {
         min-width: 12rem;
       }
+
+      .type-text {
+        display: inline-block;
```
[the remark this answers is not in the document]
leodube-aot: Good call, I'll try the Vuetify display helpers

auth-web/src/util/constants.ts
leodube-aot: Sev, that link helps me understand what each code means, but I'm still not sure how the entity codes map onto the NR types from the ux pin: Registration, Incorporation, Continuation In, Reactivation, Amalgamation, Name Change, and Alteration.

Also I'll make the enum plural.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
leodube-aot: I just had a look at `affiliations-factory.ts::updateFilter()` and it doesn't seem to get called when the popup appears, it only gets called once the Type is selected. The problem must be happening at some point before that, or as part of the vuetify component itself.

severinbeauvais: I think it's OK as you have it, but a code comment would be ideal.

auth-web/src/util/constants.ts
```diff
@@ -247,6 +247,48 @@ export enum NrEntityType {
...
+  // Incorporation
+  BC = 'Incorporation', // BC Benefit Company Incorporation
+  CC = 'Incorporation', // CCC - Incorporation/Amalgamation
+  CP = 'Incorporation', // Cooperative - Incorporation/Amalgamation
+  CR = 'Incorporation', // BC Company - Incorporation/Amalgamation
+
+  // Registration
+  FR = 'Registration', // Sole Proprietorship/General Partnership/DBA - Registration
+  UL = 'Registration', // Unlimited Liability Company
+  XCR = 'Registration', // Corporation (Foreign) - Extrapro Registration/Amalgamation
```
severinbeauvais, line 281: I think FR ("firm") is the only registration here. The others are incorporations.

auth-web/src/util/constants.ts
```diff
@@ -247,6 +247,48 @@ export enum NrEntityType {
...
+
+  // Restoration
+  RCC = 'Restoration',
+  RCP = 'Restoration',
+  RCR = 'Restoration',
+  RSO = 'Restoration',
+  BERE = 'Restoration',
+  RUL = 'Restoration',
+}
```
severinbeauvais, line 290: I think you should add the "convert" (or is it resubmit?) codes here as well.

auth-web/src/util/constants.ts
severinbeauvais: These look good to me but I'm not an expert. I recommend you confirm these with Mihai.

Also, these should be called out in your ticket under a "Test Notes" comment so that QA knows what to verify.

severinbeauvais: OK, here's another link for ya:
https://github.com/bcgov/namex/blob/7e5c88a86a0585a404bafc549a8755b74264b06a/api/namex/services/lookup/name_request_filing_actions.py#L51

seeker25: @chenhongjing -^ samething I sent you lol

seeker25:
Please fix the code smells (by logging in and setting them to ignore) before merging

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: I'm OK to resolve this conversation.

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -154,6 +154,16 @@ export const useAffiliations = () => {
     )
   }
 
+  const nameRequestType = (business: Business): string => {
+    if (isNameRequest(business)) {
+      const nrType: string = NrRequestType[business.nameRequest?.entityTypeCd]
```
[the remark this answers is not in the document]
severinbeauvais: Nice change, thanks.

auth-web/src/util/common-util.ts
```diff
@@ -268,4 +268,9 @@ export default class CommonUtils {
   static trimTrailingSlashURL (url) {
     return (url) ? url.trim().replace(/\/+$/, '') : ''
   }
+
+  /** use entity type to get NR type from enum */
+  static mapEntityTypeToNrType (entityTypeCd: string): string {
+    return NrRequestTypes[entityTypeCd]
```
severinbeauvais, line 274: What happens here if entityTypeCd is undefined (which it would be the optional chaining on affiliations-factory.ts line 161 fails)?

(I'm not actually sure whether that would cause an error here. You may need a check here.)

auth-web/src/util/constants.ts
severinbeauvais: Yes, please ask.

auth-web/src/util/constants.ts
```diff
@@ -247,6 +247,48 @@ export enum NrEntityType {
...
+  CCCT = 'Continuation In',
+  CTC = 'Continuation In',
+  CT = 'Continuation In',
+  ULCT = 'Continuation In',
+
+  // Alteration
+  BECV = 'Alteration',
+  CCV = 'Alteration',
+  BECR = 'Alteration',
+  UC = 'Alteration',
```
severinbeauvais, line 270: 1. check the other enum to identify which codes are conversions
2. handle them according to what we are discussing in Standup

Do you know if there are NR request  types not in this enum?

leodube-aot: Looking at the [namex](https://github.com/bcgov/namex/blob/7e5c88a86a0585a404bafc549a8755b74264b06a/api/namex/services/lookup/name_request_filing_actions.py#L51) file there are some that are not in this enum. The ones I saw are: Assumed Name and Reinstatement.

auth-web/src/util/common-util.ts
leodube-aot: I just tested by manually setting it to undefined and it does not cause any errors. It basically will just show "Name Request" with no emdash or type appended.

severinbeauvais: Awesome. It probably returns undefined, and the code in affiliations-factory check for non-falsy nrType so it's safe 👍

chenhongjing: I don't think it's proper to use a mapping from `entityTypeCd` to nr type. Instead, it should be from `requestTypeCd` to nr type. Because some codes like `BEC`, `BERE` etc. only exit in `requestTypeCd`. So please tests more cases like alternation, restoration, amalgamation etc. to see if it shows incorrect nr type.

leodube-aot: I see yeah that could be an issue. The `business`/`business.nameRequest` object returned from auth-api does not appear to have a `requestTypeCd`. Do you know where the `requestTypeCd` get's returned from?

chenhongjing: Yeah it's the same question I got yesterday. Please see the code in the following link. The response that fetches affiliations in fact returns `requestTypeCd`, but it's not mapped into `business.nameRequest`. And also check Network for the details.
https://github.com/bcgov/sbc-auth/blob/main/auth-web/src/store/modules/business.ts#L137

severinbeauvais: I'm not following this conversation 100%, but I do want to point out that NRs have different codes than LEAR. There are tables in namex and Legal API to convert from the NR code to the LEAR codes. Please be clear about which codes you are trying to convert to a string (or applying logic to).

severinbeauvais: > I see yeah that could be an issue. The `business`/`business.nameRequest` object returned from auth-api does not appear to have a `requestTypeCd`. Do you know where the `requestTypeCd` gets returned from?

I believe the NR object has a property named requestTypeCd (or request_type_cd).

leodube-aot: > Yeah it's the same question I got yesterday. Please see the code in the following link. The response that fetches affiliations in fact returns `requestTypeCd`, but it's not mapped into `business.nameRequest`. And also check Network for the details. https://github.com/bcgov/sbc-auth/blob/main/auth-web/src/store/modules/business.ts#L137

I see yeah checking the Network tab did reveal that. Mapping it into `business.nameRequest` worked for me!

Brief explanation Sev, when I was using `entityTypeCd`, it would return some codes from this [NrRequestTypeCodes enum](https://github.com/bcgov/bcrs-shared-components/blob/main/src/enums/name-request-codes.ts) (which could be mapped to an NR type string), as well as some codes from this [EntityType enum](https://github.com/bcgov/namerequest/blob/feature-way-of-navigating/src/enums/entity-type.ts) (which it isn't clear what NR type to map it too). Using `requestTypeCd` seems to map each Name Request to an appropriate NR type.

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -154,6 +155,18 @@ export const useAffiliations = () => {
     )
   }
 
+  /** Returns the Name Request type using the entity type code */
+  const nameRequestType = (business: Business): string => {
+    if (isNameRequest(business)) {
+      const nrType: string = CommonUtils.mapRequestTypeCdToNrType(business.nameRequest?.requestTypeCd)
```
severinbeauvais, line 161: You don't need to declare it as a string because Typescript already knows that from the function return type. (And sometimes eslint will complain about a redundant type declaration.)

auth-web/src/util/common-util.ts
```diff
@@ -268,4 +268,9 @@ export default class CommonUtils {
   static trimTrailingSlashURL (url) {
     return (url) ? url.trim().replace(/\/+$/, '') : ''
   }
+
+  /** use entity type to get NR type from enum */
+  static mapRequestTypeCdToNrType (requestTypeCd: string): string {
```
severinbeauvais, line 273: Update comment.

severinbeauvais: Also, requestTypeCd has a type (not string).

auth-web/src/models/affiliation.ts
```diff
@@ -63,6 +63,7 @@ export interface NameRequestResponse {
   folioNumber?: string
   target?: NrTargetTypes
   entityTypeCd?: string
+  requestTypeCd?: string
```
severinbeauvais, line 66: Use type (enum) instead of string.

Also in business.ts below.

auth-web/src/util/constants.ts
```diff
@@ -247,6 +247,87 @@ export enum NrEntityType {
...
+  BERE = 'Restoration', // BC Benefit Company Incorporation
+  RCC = 'Restoration', // Community Contribution Co.
+  RCP = 'Restoration', // Cooperative
+  RCR = 'Restoration', // B.C. Company
+  RFI = 'Restoration', // Financial Institution (BC)
+  RSO = 'Restoration',
+  RUL = 'Restoration', // Unlimited Liability Co.
+  XRCP = 'Restoration', // Extraprovincial Cooperative
+  XRUL = 'Restoration' // Extraprovincial Unlimited Liability Company
+}
```
severinbeauvais, line 329: Looks GGRREEAATT! Do you think you should pass this by Mihai or are you pretty confident in it?

auth-web/src/store/modules/business.ts
```diff
@@ -147,6 +147,7 @@ export default class BusinessModule extends VuexModule {
           folioNumber: nr.folioNumber,
           target: getTarget(nr),
           entityTypeCd: nr.entityTypeCd,
+          requestTypeCd: nr.requestTypeCd,
```
severinbeauvais, line 150: Nice 👍

auth-web/src/util/constants.ts
```diff
@@ -247,6 +247,87 @@ export enum NrEntityType {
     INFO = 'INFO', // special value for sub-menu
 }
 
+export enum NrRequestTypes {
```
severinbeauvais, line 250: These aren't actually NR Request Types -- they're the strings for the types. (Just trying to avoid future confusion.)

auth-web/src/util/constants.ts
severinbeauvais: I'm OK to resolve this conversation if you are.

auth-web/src/util/common-util.ts
severinbeauvais: I'm OK to resolve this conversation if you are.

auth-web/src/util/constants.ts
leodube-aot: Would a name like NrRequestTypeStrings make more sense then, or would a comment suffice?

auth-web/src/util/common-util.ts
```diff
@@ -268,4 +269,9 @@ export default class CommonUtils {
   static trimTrailingSlashURL (url) {
     return (url) ? url.trim().replace(/\/+$/, '') : ''
   }
+
+  /** use NR request type code to get NR type from enum */
+  static mapRequestTypeCdToNrType (requestTypeCd: NrRequestTypeCodes): string {
```
severinbeauvais, line 274: Technically you're returning `NrRequestTypes` (not string), but I prefer to consider it a string since you're displaying it. To be 100% correct you should cast (`as string`) what you're returning below.

severinbeauvais (approved):
LGTM (save a few small edits).

As always, please get some other approvals before merging.

I don't remember if I asked: can this merge to main branch or should it go to the feature branch?

auth-web/src/util/constants.ts
severinbeauvais: Well, it's not a list of codes and it's not a list of types, so yes, even though it's wordy.

[Naming things is one of the hard things in computer science.](https://blog.frankel.ch/hard-things-computer-science/)

auth-web/src/util/constants.ts
leodube-aot: I'm fairly confident but I will check with Mihai if he wants to have a look over it

auth-web/src/util/common-util.ts
severinbeauvais: Comment is still incorrect.

auth-web/src/util/constants.ts
seeker25: LOL -^ I was just talking about this the other day with Cam

auth-web/src/util/common-util.ts
leodube-aot: This comment was changed to "use NR request type code to get NR type from enum" a few commits ago. My latest commit fixes the comment for the nameRequestType function in affiliations-factory.ts

JazzarKarim (approved):
LGTM! 👍

seeker25:
Looks like this needs a resync - Also Severin's question needs to be answered before merging to main:

"I don't remember if I asked: can this merge to main branch or should it go to the feature branch?"

severinbeauvais:
> "I don't remember if I asked: can this merge to main branch or should it go to the feature branch?"

This was answered offline. The answer is "main branch" as this new functionality is protected by a FF.

leodube-aot:
Rebase completed and additional code smells marked as false positive (due to additional entries into the `NRRequestTypeStrings` enum). Unless anyone catches any more missing/incorrect NR type code mappings this should be good to merge.

severinbeauvais:
> Unless anyone catches any more missing/incorrect NR type code mappings this should be good to merge.

LGTM but it would be nice to get a review from someone else experience in this repo (@seeker25 @Jxio @kial).

seeker25 (approved):

seeker25:
Arg - I just realized.. probably not the best timing to merge

seeker25:
Cancelled the CD flow

PR #2431, opened 2023-08-09
Branch sync with main

seeker25:


seeker25:
Will create merge commit this time instead of squashing. We can't rebase otherwise I'd have to force push to the branch, which would cause alot of people headaches.

PR #2508, opened 2023-08-31
17134 - open modal on action button.

hfekete:
*Issue #:17134
https://github.com/bcgov/entity/issues/17134

*Description of changes:*
adding trigger on a button for opening modal when "open new request" is clicked on affiliation.

### DIFF
--- a/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
+++ b/auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
@@ -153,6 +153,7 @@
             @remove-business="$emit('remove-business', $event)"
             @business-unavailable-error="$emit('business-unavailable-error', $event)"
             @resend-affiliation-invitation="$emit('resend-affiliation-invitation', $event)"
+            @popup-manage-business-dialog="$emit('popup-manage-business-dialog', $event)"
           />
         </template>
       </base-v-data-table>
@@ -187,7 +188,7 @@ export default defineComponent({
     highlightIndex: { default: -1 }
   },
   emits: ['add-unknown-error', 'remove-affiliation-invitation', 'remove-business',
-    'business-unavailable-error', 'resend-affiliation-invitation'],
+    'business-unavailable-error', 'resend-affiliation-invitation', 'popup-manage-business-dialog'],
   setup () {
     const isloading = false
     const { loadAffiliations, affiliations, entityCount, clearAllFilters,
--- a/auth-web/src/components/auth/manage-business/EntityManagement.vue
+++ b/auth-web/src/components/auth/manage-business/EntityManagement.vue
@@ -555,6 +555,7 @@ import AffiliationInvitationService from '@/services/affiliation-invitation.serv
 import { AffiliationInvitationStatus } from '@/models/affiliation'
 import AuthorizationEmailSentDialog from './AuthorizationEmailSentDialog.vue'
 import { Base64 } from 'js-base64'
+import { Business } from '@/models/business'
 import BusinessService from '@/services/business.services'
 import ConfigHelper from '@/util/config-helper'
 import ExpandableHelp from '@/components/auth/common/ExpandableHelp.vue'
@@ -567,7 +568,6 @@ import PasscodeResetOptionsModal from '@/components/auth/manage-business/Passcod
 import SearchBusinessNameRequest from './SearchBusinessNameRequest.vue'
 import StartNewBusinessHelp from '@/components/auth/manage-business/StartNewBusinessHelp.vue'
 import { appendAccountId } from 'sbc-common-components/src/util/common-util'
-import { Business } from '@/models/business'
 
 @Component({
   components: {
@@ -803,7 +803,7 @@ export default class EntityManagement extends Mixins(AccountMixin, AccountChange
     await this.syncBusinesses()
   }
 
-  async popupBusinessDialog(business:Business) {
+  async popupBusinessDialog (business:Business) {
     this.initialBusinessIdentifierManageBusinessDialog = business.businessIdentifier
     this.showManageBusinessDialog = true
   }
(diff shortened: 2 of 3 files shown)


auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -611,6 +614,7 @@ export default class EntityManagement extends Mixins(AccountMixin, AccountChange
   snackbarText: string = null
   showSnackbar = false
   timeoutMs = 4000
+  initialBusinessIdentifierManageBusinessDialog = null
```
seeker25, line 617: I think we can use the businessIdentifier?

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -153,6 +153,7 @@
             @remove-business="$emit('remove-business', $event)"
             @business-unavailable-error="$emit('business-unavailable-error', $event)"
```
seeker25, line 154: I was talking with Rodrigo, we should probably gun these and use the store instead

Later work though

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -799,6 +803,11 @@ export default class EntityManagement extends Mixins(AccountMixin, AccountChange
     await this.syncBusinesses()
   }
 
+  async popupBusinessDialog(business:Business) {
+    this.initialBusinessIdentifierManageBusinessDialog = business.businessIdentifier
```
seeker25, line 807: possible to just use this.businessIdentifier?

auth-web/src/components/auth/manage-business/EntityManagement.vue
```diff
@@ -294,6 +295,7 @@
         :isStaffOrSbcStaff="isStaffAccount || isSbcStaffAccount"
         :userFirstName="currentUser.firstName"
         :userLastName="currentUser.lastName"
+        :initial-business-identifier="initialBusinessIdentifierManageBusinessDialog"
```
seeker25, line 298: `this.businessIdentifier`?

auth-web/src/components/auth/manage-business/EntityManagement.vue
hfekete: if there are multiple affiliation requests for different business in the same account it will have multiple possible business identifiers. So I am passing this one to open it for specific (this is 2ndary action button on affiliation table)

auth-web/src/components/auth/manage-business/EntityManagement.vue
hfekete: if there are multiple affiliation requests for different business in the same account it will have multiple possible business identifiers. So I am passing this one to open it for specific (this is 2ndary action button on affiliation table)

So I am not 100% sure.

where does this.businessIdentifier come from ?

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
hfekete: Agreed its not nice. 
I usually used event bus. I usually would use this: [mitt](https://github.com/developit/mitt).

In vue2 there is an option of emitting to the $root and then listening on it again (not 100% sure of syntax) but that is not supported in vue3. So would avoid that.

And I think adding every event to the store might get cumbersome, fast.
Tracking events with emitt is easy as you can just keep event names in enum/const.

auth-web/src/components/auth/manage-business/EntityManagement.vue
hfekete: I dont know, I could name it like that, but am afraid someone might think it is default for the page or smth.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
seeker25: We can just make a specific store for entities, I think it would make it quite a bit easier - just used on the business dashboard

auth-web/src/components/auth/manage-business/EntityManagement.vue
seeker25: Shouldn't you be setting a list then?

auth-web/src/components/auth/manage-business/EntityManagement.vue
seeker25: `:initial-business-identifier="businessIdentifier"`

auth-web/src/components/auth/manage-business/EntityManagement.vue
seeker25: Business Dialog can only handle 1 business identifier at a time I think?

auth-web/src/components/auth/manage-business/EntityManagement.vue
seeker25: Line 602

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
seeker25: I was looking at something like this perhaps:

https://runthatline.com/pinia-watch-state-getters-inside-vue-components/

seeker25: I still think it's important to use emit and props - but maybe we can shift a bit over to the store so it's less complicated?

hfekete: I usually use store for complex states, but for events, like I mentioned above, I use mitt library.
But yeah, if you want to swap it in the future, I wont mind :)
(I mean all the events are getting propagated, as the third level one is handling popups and all.

auth-web/src/components/auth/manage-business/EntityManagement.vue
hfekete: Yeah, but this variable is used just for starting it up, and is defined in the EntityManagement vue. Where there might be multiple businessIdentifiers. But sure, I can rename it to businessIdentifier

auth-web/src/components/auth/manage-business/EntityManagement.vue
hfekete: Ok, yeah, am setting businessIdentifier  too in my event.

seeker25 (approved):

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2508-2yh85q69.web.app

PR #2514, opened 2023-09-01
Adding same ability as on branch to be able to create affiliation invitation using toOrgUuid.

hfekete:
*Description of changes:*
Adding same ability as on branch to be able to create affiliation invitation using toOrgUuid.

### DIFF
--- a/auth-api/src/auth_api/services/affiliation_invitation.py
+++ b/auth-api/src/auth_api/services/affiliation_invitation.py
@@ -219,7 +219,7 @@ class AffiliationInvitation:
             return to_org.id
 
         except DataError as err:
-            raise BusinessException(Error.INVALID_INPUT, err)
+            raise BusinessException(Error.INVALID_INPUT, err) from err
 
     @staticmethod
     @user_context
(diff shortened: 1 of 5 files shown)


ochiu (commented):
Can you add tests that makes use of this new attribute toOrgUuid?

auth-api/src/auth_api/services/affiliation_invitation.py
```diff
@@ -207,6 +208,19 @@ def _get_invitation_email(affiliation_invitation_info: Dict,
 
         return affiliation_invitation_info.get('recipientEmail', None)
 
+    @staticmethod
+    def _get_org_id_from_org_uuid(to_org_uuid):
+        try:
```
seeker25, line 213: I think you're fine without a try raise here - but that's minor  - this is already in the feature branch - approving

seeker25 (approved):

seeker25:
Please fix linting and add a parameter to the unit test as per Ody's comment before merging.

ochiu (approved):

PR #2521, opened 2023-09-05
17385 - Fix amalgamation NR display in MBR

leodube-aot:
*Issue #:*
https://github.com/bcgov/entity/issues/17385

*Description of changes:*

- Use the action code for amalgamation to display Name Request type.

*Notes:*
While the `requestTypeCd` returned from LEAR can be used for the NR type in most cases, in this situation we have to use the `requestActionCd`. This is because for _Incorporation_ and _Amalgamation_ NR types for a BC Limited Company, the `requestTypeCode` will be "CR" in both cases (so we can't differentiate using this), however the `requestActionCd`s will be "NEW" and "AML" respectively. I couldn't find a general solution using just `requestActionCd`s as many of the codes have dual uses.

### DIFF
--- a/auth-web/src/composables/affiliations-factory.ts
+++ b/auth-web/src/composables/affiliations-factory.ts
@@ -161,7 +161,7 @@ export const useAffiliations = () => {
   const nameRequestType = (business: Business): string => {
     if (isNameRequest(business)) {
       // Try action code first, and if not found in the enum then use type code
-      let nrType = CommonUtils.mapRequestActionCdToNrType(business.nameRequest?.requestActionCd)
+      let nrType: string = CommonUtils.mapRequestActionCdToNrType(business.nameRequest?.requestActionCd)
       nrType = nrType || CommonUtils.mapRequestTypeCdToNrType(business.nameRequest?.requestTypeCd)
       if (nrType) {
         const emDash = '—' // ALT + 0151
--- a/auth-web/src/composables/affiliations-factory.ts
+++ b/auth-web/src/composables/affiliations-factory.ts
@@ -162,7 +162,7 @@ export const useAffiliations = () => {
     if (isNameRequest(business)) {
       // Try action code first, and if not found in the enum then use type code
       let nrType = CommonUtils.mapRequestActionCdToNrType(business.nameRequest?.requestActionCd)
-      nrType = nrType ? nrType : CommonUtils.mapRequestTypeCdToNrType(business.nameRequest?.requestTypeCd)
+      nrType = nrType || CommonUtils.mapRequestTypeCdToNrType(business.nameRequest?.requestTypeCd)
       if (nrType) {
         const emDash = '—' // ALT + 0151
         return `${emDash} ${nrType}`
--- a/auth-web/src/composables/affiliations-factory.ts
+++ b/auth-web/src/composables/affiliations-factory.ts
@@ -157,10 +157,12 @@ export const useAffiliations = () => {
     )
   }
 
-  /** Returns the Name Request type using the NR type code */
+  /** Returns the Name Request type using the NR action code or the NR type code */
   const nameRequestType = (business: Business): string => {
     if (isNameRequest(business)) {
-      const nrType = CommonUtils.mapRequestTypeCdToNrType(business.nameRequest?.requestTypeCd)
+      // Try action code first, and if not found in the enum then use type code
+      let nrType = CommonUtils.mapRequestActionCdToNrType(business.nameRequest?.requestActionCd)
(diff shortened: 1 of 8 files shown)


seeker25:
LGTM

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2521-84na9fbj.web.app

JazzarKarim (approved):
LGTM Leo 👍

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -157,10 +157,12 @@ export const useAffiliations = () => {
...
   }
 
-  /** Returns the Name Request type using the NR type code */
+  /** Returns the Name Request type using the NR action code or the NR type code */
   const nameRequestType = (business: Business): string => {
     if (isNameRequest(business)) {
-      const nrType = CommonUtils.mapRequestTypeCdToNrType(business.nameRequest?.requestTypeCd)
+      // Try action code first, and if not found in the enum then use type code
+      let nrType = CommonUtils.mapRequestActionCdToNrType(business.nameRequest?.requestActionCd)
+      nrType = nrType || CommonUtils.mapRequestTypeCdToNrType(business.nameRequest?.requestTypeCd)
```
severinbeauvais, line 165: The types of requestActionCd and requestTypeCd aren't the same but since they're used for a display string, this is not a problem, though you might want to type `nrType` as string just to emphasize this.

auth-web/src/util/common-util.ts
```diff
@@ -273,6 +273,11 @@ export default class CommonUtils {
     return (url) ? url.trim().replace(/\/+$/, '') : ''
   }
 
+  /** use NR request action code to get NR type from enum */
+  static mapRequestActionCdToNrType (requestActionCd: NrRequestActionCodes): string {
+    return NrRequestActionStrings[requestActionCd] as string
+  }
```
severinbeauvais, line 279: I'd rather see a static string mapping here instead of dereferencing an array with a variable.

This won't fail (it will return undefined) but it's not good practice. Ideally you would check array bounds before dereferencing, or make sure the array contains all possible values (per the enum key).

auth-web/src/util/constants.ts
```diff
@@ -248,6 +248,13 @@ export enum NrEntityType {
     INFO = 'INFO', // special value for sub-menu
 }
 
+export enum NrRequestActionStrings {
+  // Can add other NrRequestActionCodes here to use the action code instead of the NrRequestTypeCd.
+  // Must ensure that the action code does not have several potential types
+  // Example: the NEW action code can be for Incorporation or Registration, so we cannot use it for the NR type
```
severinbeauvais, line 254: I like the caveat but this further indicates that this should be all in common utils, where proper logic can be applied and no mistake can be made by dereferencing the array by "NEW".

seeker25 (approved):

seeker25:
@leodube-aot please address Severin's comments and then feel free to merge.

severinbeauvais:
Should this PR be merged into feature branch (not main)?

auth-web/src/util/common-util.ts
leodube-aot: So using something like `NrRequestActionStrings.get(requestActionCd)` with a map? I should probably also update [mapRequestTypeCdToNrType](https://github.com/bcgov/sbc-auth/blob/main/auth-web/src/util/common-util.ts#L277) since it uses the same logic as this currently.

severinbeauvais: I was thinking that you don't need a map for a single case. I was also thinking that you'd put any caveats here.

leodube-aot: Have a look at my latest change and let me know if that is what you mean.

severinbeauvais: Exactly. It's very clear now, thanks!

severinbeauvais (approved):

leodube-aot:
> Should this PR be merged into feature branch (not main)?

I believe it should be merged into the main branch. The NR type feature was feature flagged and merged into main (see: https://github.com/bcgov/sbc-auth/pull/2389), and this fix will also be covered by the FF.

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2521-84na9fbj.web.app

severinbeauvais:
> > Should this PR be merged into feature branch (not main)?
> 
> I believe it should be merged into the main branch. The NR type feature was feature flagged and merged into main (see: #2389), and this fix will also be covered by the FF.

All right, thanks. I think you have enough reviews... merge when ready!

PR #2531, opened 2023-09-11
Move the state check above applicants. Also put in an error when ther…

seeker25:
…e are no applicants.

*Description of changes:*
Came out of testing Tay's ticket with some bad data

### DIFF
--- a/auth-api/src/auth_api/services/affiliation.py
+++ b/auth-api/src/auth_api/services/affiliation.py
@@ -220,66 +220,64 @@ class Affiliation:
         entity = EntityService.find_by_business_identifier(business_identifier, skip_auth=True)
 
         # Call the legal-api to verify the NR details
-        nr_json = Affiliation._get_nr_details(business_identifier, bearer_token)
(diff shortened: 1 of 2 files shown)


seeker25:
This is what is shown to the client:

image

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -224,12 +224,16 @@ def create_new_business_affiliation(org_id,  # pylint: disable=too-many-argument
 
         if nr_json:
             status = nr_json.get('state')
-            nr_phone = nr_json.get('applicants').get('phoneNumber')
-            nr_email = nr_json.get('applicants').get('emailAddress')
 
             if status not in (NRStatus.APPROVED.value, NRStatus.CONDITIONAL.value, NRStatus.DRAFT.value):
```
seeker25, line 228: This should filter out old consumed name requests that might not have this data

ochiu (approved):

PR #2559, opened 2023-09-19
17596 - Add support for affiliations in multiple environments

kris-daxiom:
*Issue #:*
https://github.com/bcgov/entity/issues/17596

*Description of changes:*
Added environment in the affiliations model and changed the org and modified affiliation endpoints, services to pass the environment to the model.


auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -359,18 +362,18 @@ def post_organization_affiliation(org_id):
     try:
         if is_new_business:
             response, status = AffiliationService.create_new_business_affiliation(
-                org_id, business_identifier, request_json.get('email'),
+                org_id, env, business_identifier, request_json.get('email'),
```
seeker25, line 365: Too many parms, use a dataclass or dict

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -359,18 +362,18 @@ def post_organization_affiliation(org_id):
...
                 request_json.get('phone'), request_json.get('certifiedByName'),
                 bearer_token=bearer_token).as_dict(), http_status.HTTP_201_CREATED
         else:
             response, status = AffiliationService.create_affiliation(
```
seeker25, line 369: Too many parms, use a dataclass or dict

auth-api/src/auth_api/resources/v1/affiliation_invitation.py
```diff
@@ -224,3 +228,11 @@ def _verify_permissions(user, affiliation_invitation_id):
     to_org_id = affiliation_invitation.as_dict()['to_org']['id']
     if not UserService.is_user_member_of_org(user=user, org_id=to_org_id):
         raise BusinessException(Error.NOT_AUTHORIZED_TO_PERFORM_THIS_ACTION, None)
+
+
+def _get_environment():
```
ochiu, line 233: I see this duplicated, perhaps move into utils?

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -170,11 +171,13 @@ def create_affiliation(org_id, business_identifier, pass_code=None, certified_by
...
         # Ensure this affiliation does not already exist
-        affiliation = AffiliationModel.find_affiliation_by_org_and_entity_ids(org_id, entity_id)
+        affiliation = AffiliationModel.find_affiliation_by_org_and_entity_ids(org_id, entity_id, environment)
         if affiliation is not None:
             raise BusinessException(Error.DATA_ALREADY_EXISTS, None)
 
-        affiliation = AffiliationModel(org_id=org_id, entity_id=entity_id, certified_by_name=certified_by_name)
+        affiliation = AffiliationModel(org_id=org_id, entity_id=entity_id, certified_by_name=certified_by_name,
+                                       environment=environment)
+        affiliation.environment = environment
```
ochiu, line 180: Is this needed since it is already being passed in when instantiating the model?

auth-api/tests/unit/services/test_affiliation_invitation.py
```diff
@@ -81,7 +81,7 @@ def test_as_dict(session, auth_mock, keycloak_mock, business_mock, monkeypatch):
             business_identifier=entity_dictionary['business_identifier'])
 
         affiliation_invitation = AffiliationInvitationService.create_affiliation_invitation(affiliation_invitation_info,
-                                                                                            User(user), '')
+                                                                                            User(user), '', 'test')
```
ochiu, line 84: We should probably add additional tests that is more focused on testing the multi-environment nature of this change.

The existing updates are good for testing everything is working as previously, but probably not for the scenario it was implemented for, unless I am missing something?

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -293,7 +361,7 @@ def get_nr_payment_details(business_identifier):
         return invoices
 
     @staticmethod
-    def delete_affiliation(org_id, business_identifier, email_addresses: str = None,
+    def delete_affiliation(org_id, business_identifier, environment: str = None, email_addresses: str = None,   # pylint: disable=too-many-arguments  # noqa: E501
```
seeker25, line 364: too many arguments

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -203,10 +207,16 @@ def is_authorized(entity: Entity, pass_code: str) -> bool:
         return True
 
     @staticmethod
-    def create_new_business_affiliation(org_id,  # pylint: disable=too-many-arguments, too-many-locals
-                                        business_identifier=None, email=None, phone=None, certified_by_name=None,
+    def create_new_business_affiliation(affiliation_data: AffiliationData,  # pylint: disable=too-many-locals, too-many-branches, too-many-statements  # noqa: E501
```
seeker25, line 210: too many branches, means we should refactor into functions

seeker25 (commented):
see above

auth-api/src/auth_api/services/affiliation.py
seeker25: It was added in this PR, there was no too many branches code smell or linting issue before - I'm fairly positive

auth-api/tests/unit/services/test_affiliation.py
```diff
@@ -244,13 +247,14 @@ def test_create_affiliation_firms_party_not_valid(session, auth_mock, monkeypatc
...
     with pytest.raises(BusinessException) as exception:
-        AffiliationService.create_affiliation(org_id, business_identifier, 'test user')
+        AffiliationService.create_affiliation(org_id, business_identifier, 'test', 'test user')
 
     assert exception.value.code == Error.INVALID_USER_CREDENTIALS.name
 
 
 def test_find_affiliated_entities_by_org_id(session, auth_mock):  # pylint:disable=unused-argument
     """Assert that an Affiliation can be created."""
+    env = 'test'
```
seeker25, line 257: Suggestion: enum for environments?

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -293,9 +305,14 @@ def get_nr_payment_details(business_identifier):
         return invoices
 
     @staticmethod
-    def delete_affiliation(org_id, business_identifier, email_addresses: str = None,
-                           reset_passcode: bool = False, log_delete_draft: bool = False):
+    def delete_affiliation(delete_affiliation_request: DeleteAffiliationRequest, environment: str = None):
```
seeker25, line 308: suggestion: `delete_affiliation_request` -> `request`, that way you don't have to remap like on 310

auth-api/migrations/versions/2023_09_05_1ddf3b37e17f_.py
```diff
@@ -0,0 +1,34 @@
...
+# revision identifiers, used by Alembic.
+revision = '1ddf3b37e17f'
+down_revision = '98bee9969323'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('affiliations', sa.Column('environment', sa.String(length=20), nullable=False))
```
seeker25, line 21: Wont this have issues where environment is already empty?

auth-api/src/auth_api/models/affiliation.py
```diff
@@ -51,41 +52,49 @@ def affiliation_details_url(self) -> str:
...
-            .filter(Affiliation.org_id == org_id) \
-            .order_by(Affiliation.created.desc()).all()
+            .filter(Affiliation.org_id == org_id)
+        if environment:
+            query = query.filter(Affiliation.environment == environment)
+        return query.order_by(Affiliation.created.desc()).all()
 
     @classmethod
-    def find_affiliations_by_business_identifier(cls, business_identifier: str):
+    def find_affiliations_by_business_identifier(cls, business_identifier: str, environment: str):
```
seeker25, line 84: Suggestion, it might be easier to just use 

`environment = get_request_environment()` everywhere instead of passing it down

seeker25 (commented):
see above

auth-api/src/auth_api/models/affiliation.py
seeker25: Don't use the request object?  Figure out another way to determine if it's sandbox?

auth-api/migrations/versions/2023_09_05_1ddf3b37e17f_.py
seeker25: Ok, so merging this PR, it's going to explode and fail wont it? nullable should be set to True until you run the update script?

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -358,19 +364,21 @@ def post_organization_affiliation(org_id):
         return {'message': 'Business identifier requires at least 1 digit.'}, http_status.HTTP_400_BAD_REQUEST
     try:
         if is_new_business:
+            affiliation_data = AffiliationData(org_id=org_id, business_identifier=business_identifier,
```
seeker25, line 367: Too many params

seeker25 (changes requested):
I think there's a better way of doing this than lugging environment through all of the parameters. Why not just stand up a sandbox instance of auth-api? Similar to pay? ( I think this is a bit longer term goal Thor was talking about )

Another alternative, make some sort of model level code that automatically sets / queries the environment. 

If we look at the base model we already have something that uses request:

```
    @classmethod
    def create_activity(cls, obj, is_delete=False):
        """Create activity records if the model is versioned."""
        if isinstance(obj, VersionedModel) and not current_app.config.get('DISABLE_ACTIVITY_LOGS'):
            if is_delete:
                verb = 'delete'
            else:
                verb = 'update' if obj.modified_by is not None else 'create'

            activity = activity_plugin.activity_cls(verb=verb, object=obj, data={
                'user_name': g.jwt_oidc_token_info.get('preferred_username',
                                                       None) if g and 'jwt_oidc_token_info' in g else None,
                'remote_addr': fetch_remote_addr()
            })

            db.session.add(activity)
```
This already uses request -^  (`fetch_remote_addr()`)

```


def fetch_remote_addr():
    # Return None if we are outside of request context.
    if _app_ctx_stack.top is None or _request_ctx_stack.top is None:
        return
    return request.remote_addr
```

We could add in one of these to check for the entity you want to add environment for - if it's the right entity it can look at g context and grab the environment from there. 

EG.  For inserting or update the environment:

```
@event.listens_for(Org, 'before_insert')
def receive_before_insert(mapper, connection, target):  # pylint: disable=unused-argument; SQLAlchemy callback signature
    """Rejects invalid type_codes on insert."""
    org = target
    if org.type_code in (OrgTypeEnum.SBC_STAFF.value, OrgTypeEnum.STAFF.value):
        raise BusinessException(
            Error.INVALID_INPUT,
            None
        )


@event.listens_for(Org, 'before_update', raw=True)
def receive_before_update(mapper, connection, state):  # pylint: disable=unused-argument; SQLAlchemy callback signature
    """Rejects invalid type_codes on update."""
    if Org.type_code.key in state.unmodified:
        return
    raise BusinessException(
        Error.INVALID_INPUT,
        None
    )

```
You could easily set the environment here -^ for the entities you need. If its short and doesn't touch the resources / services / models - it's very easy to rip it out again and fix it later. 


For querying:

I think even just a helper method would suffice:

EG.
    @classmethod
    def query_active_users(cls):
        return cls.query.filter_by(is_deleted=False)

Probably could override the BaseQuery - check if it's the entity you need, if it is, filter out for environment - so it's done in one spot. 


EG.

````
    self.Affiliation.query_class = query_class

class Affiliation(db.Model):
    query_class = EnvironmentQuery

class EnvironmentQuery(Query):
    def all_include_environment(self):
          return self.filter(Affiliation.environment == <env>).all()
    def first_include_environment(self):
          return self.filter(Affiliation.environment == <env>).first()
````
etc


Looking at flask sqlalchemy locally:

image

It's using request for pagination.

seeker25:
I'm not comfortable with lugging around the environment variable from the resource down to the model. @ochiu agrees with me as well.

seeker25:
Sometimes it takes time for thoughts to percolate. @thorwolpert @kialj876 What are your thoughts on this PR?

seeker25:
@kris-daxiom could we set nullable = True in the migration, plus fix the merge conflict - then merge?

seeker25 (approved):

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2559-x9lrl7vq.web.app

PR #2564, opened 2023-09-20
17776 - Soft delete support for accepted affiliation invitations

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/17776

*Description of changes:*
- add is_deleted flag to affiliation_invitations table (default to false)
- update DELETE to soft delete on accepted affiliation invitations
- update searches and GET to not return soft deleted affiliation invitations 

### DIFF
--- a/auth-api/src/auth_api/models/affiliation_invitation.py
+++ b/auth-api/src/auth_api/models/affiliation_invitation.py
@@ -15,7 +15,7 @@
 
 from datetime import datetime, timedelta
 
-from sqlalchemy import Column, DateTime, ForeignKey, Integer, String, or_
+from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Integer, String, or_
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import relationship
 
@@ -50,6 +50,7 @@ class AffiliationInvitation(BaseModel):  # pylint: disable=too-many-instance-att
     invitation_status_code = Column(ForeignKey('invitation_statuses.code'), nullable=False, default='PENDING')
     type = Column(ForeignKey('affiliation_invitation_types.code'), nullable=False, default='EMAIL')
     additional_message = Column(String(4000), nullable=True)
+    is_deleted = Column(Boolean(), default=False)
 
     invitation_status = relationship('InvitationStatus', foreign_keys=[invitation_status_code])
     sender = relationship('User', foreign_keys=[sender_id])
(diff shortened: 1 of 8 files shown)


auth-api/src/auth_api/resources/v1/affiliation_invitation.py
```diff
@@ -105,8 +105,9 @@ def post_affiliation_invitation():
 @_jwt.requires_auth
 def get_affiliation_invitation(affiliation_invitation_id):
     """Get the affiliation invitation specified by the provided id."""
-    if not (affiliation_invitation := AffiliationInvitationService.
-            find_affiliation_invitation_by_id(affiliation_invitation_id)):
+    affiliation_invitation = AffiliationInvitationService.find_affiliation_invitation_by_id(affiliation_invitation_id)
+
+    if not affiliation_invitation or affiliation_invitation.as_dict().get('is_deleted'):
```
seeker25, line 110: sucks having to do this :(

auth-api/src/auth_api/models/affiliation_invitation.py
```diff
@@ -50,6 +50,7 @@ class AffiliationInvitation(BaseModel):  # pylint: disable=too-many-instance-att
     invitation_status_code = Column(ForeignKey('invitation_statuses.code'), nullable=False, default='PENDING')
     type = Column(ForeignKey('affiliation_invitation_types.code'), nullable=False, default='EMAIL')
     additional_message = Column(String(4000), nullable=True)
+    is_deleted = Column(Boolean(), default=False)
```
seeker25, line 53: might make more sense to put this as `is_viewed`?  `is_deleted` is fine though

seeker25 (approved):

auth-api/src/auth_api/resources/v1/affiliation_invitation.py
hfekete: I think it could be avoided by moving the check if something is deleted on the model class.
I didnt test following snippet but I think it could be something along these lines:
```   @classmethod
    def find_invitation_by_id(cls, invitation_id, is_deleted=False):
        """Find an affiliation invitation record that matches the id."""
        return cls.query.filter_by(is_deleted=is_deleted).filter_by(id=invitation_id).first()```
        
It should maintain same behavior as it has now.

hfekete (approved):
Added one comment, but not necessary to act on it.

auth-api/src/auth_api/resources/v1/affiliation_invitation.py
ochiu: @hfekete thanks for the feedback, that is a good thought. I will keep it in mind for future similar situations.

In this case I would prefer a find by id method to return the object regardless of its state. I think if we start leveraging the is_deleted flag more for affiliation invitations I would add a new method in the model to filter as per your suggestion. Thanks!

auth-api/src/auth_api/models/affiliation_invitation.py
ochiu: Will keep is as `is_deleted` then, thanks

PR #2565, opened 2023-09-21
Feature flag RTR's row and dialog option (Requested by OCM). Fix staf…

seeker25:
*Description of changes:*
Feature flag RTR's row and dialog option (Requested by OCM). Fix staff detection so it works regardless of current org. (SBC STAFF only works for current org).

### DIFF
--- a/auth-web/src/stores/business.ts
+++ b/auth-web/src/stores/business.ts
@@ -148,11 +148,11 @@ export const useBusinessStore = defineStore('business', () => {
     }
 
     const pendingAffiliationInvitations = await AffiliationInvitationService.getAffiliationInvitations(currentOrganization.value.id) || []
+    const includeAffiliationInviteRequest = LaunchDarklyService.getFlag(LDFlags.EnableAffiliationDelegation) || false
 
     for (const affiliationInvite of pendingAffiliationInvitations) {
       // Skip over affiliation requests for type REQUEST for now.
-      if (affiliationInvite.type === AffiliationInvitationType.REQUEST &&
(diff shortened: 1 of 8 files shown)


ochiu (approved):

auth-web/src/util/constants.ts
```diff
@@ -435,7 +435,8 @@ export enum LDFlags {
     SentryEnable = 'sentry-enable',
     SupportRestorationEntities = 'supported-restoration-entities',
     AllowableBusinessSearchTypes = 'allowable-business-search-types',
-    AllowableBusinessPasscodeTypes = 'allowable-business-passcode-types'
+    AllowableBusinessPasscodeTypes = 'allowable-business-passcode-types',
+    EnableAffiliationDelegation = 'enable-affiliation-delegation'
```
seeker25, line 439: Not sorting these alphabetically, I'd rather leave them so the more recent ones are at the bottom. That way when we do cleanup - we know which flags are the oldest.

auth-web/src/stores/business.ts
```diff
@@ -147,8 +148,13 @@ export const useBusinessStore = defineStore('business', () => {
     }
 
     const pendingAffiliationInvitations = await AffiliationInvitationService.getAffiliationInvitations(currentOrganization.value.id) || []
+    const includeAffiliationInviteRequest = LaunchDarklyService.getFlag(LDFlags.EnableAffiliationDelegation) || false
```
seeker25, line 151: Default is false when launch darkly fails.

seeker25:
image

PR #2567, opened 2023-09-22
17879 - Fix isBusinessAffiliated check and missing corpNum

seeker25:
https://github.com/bcgov/entity/issues/17879

### DIFF
--- a/auth-web/src/composables/affiliations-factory.ts
+++ b/auth-web/src/composables/affiliations-factory.ts
@@ -43,7 +43,7 @@ export const useAffiliations = () => {
   }
 
   const isBusinessAffiliated = (businessIdentifier: string): boolean => {
-    if (businessIdentifier === undefined) {
+    if (!businessIdentifier) {
       return false
     }
     return affiliations.results.some(business => businessIdentifier === business.businessIdentifier)
--- a/auth-web/src/composables/affiliations-factory.ts
+++ b/auth-web/src/composables/affiliations-factory.ts
@@ -43,6 +43,9 @@ export const useAffiliations = () => {
   }
 
   const isBusinessAffiliated = (businessIdentifier: string): boolean => {
+    if (businessIdentifier === undefined) {
+      return false
+    }
     return affiliations.results.some(business => businessIdentifier === business.businessIdentifier)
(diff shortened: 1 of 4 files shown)


auth-web/src/composables/affiliations-factory.ts
```diff
@@ -43,6 +43,9 @@ export const useAffiliations = () => {
   }
 
   const isBusinessAffiliated = (businessIdentifier: string): boolean => {
+    if (businessIdentifier === undefined) {
+      return false
+    }
```
severinbeauvais, line 48: I would do `if (!businessIdentifier)` to handle the null case, too. (I think we never have id=0.)

Walk me through this. This functions returns false, meaning the business is not affiliated, meaning...? How did the user get here and what happens now?

seeker25: I think it has to do with another change that was missing - the corpNum wasn't being brought up

auth-web/src/stores/business.ts
```diff
@@ -123,7 +123,8 @@ export const useBusinessStore = defineStore('business', () => {
       requestActionCd: nr.requestActionCd,
       natureOfBusiness: nr.natureBusinessInfo,
       expirationDate: nr.expirationDate,
-      applicants: nr.applicants
+      applicants: nr.applicants,
+      corpNum: nr.corpNum
```
severinbeauvais, line 127: I think there are a few interfaces that should be updated. I think Hongjing did something similar about a month ago with a new property from Namex API.

seeker25: Let me look at her changes and make sure I didn't miss them refactoring part of this

auth-web/src/composables/affiliations-factory.ts
seeker25: I think it has to do with another change that was missing - the corpNum wasn't being brought up

auth-web/src/stores/business.ts
seeker25: corpNum was missing, that's it I think

severinbeauvais (approved):

PR #2569, opened 2023-09-25
17519 - Indicate to user that cancelled/expired NRs will not be returned by NR search

chenhongjing:
*Issue #:*
https://github.com/bcgov/entity/issues/17519

*Description of changes:*
on My Business Registry Dashboard
- Included the word 'active' in initial text
- Changed no matches state text for NR search
- app version = 2.4.26

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the sbc-auth license (Apache 2.0).

PR #2573, opened 2023-09-27
Update sbc-common-components.

seeker25:

This should include the modified getFlag, which accepts a defaultValue.

### DIFF
--- a/auth-web/package-lock.json
+++ b/auth-web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "auth-web",
-  "version": "2.4.26",
+  "version": "2.4.27",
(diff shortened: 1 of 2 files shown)


auth-web/package-lock.json
```diff
@@ -8131,6 +8131,7 @@
       "version": "1.4.0",
       "inBundle": true,
       "license": "MIT",
+      "optional": true,
```
seeker25, line 8134: Hmm, I'm using node 16.15.0

ochiu: I think https://github.com/bcgov/sbc-auth/pull/2569 changed it previously.

PR #2642, opened 2023-11-20
18522, 18525 - updates to REQUEST affiliation

hfekete:
*Issue #:18522* https://github.com/bcgov/entity/issues/18522
*Issue #:18525* https://github.com/bcgov/entity/issues/18525

*Description of changes:*
Updates to REQUEST affiliation. 
- filtered out SBC_STAFF and STAFF accounts from the list.
- added branch name in list display.


auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -400,7 +400,12 @@ def get_org_details_by_affiliation(business_identifier):
     try:
         data = OrgService.search_orgs_by_affiliation(business_identifier, pagination_info, environment)
 
-        org_details = [{'name': org.name, 'uuid': org.uuid} for org in data['orgs']]
+        org_details = \
+            [
+                {'name': org.name, 'uuid': org.uuid, 'branchName': org.branch_name} for org
+                in data['orgs']
+                if org.type_code.upper() not in ['SBC_STAFF', 'STAFF']
```
ochiu, line 407: I believe there is an org type enum you can use.

ochiu: Does this change affect the pagination since you are potentially only including a subset of orgs after querying?

seeker25: I'd tuck this into the service, I think it's only your code using `search_orgs_by_affiliation`

hfekete: Updated so it takes `excluded_org_types` as param.

ochiu (approved):

auth-api/src/auth_api/models/org.py
```diff
@@ -158,12 +159,19 @@ def search_org(cls, search: OrgSearch, environment: str):
         return pagination.items, pagination.total
 
     @classmethod
-    def search_orgs_by_business_identifier(cls, business_identifier, pagination_info: PaginationInfo, environment):
+    def search_orgs_by_business_identifier(cls,
```
seeker25, line 162: Alotta params, would have combined pagination_info, excluded_org_types together

seeker25 (approved):

PR #2661, opened 2023-12-12
18940 Add new draft type for amalgamation

kzdev420:
*Issue #:*
https://github.com/bcgov/entity/issues/18940

*Description of changes:*

### DIFF
--- a/auth-api/src/auth_api/services/affiliation.py
+++ b/auth-api/src/auth_api/services/affiliation.py
@@ -36,7 +36,7 @@ from auth_api.schemas import AffiliationSchema
 from auth_api.services.entity import Entity as EntityService
 from auth_api.services.org import Org as OrgService
 from auth_api.services.user import User as UserService
-from auth_api.utils.enums import ActivityAction, CorpType, NRNameStatus, NRStatus
+from auth_api.utils.enums import ActivityAction, CorpType, NRActionCodes, NRNameStatus, NRStatus
 from auth_api.utils.passcode import validate_passcode
 from auth_api.utils.roles import ALL_ALLOWED_ROLES, CLIENT_AUTH_ROLES, STAFF
 from auth_api.utils.user_context import UserContext, user_context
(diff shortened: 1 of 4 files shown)


auth-api/src/auth_api/services/affiliation.py
```diff
@@ -457,6 +457,10 @@ def _combine_affiliation_details(details):
             if 'nrNumber' in business and (nr_num := business['nrNumber']):
                 if business['nrNumber'] in name_requests:
                     business['nameRequest'] = name_requests[nr_num]['nameRequest']
+                    # Update the draft type if the draft NR request is for amalgamation
+                    if business.get('draftType', None) \
```
seeker25, line 461: Should we throw in a unit test for this? Not sure if any exist

seeker25 (approved):

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -457,6 +457,10 @@ def _combine_affiliation_details(details):
...
+                            and business['nameRequest']['request_action_cd'] == NRActionCodes.AMALGAMATE.value:
+                        business['draftType'] = CorpType.ATMP.value
```
severinbeauvais, line 463: I expected to see this code alongside other code that sets TMP or RTMP ...?

auth-api/src/auth_api/utils/enums.py
```diff
@@ -353,3 +354,21 @@ class NotificationTypes(Enum):
...
+    CONVERSION = 'CNV'  # aka Alteration
+    DBA = 'DBA'  # doing business as
+    MOVE = 'MVE'  # continuation in
+    NEW_BUSINESS = 'NEW'  # incorporate or register
+    RESTORE = 'REH'  # restore or reinstate
+    RENEW = 'REN'  # restore with new name request
+    RESTORATION = 'REST'  # FUTURE: unused? delete?
+    RESUBMIT = 'RESUBMIT'  # FUTURE: unused? delete?
+
+    INFO = 'INFO'  # special value for sub-menu
```
severinbeauvais, line 374: You don't need INFO in the API; it's only for Namerequest UI.

kzdev420: Updated

auth-api/src/auth_api/services/affiliation.py
kzdev420: It is here line 450. `sets TMP or RTMP` but at the moment, we don't have the NameRequests.
So, I updated the draftType when combine the NR

severinbeauvais: I see. I still think that AAs are fundamentally the same as IAs and Regs (even including NRs) but if you and Travis think this is OK then I'm OK with this, too.

PR #2689, opened 2024-01-22
Update jinja to version 3.1.3

seeker25:


seeker25:
Hmm looks like we need to upgrade flask to make this happen

PR #2780, opened 2024-05-02
20477- Statement Number added to Statement Page

rodrigo-barraza:
*Issue #:*
https://github.com/bcgov/entity/issues/20477

*Description of changes:*


auth-web/src/components/auth/account-settings/statement/Statements.vue
```diff
@@ -99,6 +99,14 @@
             <span>{{ paymentMethodsDisplay(item.paymentMethods) }}</span>
           </div>
         </template>
+        <template
+          v-if="enableEFTPaymentMethod && hasEFTPaymentMethod"
```
ochiu, line 103: I think this change is supposed to be applied to statements in general, not just EFT. 
Maybe confirm this.

ochiu (approved):

auth-web/src/components/auth/account-settings/statement/Statements.vue
seeker25: yup, should be in general

seeker25 (approved):

PR #2818, opened 2024-05-11
21107 - Add T&C to Selecting EFT Payment

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/21107

*Description of changes:*

### DIFF
--- a/auth-web/src/components/auth/common/TermsOfUseDialog.vue
+++ b/auth-web/src/components/auth/common/TermsOfUseDialog.vue
@@ -94,7 +94,6 @@
 <script lang="ts" setup>
 import { computed, defineComponent, onMounted, ref } from '@vue/composition-api'
 import TermsOfUse from '@/components/auth/common/TermsOfUse.vue'
-import { mapState } from 'pinia'
 import { useUserStore } from '@/stores/user'
 
 export default defineComponent({
--- a/auth-web/src/components/auth/common/TermsOfUseDialog.vue
+++ b/auth-web/src/components/auth/common/TermsOfUseDialog.vue
@@ -27,7 +27,7 @@
             tabindex="0"
             @keyup.enter="openDialog()"
             @click.stop="openDialog()"
-          >terms and conditions</strong>
+          >{{ tosText }}</strong>
           {{ tosCheckBoxLabelAppend }}
         </span>
       </template>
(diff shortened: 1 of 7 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2818-gmoammzq.web.app

auth-web/src/components/auth/common/TermsOfUseDialog.vue
```diff
@@ -111,6 +111,7 @@ export default class TermsOfUseDialog extends Vue {
   @Prop({ default: 'termsofuse' }) tosType: string
```
seeker25, line 111: Looks good Jia, is it possible to refactor this into composition-api?

seeker25 (approved):

auth-web/src/components/auth/common/TermsOfUseDialog.vue
seeker25: Here's a prompt that might help you out - haven't tried but we should figure something out that can do 85% of the work:

```
You are code translator, that takes a vuejs 2 component written with vue-class-component and vue-property-decorator and converts it to a vuejs 3 using composition API. These are your guidelines:
1. Replace class-based component with the Composition API.
2. Use the `<script lang="ts" setup>` syntax for the script section.
3. Define props using `defineProps`, maintaining and specifying their types with type annotations, eg.: `const props = defineProps<{ propName: Type; }>();`
4. Convert methods to regular anonymous functions inside constants, eg.: `const coolFunction = () => {};`.
5. Replace `this.$t` and `i18n.t` function calls with the `t` function, which you can import with `const { t } = useI18n();`.
6. Preserve translation keys in the format of 'NAMESPACE.KEY'.
7. Use 4 spaces for indentation.
8. Avoid destructuring props; use `props.propName`.
9. Convert computed properties to use the `computed` function.
10. Do not import anything, even `ref` and `computed` from vue, as everything is globally imported.
11. Do not return the template part under any circumstances, even as an example, I only care about the script.
12. Use TypeScript and type annotations wherever possible.
```

seeker25 (approved):

auth-web/src/components/auth/common/TermsOfUseDialog.vue
Jxio: yes, I can do it today

seeker25:
also.. I should ask... this doesn't have any impact on an existing flow without EFT?

Jxio:
> also.. I should ask... this doesn't have any impact on an existing flow without EFT? EG. PAD etc.
it only used by PAD and EFT for now

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2818-gmoammzq.web.app

PR #2819, opened 2024-05-13
20929 - Implemented GET involuntary dissolution configurations endpoint

JazzarKarim:
*Issue #:*
https://github.com/bcgov/entity/issues/20929

### This PR implements the GET call for now. Next PR will implement the PUT call (the BE is still WIP). I'm doing this now since we're demoing tomorrow and I would like to show this.

*Description of changes:*
- Interface Changes for the dissolution
- Implemented the GET call of the configurations
- Misc. fixes

### DIFF
--- a/auth-web/src/components/auth/staff/DissolutionSchedule.vue
+++ b/auth-web/src/components/auth/staff/DissolutionSchedule.vue
@@ -16,7 +16,7 @@
           >
             <v-text-field
               ref="numberOfBusinessesRef"
-              v-model="numberOfBusinesses"
+              v-model="numberOfBusinessesEdit"
               filled
               type="number"
               label="Dissolution Batch Size"
@@ -79,7 +79,7 @@
               large
               outlined
               class="mr-3"
-              @click="cancelBtnClicked()"
+              @click="triggerEditOnOff()"
             >
               Cancel
             </v-btn>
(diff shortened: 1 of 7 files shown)


auth-web/src/components/auth/staff/DissolutionSchedule.vue
```diff
@@ -103,62 +103,45 @@ import { useStaffStore } from '@/stores/staff'
...
 export default defineComponent({
   name: 'DissolutionSchedule',
-  emits: ['update:onHold'],
-  setup (_, { emit }) {
+  setup () {
     const state = reactive({
       menu: false,
-      numberOfBusinesses: 0,
+      numberOfBusinessesEdit: 0,
+      numberOfBusinessesNonEdit: 0,
```
JazzarKarim, line 110: I know, this looks weird (having two properties that look similar). I have an explanation though 😆 

The idea is, when the edit panel is open, the schedule summary is going to be reactive to what's being typed in the text field. However, when it is not open, it's going to be what the DB is returning of the current job batch size.

So, I decided to separate those fields. One of the biggest reasons is that, in doing so, I can store the current batch size from the GET call into the nonEdit field. This way, I only need to use the GET call once on mount. Otherwise, I have to do the GET call each time cancel or save is clicked which is really not good haha

Also, this provides so much flexibility and it separates of when the edit panel is open or not.

auth-web/src/components/auth/staff/DissolutionSchedule.vue
```diff
@@ -103,62 +103,45 @@ import { useStaffStore } from '@/stores/staff'
...
-
     /**
      * Save button is clicked. Update the dissolution batch size.
      * Only save if the inputted number is valid.
      * TODO: Implement logic (job) once the BE is done.
      */
     const saveBtnClicked = (): void => {
       if (state.numberOfBusinessesRef.validate()) {
-        staffStore.updateDissolutionBatchSize(state.numberOfBusinesses)
+        // staffStore.updateDissolutionBatchSize(state.numberOfBusinesses)
```
JazzarKarim, line 143: FUTURE: When I work on the PUT.

auth-web/src/views/auth/staff/InvoluntaryDissolution.vue
```diff
@@ -45,11 +45,8 @@
               badgeText="Paused"
               icon="mdi-calendar-clock"
               label="Automated Dissolution Schedule"
-              :showBadge="isOnHold"
```
JazzarKarim, line 48: Removed all badge logic here as designs have changed. No more badge.

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2819-b8ecll37.web.app

seeker25 (approved):

PR #2848, opened 2024-05-31
20159 - Add view_address permission to admin account

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/20159

*Description of changes:*

### DIFF
--- a/auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
+++ b/auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
@@ -23,4 +23,4 @@ def upgrade():
 
 
 def downgrade():
-    op.execute('DELETE FROM permissions WHERE id=33')
+    op.execute('DELETE FROM permissions WHERE id=84')
--- a/auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
+++ b/auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
@@ -1,0 +1,26 @@
+"""add view address permission to admin
+
+Revision ID: 39b4f331c003
+Revises: 41fa6588c76e
+Create Date: 2024-05-30 20:21:15.422128
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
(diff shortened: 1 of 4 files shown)


seeker25:
Should we update package.json and package-lock?

auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def upgrade():
+    op.execute(
+        "INSERT INTO permissions (id, membership_type_code, actions) VALUES (84, 'ADMIN', 'view_address')"
+    )
+
+
+def downgrade():
+    op.execute('DELETE FROM permissions WHERE id=84')
```
seeker25, line 26: This isn't right

seeker25 (changes requested):
migration needs to be fixed

auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = '39b4f331c003'
+down_revision = '41fa6588c76e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        "INSERT INTO permissions (id, membership_type_code, actions) VALUES (84, 'ADMIN', 'view_address')"
```
seeker25, line 21: I wouldn't be including ids, this should be automatic

auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
seeker25: `DELETE from permissions where actions = 'view_address'`

auth-api/migrations/versions/2024_05_30_39b4f331c003_admin_view_address_permission.py
Jxio: yup updated

seeker25 (approved):

PR #2855, opened 2024-06-05
21134 - Updated business lookup

severinbeauvais:
*Issue #:* bcgov/entity#21134

*Description of changes:*
- app version = 2.6.25
- added safe fallback in case FF is not returned / null in Business Lookup component
- removed legalType fallback in Business Lookup Service
- updated unit test

### DIFF
--- a/auth-web/src/components/auth/manage-business/BusinessLookup.vue
+++ b/auth-web/src/components/auth/manage-business/BusinessLookup.vue
@@ -173,8 +173,9 @@ export default defineComponent({
 
       states.state = States.SEARCHING
       const searchStatus = null // search all (ACTIVE + HISTORICAL)
-      const legalType = launchdarklyServices.getFlag(LDFlags.AllowableBusinessSearchTypes) ||
-        'BEN,SP,GP,CP' // safe fallback to same value as Prod
+      // fetch list of legal types from LaunchDarkly
+      // if not available, fall back to same value as Prod
+      const legalType = launchdarklyServices.getFlag(LDFlags.AllowableBusinessSearchTypes, 'BEN,SP,GP,CP')
 
       // Use appropriate service based on lookupType
       const searchService = (props.lookupType === LookupType.NR)
--- a/auth-web/src/components/auth/manage-business/BusinessLookup.vue
+++ b/auth-web/src/components/auth/manage-business/BusinessLookup.vue
@@ -170,9 +170,12 @@ export default defineComponent({
       if (!states.searchField || states.searchField?.length <= 2) {
         return
       }
+
       states.state = States.SEARCHING
       const searchStatus = null // search all (ACTIVE + HISTORICAL)
-      const legalType = launchdarklyServices.getFlag(LDFlags.AllowableBusinessSearchTypes)
+      const legalType = launchdarklyServices.getFlag(LDFlags.AllowableBusinessSearchTypes) ||
+        'BEN,SP,GP,CP' // safe fallback to same value as Prod
+
       // Use appropriate service based on lookupType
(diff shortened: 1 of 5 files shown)


seeker25 (approved):

severinbeauvais:
Also, as part of the subject ticket, I updated the respective feature flag (sbc-auth:allowable-business-search-types).

Dev and Test (now includes CBEN and CCC):
image

Prod:
image

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2855-wg4bau8c.web.app

auth-web/src/components/auth/manage-business/BusinessLookup.vue
```diff
@@ -170,9 +170,12 @@ export default defineComponent({
       if (!states.searchField || states.searchField?.length <= 2) {
         return
       }
+
       states.state = States.SEARCHING
       const searchStatus = null // search all (ACTIVE + HISTORICAL)
-      const legalType = launchdarklyServices.getFlag(LDFlags.AllowableBusinessSearchTypes)
```
severinbeauvais, line 175: Previously, if LaunchDarkly was not available (or wrong key), this would return `null`.

Then, the Business Lookup Service would fallback to the complete list of entity types. This is not a good fallback.

The new fallback is "same as in Prod".

auth-web/tests/unit/services/business-lookup.service.spec.ts
```diff
@@ -14,15 +14,14 @@ describe('Business Lookup Services', () => {
...
-      '&categories=legalType:A,BC,BEN,C,CC,CP,CUL,FI,GP,LL,LLC,LP,PA,S,SP,ULC,XCP,XL,XP,XS' +
-      `&query=value:${encodeURIComponent('FM1000002')}`
+      `&categories=legalType:SP&query=value:${encodeURIComponent('FM1000002')}`
 
     sinon.stub(axios, 'get').withArgs(url).returns(
       Promise.resolve({ data: { searchResults: { results: [result] } } })
     )
 
     // search and look at results
-    const results = await BusinessLookupServices.search('FM1000002')
```
severinbeauvais, line 25: This was a syntax error because the second parameter was missing (and the third is optional).

auth-web/tests/unit/services/business-lookup.service.spec.ts
```diff
@@ -14,15 +14,14 @@ describe('Business Lookup Services', () => {
     }
 
     const url = BusinessLookupServices.registriesSearchApiUrl + 'businesses/search/facets?start=0&rows=20' +
-      '&categories=legalType:A,BC,BEN,C,CC,CP,CUL,FI,GP,LL,LLC,LP,PA,S,SP,ULC,XCP,XL,XP,XS' +
-      `&query=value:${encodeURIComponent('FM1000002')}`
+      `&categories=legalType:SP&query=value:${encodeURIComponent('FM1000002')}`
```
severinbeauvais, line 17: Instead of updating this list, which would serve no purpose here, I simplified this.

auth-web/src/components/auth/manage-business/BusinessLookup.vue
seeker25: I think you could also use launchdarklyServices.getFlag(LDFlags.AllowableBusinessSearchTypes, 'SP,GP...')

seeker25: I've changed that function recently so the fall backs should work, I think I used ublock origins to test it

severinbeauvais: Good call. I've updated and tested this.

PR #2928, opened 2024-07-22
21983 - Enhance Staff Review Table

ArwenQin:
*Issue #:*21983
https://github.com/bcgov/entity/issues/21983

*Description of changes:*
- Enabled filtering and sorting for the staff continuation review table
- Filter and sort by all fields
- Pass the params to the API (API doesn't support it for now)
- "Clear Filters" button
- Automatically save filters to user's local storage (and resume)


ArwenQin:
image

ArwenQin:
Question: the Figma design used "Accepted", but the data fetched from API uses "Approved"
image
Which one should we use?

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div id="continuation-application-review-table">
-    <v-form class="continuation-application-search continuation-application-search">
+    <v-form class="fas-search continuation-review-search">
```
ArwenQin, line 3: Re: "fas-seach", the name doesn't make sense. It's imported from '~/fas-ui/src/assets/scss/search.scss';
But if I extend it or rename it, the style is lost. I keep it here so the search column style is correct.

ArwenQin:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=bcgov_sbc-auth&pullRequest=2928) **Quality Gate failed**
> Failed conditions  [1 Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=bcgov_sbc-auth&pullRequest=2928&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=bcgov_sbc-auth&pullRequest=2928)

I see, this is due to the method to display status:
in database it shows like "AWAITING_REVIEW", we want to show "Awaiting Review"
image

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
severinbeauvais: This is a dependency on an external project that we really shouldn't have, but I realize the staff account table components use this, too.

I suggest we wait until we port this Entity component to a different project to resolve this CSS issue.

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
```diff
@@ -97,60 +158,219 @@
...
-        { text: 'Date Submitted', value: 'date' },
+        { text: 'Date Submitted', value: 'submissionDate' },
         { text: 'NR Number', value: 'nrNumber' },
-        { text: 'Identifying Number', value: 'businessIdentifier' },
+        { text: 'Identifying Number', value: 'identifier' },
         { text: 'Completing Party', value: 'completingParty' },
         { text: 'Status', value: 'status' },
         { text: 'Actions', value: 'action' }
       ],
+      statusTypes: ref<{ text: string; value: string }[]>([
```
severinbeauvais, line 209: Why do you need `ref<{...` here? Isn't this a readonly object just like headers, above?

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
```diff
@@ -97,60 +158,219 @@
...
+                params.completingParty.length > 0 ||
+                params.status.length > 0
+    }
+
+    // Method to display the status
+    function displayStatus (status: string): string {
+      return status
+        .toLowerCase()
+        .replace(/_/g, ' ')
+        .replace(/(^\w{1})|(\s+\w{1})/g, letter => letter.toUpperCase())
```
severinbeauvais, line 335: This is giving a [security warning](https://sonarcloud.io/project/security_hotspots?id=bcgov_sbc-auth&pullRequest=2928&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true).

Maybe it would be easier to map "status" using a small lookup object?

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
ArwenQin: Will update it

auth-web/src/services/business.services.ts
```diff
@@ -202,41 +202,18 @@ export default class BusinessService {
...
-          status: 'Awaiting Review'
-        }
-      ]
-    } as any))
+  static async fetchContinuationReviews (): Promise<ContinuationReviewIF[]> {
+    const url = `${ConfigHelper.getLegalAPIV2Url()}/admin/reviews`
+    const response = await axios.get(url)
+    const data = response.data.reviews
+    return data
+  }
```
severinbeauvais, line 210: This will throw an error if response or response.data are falsy (null/undefined).

I recommend structuring this as on line 129 above.

Also, what happens if line 207 throws an error (eg, 403 or 404 or 5xx)? Should you return null from this method, or perhaps `[]`? Does the table support this?

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
ArwenQin: Thanks, will do. A quick question: the Figma design used "Accepted", but the data fetched from API and in enum ReviewStatus use "Approved". Which should we use?
image
Figma:
image

auth-web/src/services/business.services.ts
severinbeauvais: This function is obsolete now, right? If so then I recommend deleting it.

(And I see you handle my comments above in the other function: 👍 )

auth-web/src/stores/staff.ts
```diff
@@ -248,6 +250,19 @@ export const useStaffStore = defineStore('staff', () => {
     return {}
   }
 
+  async function searchReviews (filterParams: ReviewFilterParams) {
+    const response = await BusinessService.searchReviews(filterParams)
```
severinbeauvais, line 254: If the Axios function experiences an error (eg, 403, 404, 5xx, etc) then this call will throw an exception.

I recommend adding a chained catch statement that returns, say, null, eg
`.catch(() => null) // return null on error`

severinbeauvais (commented):
I left a few comments. Overall looks good. Is this all the code changes (other than maybe some extra unit tests)?

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
severinbeauvais: I've gone with "approved" on the review page. Can you double-check with Andy?

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2928-f8q1wi1m.web.app

severinbeauvais:
The UI designs show checkboxes in this menu so that multiple statuses can be selected:

image

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
ArwenQin: Updated:
image

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
ArwenQin: Confirmed with Andy, use "Approved"

auth-web/src/stores/staff.ts
ArwenQin: Updated

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
ArwenQin: Updated to display status using a map

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
```diff
@@ -97,60 +163,222 @@
...
+    const staffStore = useStaffStore()
+    const statusDisplayMap = {
+      AWAITING_REVIEW: 'Awaiting Review',
+      CHANGE_REQUESTED: 'Change Requested',
+      RESUBMITTED: 'Resubmitted',
+      REJECTED: 'Rejected',
+      APPROVED: 'Approved',
+      ABANDONED: 'Abandoned'
+    }
+
```
ArwenQin, line 211: The Quality Gate failed warning should not be here anymore.

ArwenQin:
> The UI designs show checkboxes in this menu so that multiple statuses can be selected:
> 
> image

Enabled multiple selection for status now:
image

When more than one status are selected, the display is not good:
image
Will work on this further, any suggestions?

auth-web/src/stores/staff.ts
```diff
@@ -251,16 +251,21 @@ export const useStaffStore = defineStore('staff', () => {
...
-        reviews: response.data.reviews
+    try {
+      const response = await BusinessService.searchReviews(filterParams)
+      if (response?.data) {
+        return {
+          limit: response.data.limit,
+          page: response.data.page,
+          total: response.data.total,
+          reviews: response.data.reviews
+        }
```
severinbeauvais, line 262: Why return a new object here? Can't you just return `response.data`?

severinbeauvais:
> Will work on this further, any suggestions?

I didn't look at this deeply but it looks like the list isn't correctly anchored to the dropdown menu...

severinbeauvais:
Please update the app version :)

severinbeauvais:
> > Will work on this further, any suggestions?
> 
> I didn't look at this deeply but it looks like the list isn't correctly anchored to the dropdown menu...

https://v2.vuetifyjs.com/en/components/selects/#multiple

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
severinbeauvais: I don't know why SonarCloud didn't automatically pick up the fixed code and I don't know how to re-rerun the analysis, so I went in to SonarCloud and told it the code was fixed.

auth-web/src/stores/staff.ts
severinbeauvais: In fact, why not move this block into BusinessService.searchReviews(), so that it returns an object of type ReviewList (instead of the Axios response)?

I think it's better to keep the Axios stuff contained in the services file. THIS file shouldn't know about an Axios response.

auth-web/src/stores/staff.ts
severinbeauvais: Your update is OK but next time you can keep the code simpler by catching the exception as per my example.

auth-web/src/components/auth/staff/continuation-application/ContinuationApplicationTable.vue
severinbeauvais: Yes, I fixed it. See previous comment.

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2928-f8q1wi1m.web.app

auth-web/src/stores/staff.ts
ArwenQin: I used the similar structure as line 240 function searchOrgs() in this file. If we move this block into BusinessService.searchReviews(), it looks like we don't need the function in the staff.ts file anymore? We can call BusinessService.searchReviews() directly in the vue.

severinbeauvais: Maybe...

Normally a services method should throw an exception if there's a network error (403, 404, 5xx, etc) and the caller should handle that. However, I'm OK if you decide to handle the error in the services method and return null or `[]` accordingly, so that you can use the service method directly in the Vue template.

ArwenQin: Moved the function in staff.ts to BusinessService.searchReviews(). Now call BusinessService.searchReviews() directly in table.vue

ArwenQin:
> > > Will work on this further, any suggestions?
> > 
> > 
> > I didn't look at this deeply but it looks like the list isn't correctly anchored to the dropdown menu...
> 
> https://v2.vuetifyjs.com/en/components/selects/#multiple

When multiple statuses are selected, showing "Multiple":
image

When only one is selected, showing that status:
image

How about this?

severinbeauvais:
> How about this?

Looks OK to me but double-check with Andy.

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2928-f8q1wi1m.web.app

severinbeauvais:
I think the date filter should display a date picker (or 2 date pickers to select a range), and the URL parameters should be in yyyy-mmm-dd format. Did you find any examples that do this? If it's easy, do it in this ticket. Otherwise, create another ticket.

ArwenQin:
> I think the date filter should display a date picker (or 2 date pickers to select a range), and the URL parameters should be in yyyy-mmm-dd format. Did you find any examples that do this? If it's easy, do it in this ticket. Otherwise, create another ticket.

I find an example here:
https://dev.assets.bcregistry.gov.bc.ca/dashboard
image
May I know which project/file is it in? Thanks!

severinbeauvais:
> May I know which project/file is it in? Thanks!

https://github.com/bcgov/ppr/tree/main/ppr-ui

Look in views/ for the dashboard.

seeker25 (approved):
LGTM

severinbeauvais:
Arwen, it looks like you want to implement the date filter before this gets merged, right?

ArwenQin:
> Arwen, it looks like you want to implement the date filter before this gets merged, right?

Yes, I'm working on the time picker now :)

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2928-f8q1wi1m.web.app

PR #2943, opened 2024-08-01
22039 - MBR Filing Status

ArwenQin:
*Issue #:*
https://github.com/bcgov/entity/issues/22039

*Description of changes:*
- Updated the status and action displays for different filing intermediate states
- FE will display as  "Future Effective - Paid", action will be Open Application/Open Registration". No delete action.
- Updated for AWAITING_REVIEW, CHANGE_REQUESTED, REJECTED
- Added unit tests


auth-web/src/stores/business.ts
```diff
@@ -90,6 +90,8 @@ export const useBusinessStore = defineStore('business', () => {
       ...addConditionalProperty(resp.modified, { modified: resp.modified }),
       ...addConditionalProperty(resp.modifiedBy, { modifiedBy: resp.modifiedBy }),
       ...addConditionalProperty(resp.nrNumber, { nrNumber: resp.nrNumber }),
+      ...addConditionalProperty(resp.state, { status: resp.state }),
+      ...addConditionalProperty(resp.effectiveDate, { effectiveDate: resp.effectiveDate }),
```
ArwenQin, line 94: I tried to get the status and effectiveDate property. But the Business Object is still not getting the state and effectiveDate property by doing so, can I get some guidance or suggestions?

ArwenQin:
Will add unit tests later

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -82,6 +82,16 @@ export const useAffiliations = () => {
   /** Returns the status of the affiliation. */
   const status = (business: Business): string => {
     if (isTemporaryBusiness(business)) {
+      // Reflect 'Pending | Payment Incomplete' status
+      if (business.status === BusinessState.PENDING) {
```
seeker25, line 86: Would recommend moving this to the API if possible

auth-web/src/stores/business.ts
seeker25: take a look at the search api?
image

ArwenQin:
FE:
image
AWAITING_REVIEW:
image
CHANGE_REQUESTED:
image

auth-web/src/stores/business.ts
ArwenQin: > I tried to get the status and effectiveDate property. But the Business Object is still not getting the state and effectiveDate property by doing so, can I get some guidance or suggestions?

effectiveDate property is working! I assume the state/status is not updated in the api?

severinbeauvais: The Auth API probably needs a small update to pass things along. I assume @vysakh-menon-aot will work on it next as part of his ticket.

vysakh-menon-aot: image

severinbeauvais:
image

This doesn't match the latest UI design. To be honest, Andy is still working on them, but between the 3 of us we should be able to figure out what to do.

https://www.figma.com/design/wVF8EcUCUR20YyqA0wHVJp/Continuation-In-V5?node-id=2983-13864&t=tdBPWazBJSVFFHYW-0

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2943-r0ywaf4j.web.app

ArwenQin:
> image
> 
> This doesn't match the latest UI design. To be honest, Andy is still working on them, but between the 3 of us we should be able to figure out what to do.
> 
> https://www.figma.com/design/wVF8EcUCUR20YyqA0wHVJp/Continuation-In-V5?node-id=2983-13864&t=tdBPWazBJSVFFHYW-0

I will update!

auth-web/src/composables/affiliations-factory.ts
severinbeauvais: Arwen, did you do this because some combined states are more important than others and we want to show those?

Example 1: don't show future effective -- awaiting review is more important
Example 2: don't show future effective -- change requested is more important)?

If so, please make this clear in the code with sufficient comments.

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -82,6 +82,18 @@ export const useAffiliations = () => {
   /** Returns the status of the affiliation. */
   const status = (business: Business): string => {
     if (isTemporaryBusiness(business)) {
+      // Reflect 'Pending | Payment Incomplete' status
+      if (business.status === 'PENDING') {
+        return BusinessState.PENDING
+      }
+      if (business.status) {
+        return BusinessState[business.status]
+      }
```
severinbeauvais, line 91: I'm confused why you need lines 86-88. Won't like 90 handle it?

auth-web/src/stores/business.ts
```diff
@@ -90,10 +88,11 @@ export const useBusinessStore = defineStore('business', () => {
       ...addConditionalProperty(resp.modified, { modified: resp.modified }),
       ...addConditionalProperty(resp.modifiedBy, { modifiedBy: resp.modifiedBy }),
       ...addConditionalProperty(resp.nrNumber, { nrNumber: resp.nrNumber }),
+      ...addConditionalProperty(resp.effectiveDate, { effectiveDate: resp.effectiveDate }),
       ...addBooleanProperty(resp.adminFreeze, false, 'adminFreeze'),
       ...addBooleanProperty(resp.goodStanding, true, 'goodStanding'),
       ...addBooleanProperty(resp.inDissolution, false, 'inDissolution'),
-      ...addConditionalProperty(resp.state, { status: resp.state })
+      status: resp.state || resp.draftStatus
```
ArwenQin, line 95: I updated this:
if there is a state ("ACTIVE" or "HISTORICAL"), business.status = state
if there is a draftStatus, the business.status would be the draftStatus.
It should work, but after I update it, there is an error:
image
Could you please help debug? @severinbeauvais  Thanks!

auth-web/src/composables/affiliations-factory.ts
ArwenQin: Yes, I will add comments

auth-web/src/stores/business.ts
severinbeauvais: You have a null string on which `toUpperCase` is being called. Check your data. Use conditional chaining if needed.

auth-web/src/composables/affiliations-factory.ts
ArwenQin: Agree, I will fix it.

severinbeauvais:
The third item here should be in Awaiting Review status.

image

image

auth-web/src/composables/affiliations-factory.ts
```diff
@@ -98,12 +110,17 @@ export const useAffiliations = () => {
         return NrDisplayStates[NrState.HOLD]
       } else return NrDisplayStates[state] || 'Unknown'
     }
-    if (business.status) {
-      return business.status.charAt(0)?.toUpperCase() + business.status?.slice(1)?.toLowerCase()
+    if (business.status === 'HISTORICAL') {
+      return BusinessState.HISTORICAL
```
severinbeauvais, line 114: Is this code needed or will line 88 cover it?

ArwenQin:
> The third item here should be in Awaiting Review status.
> 
> image
> 
> image

This is because the draftStatus is not correctly read yet, I will fix that issue!

auth-web/src/composables/affiliations-factory.ts
severinbeauvais: Oh, sorry, I see you already have those comments further down (lines 92-93).

@seeker25 , what are you thinking should be in the API instead?

auth-web/src/composables/affiliations-factory.ts
severinbeauvais: Also, isn't there always a `status` property in the business object? In this case, the following code will not be run...?

auth-web/src/stores/business.ts
ArwenQin: Even when I set the status to be null, or empty string, there is no error. The error only shows when I set the status to be resp.draftStatus.

auth-web/src/composables/affiliations-factory.ts
seeker25: return the enum that the UI needs and not create these pseudo statuses that live just in the UI, the result will be cleaner UI code that we could easily rewire up to a new UI if required

seeker25: There's already way too many conditional branches in the UI for this kind of thing @hfekete stated this before

severinbeauvais: @seeker25 , please wait for the next commit. I believe a lot of this code will resolve to a single statement.

By the way, this code is not about creating pseudo-statuses -- it's about changing an enum to a user-friendly string. Do  you agree that this belongs in the UI?

PS - If I'm not mistaken, this code is handling the response of the `affiliations` call.

seeker25: Why not just return the correct status enum from the API, then map that to a UI user friendly string?

severinbeauvais: > Why not just return the correct status enum from the API, then map that to a UI user friendly string? No conditions needed

I believe that is exactly what the code is doing presently. `BusinessState` is a string lookup enum:

image

seeker25: Looks conditional to me and not just looking at the enum:

image

severinbeauvais: This logic is being rewritten. Please wait for the next commit.

seeker25: anyways just my thoughts, I've seen people try to refactor this code and it proved very difficult

auth-web/src/composables/affiliations-factory.ts
ArwenQin: Updated the logic

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2943-r0ywaf4j.web.app

auth-web/src/composables/affiliations-factory.ts
ArwenQin: 1.  When the business has a effectiveDate and no other draft statuses (like "AWAITING_REVIEW" and "CHANGE_REQUESTED"), the draftStatus fetched from api is PAID. So the business.status would be PAID. No need to check business.effectiveDate again.
2. When the business has a effectiveDate and has a draft status, the draftStatus (like "AWAITING_REVIEW") will override PAID. The business.status would be the draftStatus fetched from api, so PAID (Effective Effective) won't show.
3. Same for PENDING

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -139,6 +139,12 @@
         <!-- Status -->
         <template #item-slot-Status="{ item }">
           <span>{{ status(item) }}</span>
+          <!-- Future Effective Icon for PAID status -->
+          <template v-if="status(item) === BusinessState.PAID">
+            <v-icon class="future-effective-icon">
+              mdi-information-outline
+            </v-icon>
+          </template>
```
severinbeauvais, line 147: 1. A filing may be in Paid (or Approved) status temporarily if the Filer is stuck. I think you also need to check the future-effective flag.
2. A tooltip should be attached to this icon. Andy has prepared a short tooltip that includes the FE date.

ArwenQin: Just to confirm:
- Paid and Future Effective: show Future Effective and the tooltip that includes the FE date
- Paid and not FE (the Filer is stuck): show Paid.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -525,4 +533,9 @@ export default defineComponent({
   }
 }
 
+.future-effective-icon {
+  color: #4169E1 !important;
+  margin-left: 4px;
```
severinbeauvais, line 538: I recommend using a class of "ml-1" for alignment.

Also, can you use a color of `#F8661A` ? If that's acceptable then I think you can specify it right in the template (color="xxx"), so you don't need this CSS.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: Correct!

severinbeauvais:
These NRs have the wrong status:

image

ArwenQin:
> These NRs have the wrong status:
> 
> image

May I know what is the correct status for these NRs. They are in "Pending Staff Review" currently, before my changes

severinbeauvais:
> > These NRs have the wrong status:
> 
> May I know what is the correct status for these NRs. They are in "Pending Staff Review" currently, before my changes

My mistake. You are correct: these NRs are in Pending Staff Review status. Please ignore my comment.

seeker25 (approved):
LGTM, someone from entities should approve though as well

auth-web/src/util/constants.ts
```diff
@@ -405,7 +405,13 @@ export enum NrRequestTypeStrings {
...
 export enum BusinessState {
     ACTIVE = 'Active',
-    DRAFT = 'Draft'
+    DRAFT = 'Draft',
+    AWAITING_REVIEW = 'Pending Staff Review',
+    CHANGE_REQUESTED = 'Change Requested',
+    REJECTED = 'Rejected',
+    HISTORICAL = 'Historical',
+    PENDING = 'Pending | Payment Incomplete',
+    PAID = 'Future Effective' // Future effective and paid
```
severinbeauvais, line 414: To be clear, this state could mean:
- a future effective filing
- an immediate filing that the Filer hasn't processed yet (maybe "Paid | Pending"? ask Andy) (this could also be an overdue future effective filing)

In order to keep the strings here, I think you'll need a new business state here (PAID_FE?), and then you'll need to select between them in affiliations-factory.ts.

Also, you should handle APPROVED state the same as PAID (and Approved could also be FE).

severinbeauvais:
https://bcregistry-account-dev--pr-2943-r0ywaf4j.web.app/account/2288/business

image

image

image

It looks like `state` is null/undefined above.

ArwenQin:
> https://bcregistry-account-dev--pr-2943-r0ywaf4j.web.app/account/2288/business
> 
> image
> 
> image
> 
> image
> 
> It looks like `state` is null/undefined above.
> 
> **Update: this may be because you're doing this:**
> 
> image
> 
> **but the status (in this case, APPROVED) in not in the enum.**

Yes, it's due to APPROVED, I will update now.

auth-web/src/util/constants.ts
severinbeauvais: Sample future-effective, approved continuation application:

https://dev.business.bcregistry.gov.bc.ca/Tzg27dGHVo?filing_id=150492&accountid=2288

severinbeauvais:
Please also update the app version again. (Your change was swallowed by the rebase.)

auth-web/src/util/constants.ts
severinbeauvais: Would it make sense to have a "getBusinessState" function in which to encapsulate the string lookup and special logic? Maybe we don't need the enum either if everything is in one place.

Idea:
```
function getBusinessState (status): string {
   switch (status) {
      case 'PAID': return (isFutureEffective) ? 'Future Effective' : 'Paid | Pending'
      case 'DRAFT': return 'Draft'
      ...
      default: return '' // or 'Unknown'?
   }
}
```

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
ArwenQin: Updated!
- Paid and Future Effective (approved or not-approved with no other status): show Future Effective and the tooltip that includes the FE date
- Paid and not FE (the Filer is stuck): show Paid | Pending.
- Approved: show Approved if it's not a FE

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
ArwenQin: Updated

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2943-r0ywaf4j.web.app

auth-web/src/util/constants.ts
ArwenQin: Updated status:
- PAID (filer pending)
- APPROVED
- PAID_FE

I think may be better to keep the enums, because I used them in the AffliationAction file to determine actions for different statuses; and computed some statuses when fetch from api

ArwenQin:
2 action items left for now:
- The 'Paid | Pending' status, for immediate filing, and for overdue FE (confirming with Andy)
- Will add unit tests

severinbeauvais:
Re:
image

Ask Andy but I think we should display the future effective time also.

severinbeauvais:
Looks great! (BCREG0020 / Severin Dev Account)

image

- Do we really want to say "Manage Business" for a FE continuation? It's not a business yet. I think it should saw "Open Application". Confirm with Andy.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
severinbeauvais: Perfect, thanks!

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
```diff
@@ -310,13 +328,19 @@ export default defineComponent({
       return affiliationInviteInfos.length > 0 && affiliationInviteInfos[0].status
     }
 
+    // Method to format dates
+    const formatDate = (dateString) => {
+      return moment(dateString).format('MMMM D, YYYY') // Format like "May 5, 2024"
```
severinbeauvais, line 333: Does this fail safely if dateString is null/undefined/empty?

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2943-r0ywaf4j.web.app

auth-web/src/util/constants.ts
severinbeauvais: OK. I appreciate the effort to not make too many changes. Hopefully when this is ported to another UI, we can simplify the code somewhat.

auth-web/src/components/auth/manage-business/AffiliatedEntityTable.vue
ArwenQin: Will update this.

ArwenQin: Updated, and show the date + time

ArwenQin:
Will update the package version when it's ready to merge, in case rebase is needed

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-2943-r0ywaf4j.web.app

severinbeauvais (approved):
LGTM!

Please go ahead and update the app version, and then we can merge this.

~While you're here, can you fix this bug? Should say "This continuation application" in this case.~ _will be new ticket_

severinbeauvais:
Oh, here's another one:

image

Should be "This registration".

severinbeauvais:
And another. If this is too much, let me know and I'll create a separate bug ticket.

image

PR #2952, opened 2024-08-08
20721 - Client changes from EFT to PAD

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/20721

*Description of changes:*


auth-web/src/components/pay/CompletePaymentDetails.vue
```diff
@@ -142,50 +184,127 @@ export default defineComponent({
...
         }
       }
+      state.isLoading = false
     }
 
+    const cancelButtonText = computed(() => {
+      return props.changePaymentType === PaymentTypes.PAD ? 'Back' : 'Cancel'
+    })
+
+    const nextButtonText = computed(() => {
```
seeker25, line 293: these should be wrapped under the `state`?

auth-web/src/components/auth/common/PaymentMethods.vue
```diff
@@ -353,8 +352,27 @@ export default defineComponent({
...
     }
 
-    const paymentMethodSelected = (payment, isTouch = true) => {
-      if (payment.type === PaymentTypes.BCOL && isTouch && selectedPaymentMethod.value !== PaymentTypes.BCOL) {
+    const hasBalanceOwing = async () => {
+      try {
+        const responseData = await getStatementsSummary(props.currentOrganization.id)
+        return responseData?.totalDue || responseData?.totalInvoiceDue
+      } catch (error) {
+        console.log(error)
```
seeker25, line 360: Probably should do something if the call fails?

auth-web/src/components/pay/CompletePaymentDetails.vue
```diff
@@ -142,41 +186,106 @@ export default defineComponent({
...
       goToPaymentOptions()
     }
 
-    const complete = async () => {
+    function isPADValid (isValid) {
+      state.padValid = isValid
+    }
+
+    async function verifyPAD () {
+      const verifyPad: PADInfoValidation = await orgStore.validatePADInfo()
```
seeker25, line 204: What if this fails?

auth-web/src/components/pay/CompletePaymentDetails.vue
```diff
@@ -142,41 +186,106 @@ export default defineComponent({
...
+          return
+        }
+        createRequestBody = {
+          paymentInfo: {
+            paymentMethod: PaymentTypes.BCOL
+          },
+          bcOnlineCredential: state.bcolInfo
+        }
+      } else if (props.changePaymentType === PaymentTypes.PAD && state.padValid) {
+        isValid = await verifyPAD()
```
seeker25, line 238: What if this fails?

auth-web/src/components/pay/CompletePaymentDetails.vue
```diff
@@ -142,41 +186,106 @@ export default defineComponent({
...
+      const { isValid, createRequestBody } = await getCreateRequestBody()
+
+      if (isValid) {
+        try {
+          await orgStore.updateOrg(createRequestBody)
+          orgStore.setCurrentOrganizationPaymentType(props.changePaymentType)
+          state.isPaymentChanged = true
+          goToPaymentOptions()
+        } catch (error) {
+          switch (error.response.status) {
```
seeker25, line 277: Should probably do something like this for the two requests above, I've see banking validation fail before

seeker25 (approved):
Approving, risky if the calls fail though

auth-web/src/components/auth/common/PaymentMethods.vue
seeker25: unlikely to fail

auth-web/src/components/pay/CompletePaymentDetails.vue
seeker25: I think this calls the CFS service, could easily fail

auth-web/src/components/pay/CompletePaymentDetails.vue
seeker25: I think this calls the CFS service, could easily fail

auth-web/src/components/pay/CompletePaymentDetails.vue
seeker25: Already handled we're good

auth-web/src/components/pay/CompletePaymentDetails.vue
seeker25: handled above

PR #2982, opened 2024-08-28
22776 - Shortname History Table Fixes

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/22776

*Description of changes:*
- Fix reload should go back to page 1
- Fix issue where infinite scroll back is no longer triggering if it previously has reached an end state even after reload

### DIFF
--- a/auth-web/src/components/pay/eft/ShortNamePaymentHistory.vue
+++ b/auth-web/src/components/pay/eft/ShortNamePaymentHistory.vue
@@ -67,6 +67,7 @@
     </ModalDialog>
     <BaseVDataTable
       id="eft-transactions-table"
+      ref="historyTable"
       class="transaction-list"
       itemKey="id"
       :loading="state.loading"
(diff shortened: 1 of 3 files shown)


Jxio (approved):

auth-web/src/components/pay/eft/ShortNamePaymentHistory.vue
```diff
@@ -152,6 +153,7 @@ import { DEFAULT_DATA_OPTIONS } from '../../datatable/resources'
 import { EFTTransactionState } from '@/models/eft-transaction'
 import ModalDialog from '@/components/auth/common/ModalDialog.vue'
 import PaymentService from '@/services/payment.services'
+import { Vue } from 'vue-property-decorator'
```
seeker25, line 156: Not too sure why we're introducing this, we're actively trying to get away from using Vue-class-component and vue-property-decorartor

seeker25: `import Vue from 'vue'`

ochiu: Will fix this in my next PR, as it is another FE ticket.

seeker25: all good, I was too slow on the review haha

PR #3140, opened 2024-11-08
Use pg8000 consistently

seeker25:


seeker25:
I'm pretty sure @bolyachevets  already fixed all the issues at runtime, which is why this just works? unless I'm missing something

pwei1018 (approved):

PR #3143, opened 2024-11-12
Split out sandbox payments from Auth

sameer0422:
*Issue #:*
https://github.com/bcgov/entity/issues/24249

*Description of changes:*


auth-api/src/auth_api/services/api_gateway.py
```diff
@@ -213,6 +215,36 @@ def _consumer_exists(cls, email, env):
     @classmethod
     @user_context
     def _create_payment_account(cls, org: OrgModel, **kwargs):
+        """Create a payment account for environment.
+
+        Steps:
+        - Get the payment account details.
+        - Mask the details and call pay endpoint to create a account.
+        """
+        if not current_app.config.get("PAY_API_URL"):
```
seeker25, line 224: Which PAY_API_URL is this? will it ever be undefined? if it is, shouldn't we exception?

auth-api/src/auth_api/services/api_gateway.py
```diff
@@ -213,6 +215,36 @@ def _consumer_exists(cls, email, env):
...
+            logger.warning("URL not provided, skipping pay account creation")
+            return
+        user: UserContext = kwargs["user_context"]
+        pay_account = cls._get_pay_account(org, user)
+        pay_request = {
```
seeker25, line 229: Alot of code duplication here?

auth-api/src/auth_api/services/api_gateway.py
```diff
@@ -240,6 +272,14 @@ def _create_payment_account(cls, org: OrgModel, **kwargs):
         }
 
         cls._create_sandbox_pay_account(pay_request, user)
+    
+    @classmethod
+    def _create_pay_account(cls, pay_request, user):
```
seeker25, line 277: duplicate below, when the only difference is the config secret

seeker25:
re-open if needed

PR #3169, opened 2024-12-05
21429 - staff account search for members

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/21429

*Description of changes:*
- search_orgs by member name or username support and members payload

### DIFF
--- a/auth-api/src/auth_api/models/org.py
+++ b/auth-api/src/auth_api/models/org.py
@@ -171,6 +171,20 @@ class Org(Versioned, BaseModel):  # pylint: disable=too-few-public-methods,too-m
             query = query.filter(Org.branch_name.ilike(f"%{search.branch_name}%"))
         if search.name:
             query = query.filter(Org.name.ilike(f"%{search.name}%"))
+        if search.member_search_text:
+            member_exists_subquery = text("""
+                EXISTS (
+                    SELECT 1
+                    FROM memberships
+                    JOIN users ON users.id = memberships.user_id
+                    WHERE memberships.org_id = orgs.id
+                    AND memberships.status = 1
+                    AND users.status = 1
+                    AND CONCAT(users.last_name, ' ', users.first_name, ' ', users.username) ILIKE :member_search_text
+                )
+                """).params(member_search_text=f"%{search.member_search_text}%")
+
+            query = query.filter(member_exists_subquery)
 
         query = cls._search_by_business_identifier(query, search.business_identifier, environment)
         query = cls._search_for_statuses(query, search.statuses)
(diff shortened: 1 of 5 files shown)


auth-api/src/auth_api/models/org.py
```diff
@@ -171,6 +171,20 @@ def search_org(cls, search: OrgSearch, environment: str):
...
+            member_exists_subquery = text("""
+                EXISTS (
+                    SELECT 1
+                    FROM memberships
+                    JOIN users ON users.id = memberships.user_id
+                    WHERE memberships.org_id = orgs.id
+                    AND memberships.status = 1
+                    AND users.status = 1
+                    AND CONCAT(users.last_name, ' ', users.first_name, ' ', users.username) ILIKE :member_search_text
+                )
```
ochiu, line 184: Needed to use string subquery due to model cyclic dependencies. 
Membership model has an org import.

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -63,6 +63,8 @@ def search_organizations():
         extract_numbers(request.args.get("id", None)),
         request.args.get("decisionMadeBy", None),
         request.args.get("orgType", None),
+        bool(request.args.get("includeMembers", False)),
+        request.args.get("members", None),
```
ochiu, line 67: Added query param (includeMembers) to actually include members payload within an org, by default this is off so there is less data coming back.

Query param "members" is the actual search text used for name or username

auth-api/src/auth_api/services/org.py
```diff
@@ -800,6 +800,11 @@ def search_orgs(search: OrgSearch, environment):  # pylint: disable=too-many-loc
                         if include_invitations and org.invitations
                         else []
                     ),
+                    "members": (
+                        MembershipSchema(exclude=("org", "user.contacts")).dump(org.members, many=True)
```
ochiu, line 804: org already has a relationship in the model for members, we can just pull the data and use the schema conversion

auth-api/src/auth_api/models/org.py
seeker25: Haha, this should be refactored to a service

seeker25: As long as there is a unit test should be ok, but I'll leave it up to you to decide

seeker25: I think we should make a ticket for the `sbc-connect` team to break alot of these dependencies as well?

ochiu: Yes, I started looking at doing a refactor, and it became a larger undertaking then I wanted for the scope of this ticket. Since it was just a subquery I kept it simple. 

I agree we should probably have some kind of tech debt ticket to clean this up.

PR #3192, opened 2025-01-02
24920 - Redirect to new BRD based on FF and Query Param

JazzarKarim:
*Issue #:*
https://github.com/bcgov/entity/issues/24920

*Description of changes:*
- Redirect to the new BRD if the FF `enable-business-registry-dashboard` and if the Query Param `newbrd` is true.


auth-web/src/routes/router.ts
```diff
@@ -229,6 +230,13 @@ export function getRoutes (): RouteConfig[] {
     {
       path: '/account/:orgId',
       name: 'account',
+      beforeEnter: (to, from, next) => {
+        if (LaunchDarklyService.getFlag(LDFlags.EnableBusinessRegistryDashboard) && to.query.newbrd === 'true') {
+          window.location.href = 'https://business-registry-dev.web.app/en-CA/'
```
JazzarKarim, line 235: That query parameter part is just a hacky way of doing targeted FFs. Sev and I had a conversation about this, enabling targeted FFs via LD is kind of a mess. This way (although not ideal), achieves that.

The nice thing about this `beforeEnter` is that it actually doesn't do the multiple redirects that the business dashboard UI was suffering from. The redirect is actually very fast.

One small thing though, I have the link hardcoded now. Do you think I should move to using the environment file and add the link to 1Pass/Vaults?

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app

auth-web/src/routes/router.ts
seeker25: yes

JazzarKarim: Deal. Can you please help me add the links to 1Pass Travis? (I don't have access)

auth-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.6.128",
+  "version": "2.6.129",
```
severinbeauvais, line 3: If this is the beginning of the end for the MBR, maybe we should finally go to 2.7.0? 😆

auth-web/src/routes/router.ts
severinbeauvais, line 235: Is there any way to retain any additional query parameters on the redirect? I'm thinking about Kevin's magic link that route to the MBR instead of the BRD. Maybe we could solve that differently though. Thoughts?

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app

auth-web/src/routes/router.ts
JazzarKarim: You're thinking we want to retain query parameters on the redirect from the old to new one?

Ahhh, I think I understand what you're saying. The link is going to have the NR ID and number/email. That link points towards the new one but eventually there's gonna be a redirect where those params are lost, right?

severinbeauvais: The new code LGTM.

severinbeauvais: Yes, exactly.

I'm thinking about the interim time where Kevin's emails use the MBR URL. I've asked him to look at importing the BRD URL instead, so maybe we don't need to retain any URL parameters in the code above.

Let's start a chat about how (and when) to transition to the new BRD. For one thing, I doubt we can just change the Auth Web URL in 1Password, since Auth Web will still exist for the staff dashboard. So, then, how do we get all links to the old MBR to end up on the new BRD? Specifically, I mean as a next step to the code above.

JazzarKarim: If he can have that link directly point to the new one, that would be amazing. CC: @kzdev420 

As for how we would get all links to point to new BRD, thoughts about just removing the query parameter check? This way, as long as the FF is enabled, that redirection will happen (very similar to what was done for the business dashboard UI by BTR). I don't think this route deals with any other query parameters. I can double check. Thoughts?

severinbeauvais: > As for how we would get all links to point to new BRD, thoughts about just removing the query parameter check? This way, as long as the FF is enabled, that redirection will happen (very similar to what was done for the business dashboard UI by BTR). I don't think this route deals with any other query parameters. I can double check. Thoughts?

I'm concerned about all the possible places that route to the MBR. Eventually they should route directly to the BRD (without redirection). I'm thinking that several places will need to be updated to use the new env key.

It's a similar issue for the new Business Dashboard UI -- I think we'll want to keep the Filings UI URL (for local filings) and several places will need to be updated to route directly to the BD UI using its new env key.

JazzarKarim: You're absolutely right. I'm wondering now as well if there are routes from Assets' UI for example.

Sev, are you OK with me to create a ticket and investigate how are we going to do this transition? Part of it as well is asking people around (mainly the BTR team).

severinbeauvais: Yes, please create future ticket(s) (for when we decide we're not going back, and we have a bit of time).

I'm thinking we should keep a simple redirect in place of the old MBR in Auth Web, just in case we miss some place that should route to BRD instead. (As well as deleting all obsolete code.)

auth-web/src/routes/router.ts
```diff
@@ -229,6 +230,13 @@ export function getRoutes (): RouteConfig[] {
     {
       path: '/account/:orgId',
       name: 'account',
+      beforeEnter: (to, from, next) => {
+        if (LaunchDarklyService.getFlag(LDFlags.EnableBusinessRegistryDashboard) && to.query.newbrd === 'true') {
```
severinbeauvais, line 234: Wait. I think this is backwards.

We want to redirect if the FF is True and we _don't have_ a URL parameter telling us to stay here anyway.

That URL parameter should be named `dont-redirect` or something.

severinbeauvais: Also, how did you determine that this is the route to intercept? Is it the default route?

JazzarKarim: Thanks for catching this Sev. Should be fixed now.

JazzarKarim: > Also, how did you determine that this is the route to intercept? Is it the default route?

It is the default route for the MBR.

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app

auth-web/src/routes/router.ts
```diff
@@ -231,7 +231,7 @@ export function getRoutes (): RouteConfig[] {
       path: '/account/:orgId',
       name: 'account',
       beforeEnter: (to, from, next) => {
-        if (LaunchDarklyService.getFlag(LDFlags.EnableBusinessRegistryDashboard) && to.query.newbrd === 'true') {
+        if (LaunchDarklyService.getFlag(LDFlags.EnableBusinessRegistryDashboard) && to.query.noRedirect !== 'true') {
```
severinbeauvais, line 234: I wonder if you could check for just the presence of the key (regardless of its value)?

Ref: https://stackoverflow.com/questions/4557387/is-a-url-query-parameter-valid-if-it-has-no-value

severinbeauvais: Unless you find that it's bad practise. I'm still reading up on this.

severinbeauvais: Nice ref: https://developers.google.com/search/docs/crawling-indexing/url-structure

severinbeauvais: Maybe just check for it to not be empty?

JazzarKarim: Updated. Now, no matter what the noRedirect value is, it won't redirect anymore. So `?noRedirect=Waffles` will stay on the old MBR.

JazzarKarim: Small note, `noRedirect` with no value will not work. It will still redirect.

severinbeauvais: I can't wait to test this. Poor Waffles.

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app

auth-web/src/routes/router.ts
JazzarKarim: New ticket:
https://app.zenhub.com/workspaces/entities---olga-65af15f59e89f5043c2911f7/issues/gh/bcgov/entity/25136

auth-web/src/routes/router.ts
severinbeauvais: > Small note, `noRedirect` with no value will not work. It will still redirect.

Interesting. Must be the way `to.query` is constructed. No matter.

severinbeauvais: So, with this URL, I should  get redirected? It's not working for me.
https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/

JazzarKarim: > So, with this URL, I should get redirected? It's not working for me. https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/

https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/account/3040/business?noRedirect=waffles

or

https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/account/3040/business

JazzarKarim: The one above will not redirect while the one below (the normal one which people will be using), will redirect.

severinbeauvais:
While you're here, I noticed this console error:

image

I _think_ the new URL is gold, not silver.

auth-web/src/routes/router.ts
severinbeauvais: OK, that worked.

However, I was expecting this URL to redirect, too. Maybe this URL isn't changing to the format above correctly?
https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/

severinbeauvais: Similarly, I was expecting this to not redirect, but it did:
https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/decide-business?noRedirect=Waffles

JazzarKarim: No, the above link will take you to [Business Registry Home](https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/decide-business)

This part is outside the MBR/BRD's jurisdiction. This part is outside.

JazzarKarim: I'm pressing on the link and it's not redirecting 🤔

severinbeauvais: I think I am confusing myself with which route should redirect (or not) to the new BRD...

The URLs I have been testing with take me to the "decide business" page, which is staying in Auth Web, right?
Eg, https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/decide-business

It's only the MBR page that should redirect, and that page is on this example URL:
https://bcregistry-account-dev--pr-3192-2mpoe0is.web.app/account/2288/business

If the above URL is the _only_ way to get to the MBR then I think what you've done is perfect. Thoughts?

JazzarKarim: Precisely, you're spot on. This is exactly what I had in mind. I want to ONLY redirect the MBR page/portion to the new one. Otherwise, I want to stay in auth web since there are all sorts of pages/features in auth web that are outside the MBR.

What you said is exactly what I wanted to achieve and what I think is the perfect way to go on about doing this.

severinbeauvais (approved):

severinbeauvais:
One last thing: when you merge this, please let developers know that from now on, they'll be using the new BRD (and the query parameter to not redirect, if they need it).

auth-web/src/routes/router.ts
severinbeauvais: > I'm pressing on the link and it's not redirecting 🤔

I just went to the page. I didn't realize I had to click the button to see the MBR and then it would redirect.

My fault. It's all good.

auth-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.6.128",
+  "version": "2.7.0",
```
seeker25, line 3: WOOHOO!

auth-web/src/routes/router.ts
```diff
@@ -229,6 +230,13 @@ export function getRoutes (): RouteConfig[] {
     {
       path: '/account/:orgId',
       name: 'account',
+      beforeEnter: (to, from, next) => {
+        if (LaunchDarklyService.getFlag(LDFlags.EnableBusinessRegistryDashboard) && !to.query.noRedirect) {
+          window.location.href = ConfigHelper.getNewBusinessRegistryDashboardUrl()
```
severinbeauvais, line 235: I wonder if this should return a new route (instead of hard-overwriting the URL like this)? That way, it could maybe retain the URL query parameters (eg, account id)...?

Ref: https://router.vuejs.org/guide/advanced/navigation-guards.html#Per-Route-Guard

JazzarKarim: That's the plan 👍

PR #3196, opened 2025-01-06
25138 - Update firm affiliation logic to work in sandbox

sameer0422:
*Issue #:*
https://github.com/bcgov/entity/issues/25138

*Description of changes:*


argush3:
looks good.  are linting errors and tests being addressed right now?

auth-api/devops/vaults.gcp.env
```diff
@@ -30,6 +30,7 @@ NOTIFY_API_VERSION="op://API/$APP_ENV/notify-api/NOTIFY_API_VERSION"
 PAY_API_URL="op://API/$APP_ENV/pay-api/PAY_API_URL"
 PAY_API_VERSION="op://API/$APP_ENV/pay-api/PAY_API_VERSION"
 LEGAL_API_URL="op://API/$APP_ENV/legal-api/LEGAL_API_URL"
+LEGAL_SANDBOX_API_URL="op://API/sandbox/legal-api/LEGAL_API_URL"
```
pwei1018, line 33: It shouldn't need sandbox url. If the APP deploy to sandbox environment, the legal_api_url value should be sandbox url.

sameer0422: But we don't have sandbox instance for auth-api.
We are using prod instance for sandbox and prod for now

pwei1018: Are you sure this approach doesn't "pollute" the Auth production data?

argush3: this specific change won't but in general sharing auth prod with sandbox is probably not ideal.  we are using the **environment** field in the **affiliations** table to try to differentiate between the environments.

seeker25: Patrick we won't have auth-api setup in time for January 10th, this is just a short term fix. Long term we definitely need the sandbox environment running in GCP as a separate environment.

auth-api/poetry.lock
```diff
@@ -1,114 +1,99 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
```
seeker25, line 1: Updating all the deps, probably should require regression before pushing to PROD..

Is it possible to just update the deps needed?

argush3 (approved):

PR #3210, opened 2025-01-17
24648 - contact_centre_staff role fixes

rodrigo-barraza:
*Issue #:*
https://github.com/bcgov/entity/issues/24648

*Description of changes:*
Fixes for contact_centre_staff role:
Hiding business registry tile
Hiding staff personal property registry tile
Hiding Safe Email List Tile
Displaying Transaction records
Access type now can't be edited

### DIFF
--- a/auth-web/src/views/auth/staff/StaffDashboardView.vue
+++ b/auth-web/src/views/auth/staff/StaffDashboardView.vue
@@ -380,9 +380,7 @@ export default defineComponent({
       searchIdentifier: '',
       canSearchFAS: computed((): boolean => currentUser.value?.roles?.includes(Role.FasSearch)),
       canViewAccounts: computed((): boolean => currentUser.value?.roles?.includes(Role.StaffViewAccounts)),
-      canViewAllTransactions: computed((): boolean => currentUser.value?.roles?.some(
-        role => role === Role.ViewAllTransactions || role === Role.ContactCentreStaff
-      )),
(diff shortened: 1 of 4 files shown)


seeker25:
Missing the change needed for VS, we need that in main before pushing to prod. 

No worries though, we'll probably just use 2.7.1 for prod on monday

auth-web/src/views/auth/staff/StaffDashboardView.vue
```diff
@@ -376,7 +380,9 @@ export default defineComponent({
       searchIdentifier: '',
       canSearchFAS: computed((): boolean => currentUser.value?.roles?.includes(Role.FasSearch)),
       canViewAccounts: computed((): boolean => currentUser.value?.roles?.includes(Role.StaffViewAccounts)),
-      canViewAllTransactions: computed((): boolean => currentUser.value?.roles?.includes(Role.ViewAllTransactions)),
+      canViewAllTransactions: computed((): boolean => currentUser.value?.roles?.some(
+        role => role === Role.ViewAllTransactions || role === Role.ContactCentreStaff
```
ochiu, line 384: I think we should grant the new group the ability to view all transactions in this case then. I have applied this in dev. We shouldn't need to code for contact centre staff here.

Let me know if this works, if not we can dig into it a bit more

seeker25 (approved):

PR #3219, opened 2025-01-22
Dialog for unsaved changes

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/24918

*Description of changes:*


auth-web/src/components/auth/account-settings/payment/AccountPaymentMethods.vue
```diff
@@ -20,6 +20,7 @@
       @is-ejv-valid="isEJVValid"
       @cancel="cancel"
       @save="save"
+      @show-warning-dialog="showWarningDialog"
```
seeker25, line 23: I think we should use the store for this, so it's global? thoughts?

seeker25 (approved):

PR #3220, opened 2025-01-23
25396 - Enable to Select Payment Method for GovN and NON BCSC

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/23596

*Description of changes:*


auth-web/src/components/auth/common/PADInfoForm.vue
```diff
@@ -243,6 +246,12 @@ export default defineComponent({
       })
     }) as unknown) as PADInfoFormState
 
+    const getCanDirective = () => {
```
seeker25, line 249: Shouldn't we just modify the v-can directive so we can toggle it with a condition?

PR #3239, opened 2025-01-31
25595 - Bubble up error from pay-api while creating account

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/25595

*Description of changes:*

### DIFF
--- a/auth-api/src/auth_api/services/org.py
+++ b/auth-api/src/auth_api/services/org.py
@@ -227,65 +227,78 @@ class Org:  # pylint: disable=too-many-public-methods
         **kwargs,
     ):
         """Add payment settings for the org."""
-        pay_url = current_app.config.get("PAY_API_URL")
-        org_name_for_pay = f"{org_model.name}-{org_model.branch_name}" if org_model.branch_name else org_model.name
-        pay_request = {
-            "accountId": org_model.id,
-            # pay needs the most unique idenitfier.So combine name and branch name
-            "accountName": org_name_for_pay,
-            "branchName": org_model.branch_name or "",
-        }
+        try:
+            pay_url = current_app.config.get("PAY_API_URL")
+            org_name_for_pay = f"{org_model.name}-{org_model.branch_name}" if org_model.branch_name else org_model.name
+            pay_request = {
+                "accountId": org_model.id,
+                # pay needs the most unique idenitfier.So combine name and branch name
+                "accountName": org_name_for_pay,
+                "branchName": org_model.branch_name or "",
+            }
 
-        if payment_method:
-            pay_request["paymentInfo"] = {"methodOfPayment": payment_method}
+            if payment_method:
+                pay_request["paymentInfo"] = {"methodOfPayment": payment_method}
 
-        if mailing_address:
-            pay_request["contactInfo"] = mailing_address
+            if mailing_address:
+                pay_request["contactInfo"] = mailing_address
 
-        if payment_method and org_model.bcol_account_id:
-            pay_request["bcolAccountNumber"] = org_model.bcol_account_id
-            pay_request["bcolUserId"] = org_model.bcol_user_id
+            if payment_method and org_model.bcol_account_id:
+                pay_request["bcolAccountNumber"] = org_model.bcol_account_id
+                pay_request["bcolUserId"] = org_model.bcol_user_id
 
-        if (revenue_account := payment_info.get("revenueAccount")) is not None:
-            pay_request.setdefault("paymentInfo", {})
-            pay_request["paymentInfo"]["revenueAccount"] = revenue_account
+            if (revenue_account := payment_info.get("revenueAccount")) is not None:
+                pay_request.setdefault("paymentInfo", {})
+                pay_request["paymentInfo"]["revenueAccount"] = revenue_account
 
-        if payment_method == PaymentMethod.PAD.value:  # PAD has bank related details
-            pay_request["paymentInfo"]["bankTransitNumber"] = payment_info.get("bankTransitNumber", None)
-            pay_request["paymentInfo"]["bankInstitutionNumber"] = payment_info.get("bankInstitutionNumber", None)
-            pay_request["paymentInfo"]["bankAccountNumber"] = payment_info.get("bankAccountNumber", None)
-            pay_request["padTosAcceptedBy"] = kwargs["user_context"].user_name
-        # invoke pay-api
-        token = RestService.get_service_account_token()
-        if is_new_org:
-            response = RestService.post(
-                endpoint=f"{pay_url}/accounts", data=pay_request, token=token, raise_for_status=True
-            )
-        else:
-            response = RestService.put(
-                endpoint=f"{pay_url}/accounts/{org_model.id}", data=pay_request, token=token, raise_for_status=True
-            )
-
-        if response.status_code == HTTPStatus.OK:
-            payment_account_status = PaymentAccountStatus.CREATED
-        elif response.status_code == HTTPStatus.ACCEPTED:
-            payment_account_status = PaymentAccountStatus.PENDING
-        else:
-            payment_account_status = PaymentAccountStatus.FAILED
-
-        if payment_account_status != PaymentAccountStatus.FAILED and payment_method:
-            payment_method_description = (
-                PaymentMethod(payment_method).name if payment_method in [item.value for item in PaymentMethod] else ""
-            )
-            ActivityLogPublisher.publish_activity(
-                Activity(
-                    org_model.id,
-                    ActivityAction.PAYMENT_INFO_CHANGE.value,
-                    name=org_model.name,
-                    value=payment_method_description,
+            if payment_method == PaymentMethod.PAD.value:  # PAD has bank related details
+                pay_request["paymentInfo"]["bankTransitNumber"] = payment_info.get("bankTransitNumber", None)
+                pay_request["paymentInfo"]["bankInstitutionNumber"] = payment_info.get("bankInstitutionNumber", None)
+                pay_request["paymentInfo"]["bankAccountNumber"] = payment_info.get("bankAccountNumber", None)
+                pay_request["padTosAcceptedBy"] = kwargs["user_context"].user_name
+            # invoke pay-api
+            token = RestService.get_service_account_token()
+            if is_new_org:
+                response = RestService.post(
+                    endpoint=f"{pay_url}/accounts", data=pay_request, token=token, raise_for_status=True
                 )
-            )
-        return payment_account_status
+            else:
+                response = RestService.put(
+                    endpoint=f"{pay_url}/accounts/{org_model.id}", data=pay_request, token=token, raise_for_status=True
+                )
+
+            if response.status_code == HTTPStatus.OK:
+                payment_account_status = PaymentAccountStatus.CREATED
+            elif response.status_code == HTTPStatus.ACCEPTED:
+                payment_account_status = PaymentAccountStatus.PENDING
+            else:
+                payment_account_status = PaymentAccountStatus.FAILED
+                error_payload = response.json()
+                error_code = error_payload.get("code")
+                error_message = error_payload.get("message")
+                raise BusinessException(error_code, error_message)
+
+            if payment_account_status != PaymentAccountStatus.FAILED and payment_method:
+                payment_method_description = (
+                    PaymentMethod(payment_method).name
+                    if payment_method in [item.value for item in PaymentMethod]
+                    else ""
+                )
+                ActivityLogPublisher.publish_activity(
+                    Activity(
+                        org_model.id,
+                        ActivityAction.PAYMENT_INFO_CHANGE.value,
+                        name=org_model.name,
+                        value=payment_method_description,
+                    )
+                )
+            return payment_account_status
+
+        except HTTPError as http_error:
+            error_payload = http_error.response.json()
+            error_code = error_payload.get("code")
+            error_message = error_payload.get("message")
+            raise BusinessException(Error.ACCOUNT_CREATION_FAILED_IN_PAY, error_message)
 
     @staticmethod
     def _validate_and_raise_error(org_info: dict):


auth-api/src/auth_api/services/org.py
```diff
@@ -272,6 +273,8 @@ def _create_payment_settings(
             payment_account_status = PaymentAccountStatus.PENDING
         else:
             payment_account_status = PaymentAccountStatus.FAILED
+            error_payload = response.json()
```
Jxio, line 276: There are more than 200 unit tests failing because they did not throw an exception when they failed, possibly due to a 201 status code. Do we still need the exception here?

auth-api/src/auth_api/services/org.py
```diff
@@ -270,8 +271,12 @@ def _create_payment_settings(
             payment_account_status = PaymentAccountStatus.CREATED
         elif response.status_code == HTTPStatus.ACCEPTED:
             payment_account_status = PaymentAccountStatus.PENDING
+        elif response.status_code == HTTPStatus.CREATED:
+            payment_account_status = PaymentAccountStatus.FAILED
         else:
             payment_account_status = PaymentAccountStatus.FAILED
+            error_payload = response.json()
```
seeker25, line 278: What if the response doesn't include json? and it's a generic exception not a business exception?

auth-api/src/auth_api/services/org.py
seeker25: Probably just need to change the pay spec for testing so it returns the right status code? it's just calling a mock? maybe the spec needs to be updated?

auth-api/src/auth_api/services/org.py
Jxio: it's business exception on `def create_org(org_info: dict, user_id)`

Jxio: Removed throwing an exception here and logged an error instead. It breaks many unit tests because an exception was not thrown before.

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
-            payment_method_description = (
-                PaymentMethod(payment_method).name if payment_method in [item.value for item in PaymentMethod] else ""
-            )
-            ActivityLogPublisher.publish_activity(
-                Activity(
-                    org_model.id,
-                    ActivityAction.PAYMENT_INFO_CHANGE.value,
-                    name=org_model.name,
-                    value=payment_method_description,
+            if response.status_code == HTTPStatus.OK:
```
seeker25, line 272: Candidate for match case

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
+                # pay needs the most unique idenitfier.So combine name and branch name
+                "accountName": org_name_for_pay,
+                "branchName": org_model.branch_name or "",
+            }
+            error_code = None
 
-        if payment_method:
-            pay_request["paymentInfo"] = {"methodOfPayment": payment_method}
+            if payment_method:
+                pay_request["paymentInfo"] = {"methodOfPayment": payment_method}
```
seeker25, line 243: Building the request should be refactored to a separate function. There are already alot of conditional branches that cause code complexity to go up

seeker25: If org service is getting too packed, consider building a pay service for this purpose.

auth-api/src/auth_api/services/org.py
```diff
@@ -149,11 +149,12 @@ def create_org(org_info: dict, user_id):
...
         ProductService.create_subscription_from_bcol_profile(org.id, bcol_profile_flags)
 
-        Org._create_payment_for_org(mailing_address, org, payment_info, True)
+        payment_account_status, error = Org._create_payment_for_org(mailing_address, org, payment_info, True)
 
-        # TODO do we have to check anything like this below?
-        # if payment_account_status == PaymentAccountStatus.FAILED:
-        # raise BusinessException(Error.ACCOUNT_CREATION_FAILED_IN_PAY, None)
+        if payment_account_status == PaymentAccountStatus.FAILED and error is not None:
+            logger.error(f"Account update payment Error: {error}")
```
seeker25, line 155: so if this fails, it doesn't return a 500 error? Hmm that's not so great

auth-api/src/auth_api/services/org.py
```diff
@@ -426,7 +446,11 @@ def update_org(self, org_info):  # pylint: disable=too-many-locals, too-many-sta
                 Org.send_staff_review_account_reminder(relationship_id=self._model.id)
 
         if name_updated or payment_info:
-            Org._create_payment_for_org(mailing_address, self._model, payment_info, False)
+            payment_account_status, error = Org._create_payment_for_org(
+                mailing_address, self._model, payment_info, False
+            )
+            if payment_account_status == PaymentAccountStatus.FAILED and error is not None:
+                logger.error(f"Account update payment Error: {error}")
```
seeker25, line 453: so if this fails, it doesn't return a 500 error? Hmm that's not so great

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
-                Activity(
-                    org_model.id,
-                    ActivityAction.PAYMENT_INFO_CHANGE.value,
-                    name=org_model.name,
-                    value=payment_method_description,
+            if response.status_code == HTTPStatus.OK:
+                payment_account_status = PaymentAccountStatus.CREATED
+            elif response.status_code == HTTPStatus.ACCEPTED:
+                payment_account_status = PaymentAccountStatus.PENDING
+            elif response.status_code == HTTPStatus.CREATED:
```
seeker25, line 276: Maybe a comment to explain this?

image

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
+                        ActivityAction.PAYMENT_INFO_CHANGE.value,
+                        name=org_model.name,
+                        value=payment_method_description,
+                    )
+                )
+            return payment_account_status, error_code
+
+        except HTTPError as http_error:
+            error_payload = http_error.response.json()
+            error_code = error_payload.get("error", Error.ACCOUNT_CREATION_FAILED_IN_PAY)
```
seeker25, line 305: need some code to determine if it was POST or PUT and return ACCOUNT_UPDATE_FAILED_IN_PAY or ACCOUNT_CREATION_FAILED_IN_PAY

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
-                    value=payment_method_description,
+            if response.status_code == HTTPStatus.OK:
+                payment_account_status = PaymentAccountStatus.CREATED
+            elif response.status_code == HTTPStatus.ACCEPTED:
+                payment_account_status = PaymentAccountStatus.PENDING
+            elif response.status_code == HTTPStatus.CREATED:
+                payment_account_status = PaymentAccountStatus.FAILED
+            else:
+                payment_account_status = PaymentAccountStatus.FAILED
+                if response.json:
```
seeker25, line 280: maybe: error_code = (response.json() or {}).get("error", "UNKNOWN_ERROR") ?

Although I think .json() can throw an exception?

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
-        elif response.status_code == HTTPStatus.ACCEPTED:
-            payment_account_status = PaymentAccountStatus.PENDING
-        else:
-            payment_account_status = PaymentAccountStatus.FAILED
+            if payment_method == PaymentMethod.PAD.value:  # PAD has bank related details
+                pay_request["paymentInfo"]["bankTransitNumber"] = payment_info.get("bankTransitNumber", None)
+                pay_request["paymentInfo"]["bankInstitutionNumber"] = payment_info.get("bankInstitutionNumber", None)
+                pay_request["paymentInfo"]["bankAccountNumber"] = payment_info.get("bankAccountNumber", None)
+                pay_request["padTosAcceptedBy"] = kwargs["user_context"].user_name
+            # invoke pay-api
```
seeker25, line 261: remove, obvious by reading `pay_url`

auth-api/src/auth_api/services/org.py
```diff
@@ -149,11 +149,12 @@ def create_org(org_info: dict, user_id):
...
 
-        Org._create_payment_for_org(mailing_address, org, payment_info, True)
+        payment_account_status, error = Org._create_payment_for_org(mailing_address, org, payment_info, True)
 
-        # TODO do we have to check anything like this below?
-        # if payment_account_status == PaymentAccountStatus.FAILED:
-        # raise BusinessException(Error.ACCOUNT_CREATION_FAILED_IN_PAY, None)
+        if payment_account_status == PaymentAccountStatus.FAILED and error is not None:
+            logger.error(f"Account update payment Error: {error}")
+            # TODO do we have to check anything like this below?
```
seeker25, line 156: Remove todo

auth-api/poetry.lock
```diff
@@ -2416,7 +2416,6 @@ files = [
     {file = "psycopg2-2.9.10-cp311-cp311-win_amd64.whl", hash = "sha256:0435034157049f6846e95103bd8f5a668788dd913a7c30162ca9503fdf542cb4"},
     {file = "psycopg2-2.9.10-cp312-cp312-win32.whl", hash = "sha256:65a63d7ab0e067e2cdb3cf266de39663203d38d6a8ed97f5ca0cb315c73fe067"},
     {file = "psycopg2-2.9.10-cp312-cp312-win_amd64.whl", hash = "sha256:4a579d6243da40a7b3182e0430493dbd55950c493d8c68f4eec0b302f6bbf20e"},
-    {file = "psycopg2-2.9.10-cp313-cp313-win_amd64.whl", hash = "sha256:91fd603a2155da8d0cfcdbf8ab24a2d54bca72795b90d2a3ed2b6da8d979dee2"},
```
seeker25, line 2419: weird

auth-api/src/auth_api/services/org.py
seeker25: alternatively make the error generic for both

`PAYMENT_ACCOUNT_FAILED`

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
-        else:
-            payment_account_status = PaymentAccountStatus.FAILED
+            if payment_method == PaymentMethod.PAD.value:  # PAD has bank related details
+                pay_request["paymentInfo"]["bankTransitNumber"] = payment_info.get("bankTransitNumber", None)
+                pay_request["paymentInfo"]["bankInstitutionNumber"] = payment_info.get("bankInstitutionNumber", None)
+                pay_request["paymentInfo"]["bankAccountNumber"] = payment_info.get("bankAccountNumber", None)
+                pay_request["padTosAcceptedBy"] = kwargs["user_context"].user_name
+            # invoke pay-api
+            token = RestService.get_service_account_token()
+            if is_new_org:
```
seeker25, line 263:
```
method = RestService.post if is_new_org else RestService.put
endpoint = f"{pay_url}/accounts" if is_new_org else f"{pay_url}/accounts/{org_model.id}"
response = method(endpoint=endpoint, data=pay_request, token=token, raise_for_status=True)
```

auth-api/src/auth_api/services/org.py
```diff
@@ -227,65 +228,84 @@ def _create_payment_settings(
...
+                payment_account_status = PaymentAccountStatus.FAILED
+                if response.json:
+                    error_payload = response.json()
+                    error_code = error_payload.get("error", "UNKNOWN_ERROR")
+                else:
+                    error_code = "UNKNOWN_ERROR"
+                logger.error(f"Account create payment Error: {response.text}")
+
+            if payment_account_status != PaymentAccountStatus.FAILED and payment_method:
+                payment_method_description = (
```
seeker25, line 288: So it's possible to be "failed" but still have your payment account updated and skip the activity log?

auth-api/src/auth_api/services/org.py
seeker25: I think in TEST, it's possible CFS account status isn't PENDING and it's ACTIVE right away if the holding date for pad is set to -1 on DEV/TEST but 3 on PROD

auth-api/src/auth_api/services/org.py
seeker25: I have a gut feeling this "failed" isn't correct

auth-api/src/auth_api/services/org.py
Jxio: 201 is from unit test, it has around 290 unit tests return 201 here.

Jxio: I removed it, I throw business exception here before so lots of unit tests failed, I think it can be removed now cause I changed it to logger.warning

auth-api/src/auth_api/exceptions/errors.py
```diff
@@ -113,6 +113,7 @@ class Error(Enum):
     )
     BCEID_USERS_CANT_BE_OWNERS = "BCEID Users cant be owners", HTTPStatus.BAD_REQUEST
     ACCOUNT_CREATION_FAILED_IN_PAY = "Account creation failed in Pay", HTTPStatus.INTERNAL_SERVER_ERROR
+    ACCOUNT_UPDATE_FAILED_IN_PAY = "Account update failed in Pay", HTTPStatus.INTERNAL_SERVER_ERROR
```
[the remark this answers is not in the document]
Jxio: changed to PAYMENT_ACCOUNT_UPSERT_FAILED

auth-api/src/auth_api/services/org.py
Jxio: I changed to warning, should be fine

auth-api/src/auth_api/services/org.py
Jxio: I'm not quite sure on this one, I thought it might be throw exception and tried, but too many unit tests failed

seeker25: I'm pretty sure even though it "FAILED" if the response was 200, it still was updated in sbc-pay - the payment method would still be changed then, you'll have to recreate the scenario and watch the activity log?
Unless it's possible it's because of the PAD confirmation period?

Anyways might not be worth looking into, if we can just keep the same flow as before.

auth-api/src/auth_api/services/org.py
```diff
@@ -250,42 +304,17 @@ def _create_payment_settings(
...
-            response = RestService.post(
-                endpoint=f"{pay_url}/accounts", data=pay_request, token=token, raise_for_status=True
-            )
-        else:
-            response = RestService.put(
-                endpoint=f"{pay_url}/accounts/{org_model.id}", data=pay_request, token=token, raise_for_status=True
+        if payment_method == PaymentMethod.PAD.value:  # PAD has bank-related details
+            pay_request["paymentInfo"].update(
+                {
+                    "bankTransitNumber": payment_info.get("bankTransitNumber"),
```
seeker25, line 310: Aye good change

auth-api/src/auth_api/services/org.py
```diff
@@ -426,7 +455,11 @@ def update_org(self, org_info):  # pylint: disable=too-many-locals, too-many-sta
                 Org.send_staff_review_account_reminder(relationship_id=self._model.id)
 
         if name_updated or payment_info:
-            Org._create_payment_for_org(mailing_address, self._model, payment_info, False)
+            payment_account_status, error = Org._create_payment_for_org(
+                mailing_address, self._model, payment_info, False
+            )
+            if payment_account_status == PaymentAccountStatus.FAILED and error is not None:
+                logger.warning(f"Account update payment Error: {error}")
```
seeker25, line 462: Looks good, we can just gun these if we start seeing them in DEV/TEST/PROD etc

auth-api/src/auth_api/services/org.py
```diff
@@ -227,7 +226,62 @@ def _create_payment_settings(
         **kwargs,
     ):
         """Add payment settings for the org."""
-        pay_url = current_app.config.get("PAY_API_URL")
+        try:
+            pay_url = current_app.config.get("PAY_API_URL")
+            pay_request = Org._build_payment_request(org_model, payment_info, payment_method, mailing_address, **kwargs)
```
seeker25, line 231: haha mega function!!

auth-api/src/auth_api/services/org.py
Jxio: yes I keep the same flow as before

seeker25 (approved):

seeker25:
Looks like some packages need to be fixed

auth-api/poetry.lock
Jxio: clear out poetry change

PR #3251, opened 2025-02-07
Remove staff group if ejected from staff org

seeker25:


ochiu (approved):

seeker25:
Will do this one later, need to focus on completing some of the already open tickets

PR #3253, opened 2025-02-10
Fix product for admin only

seeker25:

### DIFF
--- a/auth-api/migrations/versions/2025_02_10_7425171449c9_remove_products_coodinator.py
+++ b/auth-api/migrations/versions/2025_02_10_7425171449c9_remove_products_coodinator.py
@@ -1,0 +1,22 @@
+"""Remove products coodinator
+
+Revision ID: 7425171449c9
+Revises: 3f5b8a19ce42
+Create Date: 2025-02-10 09:53:25.604521
+
+"""
+from alembic import op
+import sqlalchemy as sa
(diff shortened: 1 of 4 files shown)


auth-web/src/views/auth/AccountSettings.vue
```diff
@@ -185,7 +185,7 @@
                 <v-list-item-title>Authentication</v-list-item-title>
               </v-list-item>
               <v-list-item
-                v-if="isUserMembership"
+                v-if="isAdmin"
```
seeker25, line 188: duplicate check, rather leave this though

ochiu (approved):

auth-api/migrations/versions/2025_02_10_7425171449c9_remove_products_coodinator.py
```diff
@@ -0,0 +1,22 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = '7425171449c9'
+down_revision = '3f5b8a19ce42'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+  op.execute("delete from permissions where membership_type_code = 'COORDINATOR' and actions = 'view_request_product_package'")
```
seeker25, line 19: Need this because AccountPaymentMethods checks for permissions

PR #3296, opened 2025-03-07
26272 use businessid from query string

TVWerdal:
*Issue #:*
https://github.com/bcgov/entity/issues/26272

If a business id is not found in session storage, we check for a value in the query string

### DIFF
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.8.16",
+  "version": "2.8.17",
   "appName": "Auth Web",
   "sbcName": "SBC Common Components",
   "private": true,
--- a/auth-web/src/stores/business.ts
+++ b/auth-web/src/stores/business.ts
@@ -254,7 +254,16 @@ export const useBusinessStore = defineStore('business', () => {
   }
 
   async function loadBusiness () {
-    const businessIdentifier = ConfigHelper.getFromSession(SessionStorageKeys.BusinessIdentifierKey)
+    let businessIdentifier = ConfigHelper.getFromSession(SessionStorageKeys.BusinessIdentifierKey)
+    // If we don't find anything in SessionStorage, try to grab it from the URL
+    if (businessIdentifier === null) {
+      const queryString = window.location.search
+      const urlParams = new URLSearchParams(queryString)
+      if (urlParams.get('businessid')) {
+        businessIdentifier = urlParams.get('businessid')
+      }
+    }
+
     // Need to look at LEAR, because it has the up-to-date names.
     const learBusiness = await searchBusiness(businessIdentifier)
     const response = await BusinessService.getBusiness(businessIdentifier)


bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-test-3296-gzf2o7s9.web.app

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-3296-3n35696z.web.app

TVWerdal:
Logged in as staff, you can use this URL's like this to test: 

https://bcregistry-account-dev--pr-3296-3n35696z.web.app/businessprofile?accountid=3040&businessid=BC0883782

auth-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.8.16",
+  "version": "2.8.17",
```
severinbeauvais, line 3: Please run `npm i` and commit the lock file, too.

auth-web/src/stores/business.ts
```diff
@@ -254,7 +254,16 @@ export const useBusinessStore = defineStore('business', () => {
...
-    const businessIdentifier = ConfigHelper.getFromSession(SessionStorageKeys.BusinessIdentifierKey)
+    let businessIdentifier = ConfigHelper.getFromSession(SessionStorageKeys.BusinessIdentifierKey)
+    // If we don't find anything in SessionStorage, try to grab it from the URL
+    if (businessIdentifier === null) {
+      const queryString = window.location.search
+      const urlParams = new URLSearchParams(queryString)
+      if (urlParams.get('businessid')) {
+        businessIdentifier = urlParams.get('businessid')
+      }
+    }
```
severinbeauvais, line 265: You can get the URL query parameters from the `$route` object.

Example: https://github.com/bcgov/business-filings-ui/blob/0955745de0fcb6c0fe16d5b99e84bd94347c3505/src/App.vue#L451

seeker25:
Please use 
image

<number> - <description>

we have an automatic script that tags zenhub releases

auth-web/src/stores/business.ts
severinbeauvais: ... Except that this code is in the store, which doesn't have access to $route.

severinbeauvais (approved):
A little hacky but I'm OK with this, since you tested that user authentication using KC is untouched by this, so it doesn't open any security holes.

Besides, we'll want to move this to Entities side at some point.

The only possibly improvement might be to only do this if we are on the "change profile" route. Thoughts?

JazzarKarim (approved):

severinbeauvais (approved):
Looks like the cleanest solution by far. What do you think of it? Did your test all flows?

severinbeauvais (commented):
Remember to update app version.

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-test-3296-gzf2o7s9.web.app

bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-3296-3n35696z.web.app

TVWerdal:
> https://bcregistry-account-dev--pr-3296-3n35696z.web.app

As staff: 
  https://bcregistry-account-dev--pr-3296-3n35696z.web.app/businessprofile?accountid=3040&businessid=BC0883782

TVWerdal:
Reworked to clean up the code and make it less hacky, based on Severin's advice/help

TVWerdal:
> Looks like the cleanest solution by far. What do you think of it? Did your test all flows?

Yes, it requires the least changes overall. I have tested reaching the page from the dashboard - I couldn't find my way to the businessprofile URL from within the app itself. This will also still fail if a use does not have a session businessid  and also does not get a query string, but I was unable to find a way that could happen.

seeker25:
Thanks Thayne, been running this script for auth https://github.com/bcgov/sbc-pay/blob/main/releases/release_to_zenhub_auth.py

TVWerdal:
> I'm going to ask again, could you please change the title so its `26272 - use businessid from query string`
> 
> thanks in advance

Was it the dash that I needed to add? Apologies I thought you meant the ticket number and I was confused because it was already there.

seeker25:
Yup should work:
https://github.com/bcgov/sbc-pay/blob/3703be9da75d8d60c9c256e49fd7a8f0ea311ee6/releases/util.py#L19

TVWerdal:
> Yup should work: https://github.com/bcgov/sbc-pay/blob/3703be9da75d8d60c9c256e49fd7a8f0ea311ee6/releases/util.py#L19

I'll keep that in mind for the future, thanks for the explanation.

PR #3327, opened 2025-03-25
26556 - Add in user / membership row on API key creation

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/26556

*Description of changes:*


auth-api/src/auth_api/services/api_gateway.py
```diff
@@ -74,20 +81,41 @@ def create_key(cls, org_id: int, request_json: Dict[str, str]):
             )
             response = api_key_response.json()
 
+        cls._create_user_and_membership_for_api_user(org_id, env)
```
bolyachevets, line 84: still not very clear about sandbox environment work. but looking at this method - create_key - do we need to call it twice: prod/sandbox, in which case do we update the membership row only the first time create_key is called? and then am i correct to assume that regardless of when the method is called (during prod or sandbox api key generation) the membership row is created in prod db (sandbox db will have a different mechanism to sync this membership)

seeker25: Yes, it needs to be called twice, once for PROD (PROD will fire over the ORG to SANDBOX) then once for SANDBOX

They have two separate clients in keycloak, thus two separate service accounts.

seeker25:
Unit tests fixed in https://github.com/bcgov/sbc-auth/pull/3330

seeker25:
Will fix code duplicate tomorrow, still migration WIP

PR #3329, opened 2025-03-25
26547 - feature remove minio from auth api

bolyachevets:
*Issue #:*
https://github.com/bcgov/entity/issues/entity/26547

*Description of changes:*
Related service account permissions changes: https://github.com/bcgov/bcregistry-sre/commit/5c1d3004756d9ff66111054131796316e3571020

some more changes were needed for credentials to work both locally (with a service account's token) and inside cloudrun (without a service account token):
- Local Development: Works with GOOGLE_APPLICATION_CREDENTIALS service account keys
- Cloud Run: Automatically uses workload identity (no need for service account key with Service Account Token Creator role assigned)
- Unit Tests: Supports mocking via google.auth.default() patching

the front end changes are needed because headers for signed url need to match headers in the frontend request.

also, cors policy needs to be set for the bucket. (added cors policies to the devops/ folder)

1password will also need to be update for auth-web's VUE_APP_FILE_SERVER_URL to point to gcp static resource bucket.


pwei1018 (approved):

auth-api/tests/conftest.py
```diff
@@ -73,16 +82,48 @@ def client_ctx(app):  # pylint: disable=redefined-outer-name
...
-    """Return a session-wide initialised database."""
+    """Return a session-wide initialised database.
+
+    Drops all existing tables - Meta follows Postgres FKs
+    """
     with app.app_context():
-        if database_exists(_db.engine.url):
-            drop_database(_db.engine.url)
-        create_database(_db.engine.url)
-        _db.session().execute(text('SET TIME ZONE "UTC";'))
```
seeker25, line 83: please restore this

auth-api/tests/conftest.py
```diff
@@ -73,16 +82,48 @@ def client_ctx(app):  # pylint: disable=redefined-outer-name
         yield _client
 
 
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="session")
 def db(app):  # pylint: disable=redefined-outer-name, invalid-name
-    """Return a session-wide initialised database."""
+    """Return a session-wide initialised database.
```
seeker25, line 87: Why is this being changed?

auth-api/tests/docker/docker-cmpose-mac.yml
```diff
@@ -1,81 +0,0 @@
-version: "3"
-
-services:
-  keycloak:
```
seeker25, line 4: Jia would use this because the other docker compose wasn't working on her computer, should be ok though because I have a newer version in my PR

auth-api/src/auth_api/services/google_store.py
```diff
@@ -0,0 +1,89 @@
...
+from flask import current_app
+from google.cloud import storage
+from structured_logging import StructuredLogging
+
+from auth_api.utils.constants import AFFIDAVIT_FOLDER_NAME
+
+logger = StructuredLogging.get_logger()
+
+
+class GoogleStoreService:
```
seeker25, line 28: Isn't this duplicated here? queue_services/account-mailer/src/account_mailer/services/google_store.py

bolyachevets: potentially, but they are using different methods, auth api is only doing signed get/put urls, while mailer is only doing donwload (and upload is used in tests). but then we might as well put it in the common components at some point

auth-api/tests/conftest.py
bolyachevets: was done the other way in mailer. will check again

seeker25: unless there's a specific reason to do this, please revert

seeker25 (approved):

auth-api/tests/docker/docker-cmpose-mac.yml
seeker25: @Jxio Does the docker container in my PR work for you? https://github.com/bcgov/sbc-auth/pull/3327

auth-api/tests/conftest.py
bolyachevets: i added postgres to docker compose, it looks like running it that way does not like 'drop database' operation -> E           pg8000.exceptions.DatabaseError: {'S': 'ERROR', 'V': 'ERROR', 'C': '55006', 'M': 'cannot drop the currently open database', 'F': 'dbcommands.c', 'L': '1586', 'R': 'dropdb'} ---- hence the switch to drop schema from (which was used in the mailer). I found having db in the docker compose convenient for running unit tests, potentially i can revert it.

auth-api/tests/docker/docker-cmpose-mac.yml
bolyachevets: fyi, the remaining docker compose from this PR worked on my mac

auth-api/tests/conftest.py
seeker25: it's a common mistake, don't use a postgres named database

auth-api/tests/docker/docker-cmpose-mac.yml
Jxio: yes, I tested, it working

auth-api/tests/conftest.py
```diff
@@ -26,6 +28,14 @@
 from auth_api.utils.auth import jwt as _jwt
 
 
+def find_subpath(root_dir, target_subpath):
```
seeker25, line 31: still need?

seeker25 (approved):

PR #3330, opened 2025-03-26
Wrapping test objects in dict() because they cross polute when modifi…

seeker25:
…ed across test.


auth-api/tests/unit/services/test_invitation_auth.py
```diff
@@ -219,14 +219,14 @@ def test_change_authentication_non_govm(session, auth_mock, keycloak_mock, monke
             assert invitation_model.login_source == LoginSource.BCSC.value
             assert invitation_model.invitation_status_code == InvitationStatus.ACCEPTED.value
 
-            patch_token_info(TestJwtClaims.tester_role, monkeypatch)
```
seeker25, line 222: Wrong keycloak SUB

PR #3351, opened 2025-04-01
22113 - Removed logic that stops conditionally-approved NRs from being affiliated

JazzarKarim:
*Issue #:*
https://github.com/bcgov/entity/issues/

*Description of changes:*
Removed the logic that prevents conditionally-approved NRs from being affiliated

BEFORE:
before conditional nr

AFTER:
after conditional nr

### DIFF
--- a/auth-api/src/auth_api/services/affiliation.py
+++ b/auth-api/src/auth_api/services/affiliation.py
@@ -279,10 +279,6 @@ class Affiliation:
             if not (invoices and invoices["invoices"] and invoices["invoices"][0].get("statusCode") == "COMPLETED"):
                 raise BusinessException(Error.NR_NOT_PAID, None)
 
-        # If consentFlag is not R, N or Null for a CONDITIONAL NR throw error
-        if status == NRStatus.CONDITIONAL.value and nr_json.get("consentFlag", None) not in (None, "R", "N"):
-            raise BusinessException(Error.NR_NOT_APPROVED, None)
-
         if not user_is_staff and (
             (phone and re.sub(r"\D", "", phone) != re.sub(r"\D", "", nr_phone))
             or (email and email.casefold() != nr_email.casefold())


auth-api/src/auth_api/services/affiliation.py
```diff
@@ -279,10 +279,6 @@ def create_new_business_affiliation(  # pylint: disable=too-many-locals
             if not (invoices and invoices["invoices"] and invoices["invoices"][0].get("statusCode") == "COMPLETED"):
                 raise BusinessException(Error.NR_NOT_PAID, None)
 
-        # If consentFlag is not R, N or Null for a CONDITIONAL NR throw error
-        if status == NRStatus.CONDITIONAL.value and nr_json.get("consentFlag", None) not in (None, "R", "N"):
-            raise BusinessException(Error.NR_NOT_APPROVED, None)
-
```
severinbeauvais, line 285: So removing this allows affiliation.

Does the BRD have a check to disallow using the NR if it's still conditional (depending on consent flag)?

JazzarKarim: Yes, here's what it does. The action will say "Open Name Request":
image


If a user clicks on that, the BRD will direct to Name Request UI where there is no action there. Name Request UI will be the decisive here:
image

severinbeauvais: Thanks.

Some conditional NRs can be used, depending on the consent flag (eg, R ot N). Does the BRD handle that?

JazzarKarim: Yup! I added this not so long ago:
image

Basically when the consent is received or waived, the BRD will treat it as a regular Approved NR that can be used.

Reference: https://github.com/bcgov/business-ui/pull/68

seeker25 (approved):

auth-api/src/auth_api/services/affiliation.py
severinbeauvais: OK, perfect. Resolving.

severinbeauvais (approved):
LGTM. And nice test, too!

Is there an API version that should be updated?

JazzarKarim:
> LGTM. And nice test, too!
> 
> Is there an API version that should be updated?

Thanks!

Looking at previous Auth API PRs, no update for the API version is needed 👍

PR #3358, opened 2025-04-08
Auto approval for SANDBOX so BCEID don't need staff approval

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/27041


ochiu (approved):

seeker25:
I will continue more on this if it's needed. If it isn't we'll remove later

PR #3361, opened 2025-04-10
26087 - Auth API: support affiliations search parameters

Rajandeep98:
*Issue #:*26087
https://github.com/bcgov/entity/issues/#26087
*Description of changes:*
The affiliations endpoint needs to support parameters:
business name partial match
number partial match
type
status
paging

### DIFF
--- a/auth-api/src/auth_api/resources/v1/org.py
+++ b/auth-api/src/auth_api/resources/v1/org.py
@@ -352,7 +352,12 @@ def get_organization_affiliations(org_id):
     """Get all affiliated entities for the given org."""
     try:
         env = get_request_environment()
-        # keep old response until UI is updated
+        search_filter_status = request.args.get("status", None)
+        search_filter_name = request.args.get("name", None)
+        search_filter_type = request.args.get("type", None)
+        search_filter_identifier = request.args.get("identifier", None)
+        page = int(request.args.get("page", 1))
+        limit = int(request.args.get("limit", 50))
         if (request.args.get("new", "false")).lower() != "true":
             return (
                 jsonify({"entities": AffiliationService.find_visible_affiliations_by_org_id(org_id, env)}),
@@ -362,10 +367,56 @@ def get_organization_affiliations(org_id):
         # get affiliation identifiers and the urls for the source data
         affiliations = AffiliationModel.find_affiliations_by_org_id(org_id, env)
         affiliations_details_list = asyncio.run(AffiliationService.get_affiliation_details(affiliations, org_id))
+        start_index = (page - 1) * limit
+        end_index = start_index + limit
+        filtered_affiliations = affiliations_details_list
+        conditions = []
+        if search_filter_status:
+            conditions.append(
+                lambda affiliation: (
+                    (affiliation.get('state', '').upper()) == (search_filter_status.upper())
+                    or (affiliation.get('draftStatus', '').upper()) == (search_filter_status.upper())
+                )
+            )
+        if search_filter_name:
+            # Ensure the status filter
+            conditions.append(lambda affiliation: affiliation.get('legalName')
+                              and affiliation.get('legalName').casefold().startswith(search_filter_name.casefold()))
+        if search_filter_type:
+            # Ensure the legaltype filter
+            conditions.append(
+                lambda affiliation: (
+                    (affiliation.get('draftType', '').casefold() == search_filter_type.casefold())
+                    or (not affiliation.get('draftType') and affiliation.get('legalType')
+                        and affiliation.get('legalType').casefold() == search_filter_type.casefold())
+                )
+            )
+        if search_filter_identifier:  # Check if the search_filter_identifier is provided
+            normalized_identifier = search_filter_identifier.replace(" ", "").lower()
+            # Check if the search_filter_identifier starts with 'nr'
+            if normalized_identifier.startswith('n'):
+                print(normalized_identifier)
+                # Filter by nrNum  or nrNumber(case-insensitive)
+                conditions.append(lambda affiliation: (
+                    (affiliation.get('nrNumber')
+                        and affiliation.get('nrNumber').replace(" ", "").lower().startswith(normalized_identifier))
+                ))
+            else:
+                # Filter by identifier (case-insensitive)
+                conditions.append(lambda affiliation: affiliation.get('identifier')
+                                  and affiliation.get('identifier').strip().lower().startswith(normalized_identifier))
+
+        filtered_affiliations = [
+            affiliation for affiliation in filtered_affiliations
+            if all(condition(affiliation) for condition in conditions)
+        ]
+        total_results = len(filtered_affiliations)
+        paginated_affiliations = filtered_affiliations[start_index:end_index]
         # Use orjson serializer here, it's quite a bit faster.
         response, status = (
             current_app.response_class(
-                response=orjson.dumps({"entities": affiliations_details_list}),  # pylint: disable=maybe-no-member
+                response=orjson.dumps({"entities": paginated_affiliations,
+                                       "totalResults": total_results}),  # pylint: disable=maybe-no-member
                 status=200,
                 mimetype="application/json",
             ),
@@ -376,7 +427,6 @@ def get_organization_affiliations(org_id):
         response, status = {"code": exception.code, "message": exception.message}, exception.status_code
     except ServiceUnavailableException as exception:
         response, status = {"message": exception.error}, exception.status_code
-
     return response, status
 
 


auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -352,7 +352,12 @@ def get_organization_affiliations(org_id):
     """Get all affiliated entities for the given org."""
     try:
         env = get_request_environment()
-        # keep old response until UI is updated
+        search_filter_status = request.args.get("status", None)
+        search_filter_name = request.args.get("name", None)
+        search_filter_type = request.args.get("type", None)
+        search_filter_identifier = request.args.get("identifier", None)
+        page = int(request.args.get("page", 1))
+        limit = int(request.args.get("limit", 50))
```
severinbeauvais, line 360: As per UI, default should be 100.

severinbeauvais:
@seeker25 This is an interim step before we rewrite the query across the 3 dbs.

seeker25 (changes requested):
needs alot of changes before being merged. 

1. The filtering needs to be passed to namex/lear not auth
Otherwise you're going to be sitting there waiting for Names and LEAR to return forever and then return 10/100 rows or whichever after waiting for 1m for them to return
2. You have no associations between NR's and businesses, so you don't know which affiliations should be batched together when you're fetching
3. This shouldn't be done in the resource, it should be done in the service
4. Use a DTO (think dataclass) or something else instead of 
```
        search_filter_status = request.args.get("status", None)
        search_filter_name = request.args.get("name", None)
        search_filter_type = request.args.get("type", None)
        search_filter_identifier = request.args.get("identifier", None)
        page = int(request.args.get("page", 1))
        limit = int(request.args.get("limit", 50))
```

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -352,7 +352,12 @@ def get_organization_affiliations(org_id):
     """Get all affiliated entities for the given org."""
     try:
         env = get_request_environment()
-        # keep old response until UI is updated
+        search_filter_status = request.args.get("status", None)
```
severinbeauvais, line 355: Does this code support multiple statuses and multiple types?

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -362,10 +367,56 @@ def get_organization_affiliations(org_id):
...
+            )
+        if search_filter_identifier:  # Check if the search_filter_identifier is provided
+            normalized_identifier = search_filter_identifier.replace(" ", "").lower()
+            # Check if the search_filter_identifier starts with 'nr'
+            if normalized_identifier.startswith('n'):
+                print(normalized_identifier)
+                # Filter by nrNum  or nrNumber(case-insensitive)
+                conditions.append(lambda affiliation: (
+                    (affiliation.get('nrNumber')
+                        and affiliation.get('nrNumber').replace(" ", "").lower().startswith(normalized_identifier))
```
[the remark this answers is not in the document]
seeker25: doesn't matter this shouldn't be done in auth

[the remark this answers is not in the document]
severinbeauvais: By the same argument, this shouldn't be done in Names and it shouldn't be done in LEAR. To be clear, the query needs data from all 3 databases.

The idea here was to make a bit of progress with a small gain -- with a smaller list of affiliations (ie, a page), the UI won't choke as it does now. Also, this code implements the acceptance of URL parameters.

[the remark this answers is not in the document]
seeker25: LOL I don't think you understood my response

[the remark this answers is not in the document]
severinbeauvais: Sorry.

It's preparation for the future (passing filters and page params) and helps the UI.

We don't have a solution the BE yet. Thoughts?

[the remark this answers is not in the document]
seeker25: Almost think we should move the entire affiliation table to LEAR
Even with the names affiliations 

Think about this for the future.. You'll probably want some hierarchy of users that have an affiliation

Does that make sense to keep putting in auth? 

Also Names always eventually become a business? if they don't, just leave the data in there?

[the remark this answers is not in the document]
severinbeauvais: Are you saying that LEAR should have the cross-reference table for which entities/NRs belong to which accounts? Where does the data for a new affiliation come from?

Where is the current data for which accounts can access businesses?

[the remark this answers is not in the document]
seeker25: 1. Yes 
2. It's formed when a business is created, entity-filer calls into AUTH to create an affiliation 
Same idea with NRO, it's from a queue message/call when an affiliation is formed to an org

3. Affiliation <-> Entity table in auth - You wouldn't need the entity table if you had this in LEAR.. you'd just associate it to the `businesses/filings` table instead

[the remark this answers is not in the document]
seeker25: You wouldn't have to worry about data going out of sync between `auth-db.entities` and `lear.businesses` or `lear.filings`

it would be straight forward you could query which orgs have access to a business without looking in auth-db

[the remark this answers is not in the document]
severinbeauvais: How would an account affiliate an existing business or NR? Would it still be a dialog on the Business Registry Dashboard, which would now need to call Auth to validate and then LEAR to save the affiliation?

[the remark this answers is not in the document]
seeker25: Yes, until we get the org membership in the JWT - then the call to AUTH would go away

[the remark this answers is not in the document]
severinbeauvais: While this makes sense, it's a lot of work (including migration of data).

Is there anything we can do before we redesign this? Like this ticket? We believe we have accounts in Prod with 1000s of affiliations, so we need to do something before MVP in June.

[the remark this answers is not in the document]
seeker25: If we build the parameters into Names and LEAR that would be a good start

https://github.com/bcgov/namex/blob/f20f060aea4f4b173867605f01a3b27096d09968/api/namex/resources/requests.py#L379

https://github.com/bcgov/lear/blob/764ea672dd43af6f367cf9deb44018a46b5247e2/legal-api/src/legal_api/resources/v2/business/business.py#L165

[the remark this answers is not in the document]
seeker25: wouldn't be very difficult this could easily restore search functionality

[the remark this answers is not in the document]
seeker25: Also when LEAR creates an affiliation, it needs to kick over a subfield like an NR that is being consumed while creating the affiliation row.. so we can enable pagination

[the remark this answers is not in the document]
seeker25: Yes, way less data

"If a cross-db query is still the preferred solution, it it worth passing parameters to Names/LEAR?"

I think it's a good stop gap and wouldn't be hard to implement

"Don't you see any value in the filtering and pagination work in this PR?"

No because they'll search and cause 3 minute calls to the database potentially multiple times and cause lots of issues.
At least right now they have to load the business dashboard, if you added this search without filtering down the rows you'd be multiplying the length.

seeker25:
Talking with Rajandeep:

If you did this (as per Kials suggestion of 1 single search):

image

Really easy to add in for LEAR:

```python
search = json_input.get('search', None)
if search: 
   filter(Business.name.ilike(search)) or Business._identifier.ilike(search)
   Filing.temp_reg.ilike(search)
```   

   
Really easy to add in for Names:
```python
search_string = search.get('search', None)
if search_string:
  .filter(Name.name.ilike(search_string) or Name.nrNum.ilike(search_string)
```

severinbeauvais:
> 5. How are these changes any different than just simply filtering on the frontend? I guess you get a smaller payload returned? But it would still take almost the same amount of time?

It allows us to deprecate the internal filters in the UI (and show pagination options).
It is a smaller payload returned (esp since the UI works poorly with more than 1000 or 2000 records).
It prepares this endpoint to receive and process the filters/pagination, even though we know the underlying code will change.

auth-api/src/auth_api/resources/v1/org.py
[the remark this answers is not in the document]
severinbeauvais: > "If a cross-db query is still the preferred solution, it it worth passing parameters to Names/LEAR?"
> I think it's a good stop gap and wouldn't be hard to implement
❤️ 

> "Don't you see any value in the filtering and pagination work in this PR?"
> No because they'll search and cause 3 minute calls to the database ...

OK. I'm resolving this conversation as you've laid things out thoroughly in your comment below. Thanks! Your guidance is much appreciated.

severinbeauvais:
@argush3 FYI. I suggest reading the comments and ignoring the code for now. We need to agree on what to do.

severinbeauvais:
This is not the right solution, but this PR contains lots of good suggestions and example.

I think Rajan should keep using this PR for the Auth API work as per the updated ticket description. (No need for a second PR.)

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -482,7 +482,7 @@ def sort_key(item):
             return combined
         except ServiceUnavailableException as err:
             logger.debug(err)
-            logger.debug("Failed to get affiliations details: %s", affiliations)
+            logger.debug("Failed to get affiliations details: %s")
```
severinbeauvais, line 485: What does `%s` refer to?

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -457,7 +457,7 @@ async def get_affiliation_details(affiliations: List[AffiliationModel], org_id)
             )
 
         call_info = [
-            {"url": url, "payload": {"identifiers": identifiers}} for url, identifiers in url_identifiers.items()
+            {"url": url, "payload": {"identifiers": identifiers, "state":search_filter_status, "name":search_filter_name,"type":search_filter_type}} for url, identifiers in url_identifiers.items()
```
seeker25, line 460: linting probably going to fail?

`poetry run black .`

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -457,7 +460,18 @@ async def get_affiliation_details(affiliations: List[AffiliationModel], org_id)
             )
 
         call_info = [
-            {"url": url, "payload": {"identifiers": identifiers}} for url, identifiers in url_identifiers.items()
+            {
+                "url": url,
+                "payload": {
+                    "identifiers": identifiers,
```
seeker25, line 466: could probably just serialize your data class here

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -352,20 +353,30 @@ def get_organization_affiliations(org_id):
...
-
         # get affiliation identifiers and the urls for the source data
         affiliations = AffiliationModel.find_affiliations_by_org_id(org_id, env)
-        affiliations_details_list = asyncio.run(AffiliationService.get_affiliation_details(affiliations, org_id))
+        search_details = AffiliationInvitationSearchDeatils()
+        search_details.search_filter_status = request.args.getlist("status")
+        search_details.search_filter_name = request.args.get("name")
+        search_details.search_filter_type = request.args.getlist("type")
+        search_details.page = int(page)
+        search_details.limit = int(limit)
```
seeker25, line 370: might be a better idea to just move this into a method in the dataclass

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -352,20 +353,30 @@ def get_organization_affiliations(org_id):
     """Get all affiliated entities for the given org."""
     try:
         env = get_request_environment()
```
seeker25, line 355: Will have to resync, I've removed env

auth-api/src/auth_api/models/dataclass.py
```diff
@@ -47,6 +47,17 @@ class AffiliationInvitationSearch:  # pylint: disable=too-many-instance-attribut
...
 
 
+@dataclass
+class AffiliationInvitationSearchDeatils:  # pylint: disable=too-many-instance-attributes
+    """Used for filtering Affiliation Invitations based on filters passed."""
+
+    search_filter_status: Optional[str] = None
+    search_filter_name: Optional[str] = None
+    search_filter_type: Optional[str] = None
+    page: int = 1
```
seeker25, line 57: why not just status
name
type

auth-api/src/auth_api/models/dataclass.py
```diff
@@ -47,6 +47,17 @@ class AffiliationInvitationSearch:  # pylint: disable=too-many-instance-attribut
     is_deleted: bool = False
 
 
+@dataclass
+class AffiliationInvitationSearchDeatils:  # pylint: disable=too-many-instance-attributes
```
seeker25, line 51: typo

auth-api/src/auth_api/models/dataclass.py
```diff
@@ -47,6 +49,29 @@ class AffiliationInvitationSearch:  # pylint: disable=too-many-instance-attribut
...
+
+    identifier: Optional[str] = None
+    status: Optional[str] = None
+    name: Optional[str] = None
+    type: Optional[str] = None
+    page: int = 1
+    limit: int = 100
+
+    @classmethod
+    def from_request_args(cls, req: Request) -> "AffiliationInvitationSearchDetails":
```
seeker25, line 64: self? there's a way to annotate own class type

auth-api/src/auth_api/models/dataclass.py
```diff
@@ -47,6 +49,29 @@ class AffiliationInvitationSearch:  # pylint: disable=too-many-instance-attribut
...
+        return cls(
```
seeker25, line 65: much better, we don't really care about this too much in the resource

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -356,14 +357,19 @@ def get_organization_affiliations(org_id):
                 jsonify({"entities": AffiliationService.find_visible_affiliations_by_org_id(org_id)}),
                 HTTPStatus.OK,
             )
-
         # get affiliation identifiers and the urls for the source data
         affiliations = AffiliationModel.find_affiliations_by_org_id(org_id)
-        affiliations_details_list = asyncio.run(AffiliationService.get_affiliation_details(affiliations, org_id))
+        search_details = AffiliationInvitationSearchDetails.from_request_args(request)
```
seeker25, line 362: way cleaner

auth-api/src/auth_api/services/affiliation.py
seeker25: Let me see if I can find an example 

https://github.com/bcgov/sbc-pay/blob/04137958608dbde9fd70749760578ddbaf9052a0/pay-api/src/pay_api/utils/serializable.py#L14

But that's alot to bring over, I think what you have now is fine

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -453,19 +455,26 @@ def _affiliation_details_url(affiliation: AffiliationModel) -> str:
...
                 affiliation.entity.business_identifier
             )
-
         call_info = [
-            {"url": url, "payload": {"identifiers": identifiers}} for url, identifiers in url_identifiers.items()
+            {
+                "url": url,
+                "payload": {
+                    "identifiers": identifiers,
+                    **asdict(search_details),
```
seeker25, line 473: looks good

auth-api/src/auth_api/services/affiliation.py
Rajandeep98: just formating

auth-api/src/auth_api/models/dataclass.py
```diff
@@ -47,6 +49,29 @@ class AffiliationInvitationSearch:  # pylint: disable=too-many-instance-attribut
...
+
+    @classmethod
+    def from_request_args(cls, req: Request) -> Self:
+        return cls(
+            identifier=req.args.get("identifier"),
+            status=req.args.getlist("status") or [],
+            name=req.args.get("name"),
+            type=req.args.getlist("type") or [],
+            page=int(req.args.get("page", 1)),
+            limit=int(req.args.get("limit", 100)),
```
severinbeauvais, line 71: Since the fallback values are provided here, can they be removed the the dataclass above?

severinbeauvais (commented):
If I understand correctly, this is asking Namex API for "limit" results and is asking Legal API for "limit" results. Therefore, if limit is 100, there will be up to 200 results to combine and return, not the "up to 100" that was asked for.

Any thoughts on how to solve this? Or just leave it for now?

seeker25:
I asked Rajandeep the same thing, she is still working on it

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -465,12 +465,15 @@ async def get_affiliation_details(
...
+        search_dict = asdict(search_details)
+        has_filters = any(v not in (None, "", [], {}) for k, v in search_dict.items() if k not in {"page", "limit"})
+
         call_info = [
             {
                 "url": url,
                 "payload": {
-                    "identifiers": identifiers,
-                    **asdict(search_details),
+                    "identifiers": identifiers if has_filters else identifiers[:100],
```
severinbeauvais, line 475: Is the "100" here the default number of results if no page limits are provided?

This would break currently functionality in Dev, right?

Rajandeep98:
image
after reducing initial load to random 100 identifiers
existing payload 
image

Rajandeep98:
image

auth-api/src/auth_api/models/dataclass.py
severinbeauvais: The initial values on lines 56-59 have been updated. Should lines 66-69 be updated similarly?

(I'm really wondering why there's 2 sets of initial/default values and why they don't match.)

severinbeauvais:
> after reducing initial load to random 100 identifiers existing payload ...

What I see in these screenshots is that the second one took twice as long as the first one. And also, the second one returns a ton more data than the first one. Is this expected?

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -453,17 +455,31 @@ def _affiliation_details_url(affiliation: AffiliationModel) -> str:
...
+    async def get_affiliation_details(
+        affiliations: List[AffiliationModel], search_details: AffiliationInvitationSearchDetails, org_id
+    ) -> List:
         """Return affiliation details by calling the source api."""
         url_identifiers = {}  # i.e. turns into { url: [identifiers...] }
+        search_dict = asdict(search_details)
+        total = 0
+        has_filters = any(
+            value not in (None, "", [], {}) for field, value in search_dict.items() if field not in {"page", "limit"}
+        )
```
seeker25, line 467: build this into the dataclass? I bet there's a better way to do this

auth-api/src/auth_api/services/affiliation.py
```diff
@@ -453,17 +455,32 @@ def _affiliation_details_url(affiliation: AffiliationModel) -> str:
...
         for affiliation in affiliations:
             url = Affiliation._affiliation_details_url(affiliation)
-            url_identifiers.setdefault(url, [affiliation.entity.business_identifier]).append(
-                affiliation.entity.business_identifier
+            not has_filters and total >= 100 or (
```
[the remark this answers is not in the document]
seeker25: pagination is to be done in another PR

Rajandeep98:
full payload , no params
image
with diff aprams:
image
image
image

image
image

severinbeauvais:
You did a test above for the first 100 results, and then next 100.

I wonder why the second call is so  much faster than the first one...?

auth-api/src/auth_api/services/affiliation.py
Rajandeep98: keeping limit temp basis to 100000 till next pr for grouping of affiliations and pagination doesnt complete

seeker25 (approved):
Just the changes we discussed and I'll merge

severinbeauvais (approved):

seeker25:
I'm fixing auth-queue in another PR

PR #3370, opened 2025-04-28
Remove idp_userid and use sub instead

seeker25:


seeker25:
1. Checking the prod auth database before merging
2. Checking with OPS to make sure they don't delete out of keycloak

PR #3375, opened 2025-05-01
Remove depreciated

seeker25:


auth-api/tests/unit/api/test_org.py
```diff
@@ -1695,52 +1695,6 @@ def test_get_affiliation_without_authrized(client, jwt, session, keycloak_mock):
     assert rv.status_code == HTTPStatus.UNAUTHORIZED
 
 
-def test_get_affiliations(client, jwt, session, keycloak_mock):  # pylint:disable=unused-argument
```
seeker25, line 1698: we have a route that tests this with ?new=true

ochiu (approved):

auth-api/src/auth_api/resources/v1/org.py
```diff
@@ -350,13 +350,6 @@ def delete_organzization_contact(org_id):
 def get_organization_affiliations(org_id):
     """Get all affiliated entities for the given org."""
     try:
-        # keep old response until UI is updated
-        if (request.args.get("new", "false")).lower() != "true":
-            return (
-                jsonify({"entities": AffiliationService.find_visible_affiliations_by_org_id(org_id)}),
-                HTTPStatus.OK,
-            )
-
```
seeker25, line 359: needs to be reverted

PR #3377, opened 2025-05-01
27068 - updated affiliation_invitation_verification

hanlunBCRegistries:
*Issue: https://github.com/bcgov/entity/issues/27068*

*Description of changes:*

This PR addresses issue #27068 by modifying the authorization logic for sending affiliation invitations. Previously, only users with `ADMIN` or `COORDINATOR` roles within the target organization could perform this action.

This change allows users with the `USER` role within the target organization to also send affiliation invitations.

**Changes Made:**

1.  Added a new function `is_user_admin_or_coordinator_or_user` to the `UserService` (`auth-api/src/auth_api/services/user.py`). This function checks if a user has an `ADMIN`, `COORDINATOR`, or `USER` role.
2.  Updated the `_verify_permissions` helper function in the affiliation invitation resource (`auth-api/src/auth_api/resources/v1/affiliation_invitation.py`) to utilize the new `UserService.is_user_admin_or_coordinator_or_user` function for its permission check.


auth-api/src/auth_api/services/user.py
```diff
@@ -698,3 +698,18 @@ def is_user_admin_or_coordinator(user, org_id: int) -> bool:
             return False
 
         return current_user_membership.membership_type_code in CLIENT_ADMIN_ROLES
+
+    @staticmethod
+    def is_user_admin_or_coordinator_or_user(user, org_id: int) -> bool:
+        """Check if user provided is admin, coordinator, or user for the given org id."""
+        current_user_membership: MembershipModel = MembershipModel.find_membership_by_user_and_org(
```
seeker25, line 705: no need for typehint, it should be inferred.. if it's not we need to put it on the definition of the function

auth-api/poetry.lock
```diff
@@ -1,101 +1,104 @@
-# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
```
seeker25, line 1: Please revert this file,you don't need to update it

auth-api/src/auth_api/resources/v1/affiliation_invitation.py
```diff
@@ -232,5 +232,5 @@ def _verify_permissions(user, affiliation_invitation_id):
         raise BusinessException(Error.DATA_NOT_FOUND, None)
 
     to_org_id = affiliation_invitation.as_dict()["to_org"]["id"]
-    if not UserService.is_user_admin_or_coordinator(user=user, org_id=to_org_id):
+    if not UserService.is_user_admin_or_coordinator_or_user(user=user, org_id=to_org_id):
```
seeker25, line 235: only other spot I saw this was for delegation, does delegation need to change as well?

hanlunBCRegistries: Updated, was trying to follow the implementation from this function: https://github.com/bcgov/sbc-auth/blob/main/auth-api/src/auth_api/services/user.py#L690

auth-api/src/auth_api/services/user.py
hanlunBCRegistries: Updated, was trying to follow the implementation from this function: https://github.com/bcgov/sbc-auth/blob/main/auth-api/src/auth_api/services/user.py#L690

auth-api/poetry.lock
hanlunBCRegistries: Reverted

auth-api/src/auth_api/resources/v1/affiliation_invitation.py
hanlunBCRegistries: Will check with Sandra about it, looks like the ticket only addresses the issue with affiliation invitation

seeker25 (approved):
Awesome thanks Hanlun, carrying us to victory again!

PR #3382, opened 2025-05-07
26898 - Update to how business name is obtained for firms

TVWerdal:
*Issue #:*
https://github.com/bcgov/entity/issues/26898 

For Firms we now populate business_name from alternateNames, as legalName is often populated with org name for firms.

### DIFF
--- a/auth-api/src/auth_api/services/affiliation_invitation.py
+++ b/auth-api/src/auth_api/services/affiliation_invitation.py
@@ -314,10 +314,7 @@ class AffiliationInvitation:
             alternative_names = business["business"].get("alternateNames")
             if alternative_names is not None:
                 for alt_name in alternative_names:
-                    if (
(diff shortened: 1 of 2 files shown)


auth-api/src/auth_api/services/affiliation_invitation.py
```diff
@@ -306,10 +306,13 @@ def create_affiliation_invitation(
         affiliation_invitation.token = confirmation_token
         affiliation_invitation.login_source = mandatory_login_source
         affiliation_invitation.save()
-
+        if business["business"]["legalType"] not in ["SP", "GP"]:
+            business_name = business["business"]["legalName"]
+        else:
+            business_name = business["business"]["alternateNames"]["name"]
```
seeker25, line 312: I think this is an array? and there's some logic to grab the correct one?

image

seeker25 (approved):

PR #3414, opened 2025-05-22
28438 - Update Affiliation to display proper title

hanlunBCRegistries:
*Issue #:*
https://github.com/bcgov/entity/issues/28438

### Description of Changes

Modified the `enrich_affiliation_invitations_dict_list_with_business_data` method in `auth-api/src/auth_api/services/affiliation_invitation.py`.

When processing business details for an affiliation invitation, if the entity's `legalType` is "SP" or "GP", the code now calls `get_business_name_from_alternative_name`. This ensures that an `alternateName` (operating name), if available, is prioritized over the `legalName` for display.

**Updated the `get_business_name_from_alternative_name` method:**
- It now correctly handles different structures of the input `business_data`
- It can accept either the full business details object from the Legal API (which has a nested `business` key) or just the nested business information directly
- This makes it reusable in contexts where the data structure might vary (e.g., during invitation creation vs. enrichment of existing invitation data)
- The core logic of checking `alternateNames` using the `business_identifier` remains the same

### DIFF
--- a/auth-api/src/auth_api/services/affiliation_invitation.py
+++ b/auth-api/src/auth_api/services/affiliation_invitation.py
@@ -115,40 +115,51 @@ class AffiliationInvitation:
                     AffiliationInvitationData.OrgDetails, affiliation_invitation_dict["from_org"]
                 )
             )
-            if to_org := affiliation_invitation_dict.get("to_org"):
+            if to_org_dict := affiliation_invitation_dict.get("to_org"):
                 to_org = AffiliationInvitationData.OrgDetails(
                     **_init_dict_for_dataclass_from_dict(
-                        AffiliationInvitationData.OrgDetails, affiliation_invitation_dict["to_org"]
+                        AffiliationInvitationData.OrgDetails, to_org_dict
                     )
                 )
+            else:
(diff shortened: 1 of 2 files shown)


hanlunBCRegistries:
> unit test?

Just updated unit tests.

auth-api/src/auth_api/services/affiliation_invitation.py
```diff
@@ -352,16 +359,15 @@ def _get_multiple_business_details(business_identifiers: List[str], token: str)
...
-                    break
+    def get_business_name_from_alternative_name(business_data, default_name, business_identifier):
+        """Get the business name from the alternative name list, if it exists."""
+        business_info_dict = business_data.get("business") if "business" in business_data else business_data
 
-        return business_name
+        if alternative_names := business_info_dict.get("alternateNames"):
+            for alt_name in alternative_names:
+                if alt_name.get("identifier") == business_identifier:
+                    return alt_name.get("name")
```
seeker25, line 369: assumes name is always populated (probably fine)

seeker25 (approved):

PR #3419, opened 2025-05-29
27660 - Allow account rejection when affidavit is INACTIVE or missing

hanlunBCRegistries:
*Issue #:*
https://github.com/bcgov/entity/issues/27660

*Description of changes:*
This PR fixes a bug where staff members were unable to reject accounts in certain scenarios related to affidavit status.
Modified the `approve_or_reject` method in `auth_api/services/affidavit.py` to:
- When **rejecting**: Only filter out REJECTED affidavits, allowing INACTIVE ones to be processed
- When **approving**: Continue to filter out both INACTIVE and REJECTED affidavits (existing behavior)
- Handle cases where no affidavit is found by logging a warning and allowing the operation to continue

### DIFF
--- a/auth-api/tests/unit/services/test_affidavit.py
+++ b/auth-api/tests/unit/services/test_affidavit.py
@@ -178,23 +178,23 @@ def test_reject_org_with_inactive_affidavit(session, keycloak_mock, monkeypatch,
     affidavit = AffidavitService.create_affidavit(affidavit_info=affidavit_info)
     org = OrgService.create_org(TestOrgInfo.org_with_mailing_address(), user_id=user.id)
     org_dict = org.as_dict()
-    
+
     # Find the task and set it to HOLD (simulating staff action)
     task_model = TaskModel.find_by_task_for_account(org_dict["id"], status=TaskStatus.OPEN.value)
     task_model.status = TaskStatus.HOLD.value
     task_model.save()
-    
+
     # Create another affidavit which will make the first one INACTIVE
     new_affidavit_info = TestAffidavit.get_test_affidavit_with_contact()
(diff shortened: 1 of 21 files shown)


auth-api/tests/unit/services/test_affidavit.py
```diff
@@ -158,3 +158,93 @@ def test_reject_org(session, keycloak_mock, monkeypatch, gcs_mock):
 
     # Verify GCS mock was called
     assert gcs_mock["mock_blob"].generate_signed_url.called
+
+
+@mock.patch("auth_api.services.affiliation_invitation.RestService.get_service_account_token", mock_token)
+def test_reject_org_with_inactive_affidavit(session, keycloak_mock, monkeypatch, gcs_mock):
+    """Assert that an org can be rejected when affidavit is INACTIVE (Scenario 1)."""
+    # Setup mock for GCS
```
seeker25, line 166: ask prompt to put less comments, this is double the work for refactoring please

seeker25 (changes requested):
Don't need a comment explaining every line - it's going to be headache for refactoring

seeker25:
The affidavit process should work off of org_id, right now it's based entirely off of the keycloak user guid

This would be a bunch of rework, but it's why these rejections are happening here's an example scenario

Org 1 is created, rejected.. which rejects the affidavit for the user 
Org 2 is created since it's already rejected it causes a 500 error

auth-api/tests/unit/services/test_affidavit.py
hanlunBCRegistries: updating now, was trying to follow the comments convention of [test_reject_org](https://github.com/bcgov/sbc-auth/blob/main/auth-api/tests/unit/services/test_affidavit.py#L112)

PR #3424, opened 2025-06-03
Fix security deps

seeker25:


queue_services/account-mailer/pyproject.toml
```diff
@@ -29,7 +29,6 @@ semver = "3.0.2"
 six = "1.16.0"
 threadloop = "1.0.2"
 thrift = "0.20.0"
-tornado = "6.4.2"
```
seeker25, line 32: these should come from auth-api

seeker25:
Removing build-deps, otherwise we need to update build-deps, update auth-api then update all services

it's easier to just update auth-api and then update all services

PR #3427, opened 2025-06-06
Business Dashboard fixes

seeker25:
*Issue #:*

*Description of changes:*

### DIFF
--- a/auth-api/src/auth_api/services/affiliation.py
+++ b/auth-api/src/auth_api/services/affiliation.py
@@ -549,24 +549,28 @@ class Affiliation:
         return business
 
     @staticmethod
+    def _process_nr_for_business(business, name_requests, drafts):
+        """Process NR for a business entity."""
+        nr_num = business["nrNumber"]
+        if nr_num in name_requests:
+            business["nameRequest"] = name_requests[nr_num]["nameRequest"]
+            business = Affiliation._update_draft_type_for_amalgamation_nr(business)
+            if business["nameRequest"]["stateCd"] == NRStatus.CONSUMED.value:
+                drafts.remove(business)
+            del name_requests[nr_num]
+            return True
+        return False
+
+    @staticmethod
     def _combine_nrs(name_requests, businesses, drafts, remove_stale_drafts=True):
         """Combine NRs with the business and draft entities."""
         for business in drafts + businesses:
-            # Only drafts have nrNumber coming back from legal-api.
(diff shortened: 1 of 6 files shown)


auth-api/src/auth_api/services/affiliation.py
```diff
@@ -558,15 +561,16 @@ def _combine_nrs(name_requests, businesses, drafts):
                     del name_requests[nr_num]
                 else:
                     # If not in name_requests then it's a stale draft.
-                    drafts.remove(business)
+                    if remove_stale_drafts:
+                        drafts.remove(business)
```
seeker25, line 565: Needed to do this, otherwise our pagination counts would be off. 

We could filter in the entity_mapping DAO level, but I think it's best to leave it like this for now. 

That way we can test this easily on one account.. EG. Create two TEMPS, same NR... file 1.. the other shouldn't file or be removed by LEAR etc

seeker25:
Will fix code complexity Monday

auth-api/src/auth_api/services/entity_mapping.py
```diff
@@ -209,12 +219,27 @@ def paginate_from_affiliations(org_id: int, search_details: AffiliationSearchDet
...
+            t_identifiers = [f'"{base.identifier}"' for base in affiliation_bases if base.identifier.startswith("T")]
+            nr_identifiers = [f'"{base.identifier}"' for base in affiliation_bases if base.identifier.startswith("NR")]
+            other_identifiers = [
+                f'"{base.identifier}"'
+                for base in affiliation_bases
+                if not base.identifier.startswith("T") and not base.identifier.startswith("NR")
+            ]
+            print(f"T identifiers ({len(t_identifiers)}): {', '.join(t_identifiers)}")
+            print(f"NR identifiers ({len(nr_identifiers)}): {', '.join(nr_identifiers)}")
+            print(f"Other identifiers ({len(other_identifiers)}): {', '.join(other_identifiers)}")
```
seeker25, line 239: These make it real easy to copy and paste from VSCODE -> Postman

ochiu (approved):

Jxio (approved):

auth-api/src/auth_api/services/entity_mapping.py
seeker25: As discussed with Ody, we'll put in a simple logger for this one instead of using prints, but we need the logging.conf to come back

PR #3487, opened 2025-08-14
Revert "30049 - updated expiry in email template"

severinbeauvais:
Reverts bcgov/sbc-auth#3486.

Unfortunately, the expiry text isn't being displayed in the email. I figure it's better to revert the change so that the codebase is functional in case we need to quickly deploy a hotfix today, and I'll work on this change separately.

image

### DIFF
--- a/auth-api/pyproject.toml
+++ b/auth-api/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "auth-api"
-version = "3.0.7"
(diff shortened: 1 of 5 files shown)


bolyachevets (approved):
approving, in case you want to revert this today

seeker25:
reopen if needed

severinbeauvais:
Not needed. Sorry, I forgot to close this last week.

PR #3495, opened 2025-08-22
30309 - Changes to support payment method changes

seeker25:
*Issue #:*
https://github.com/bcgov/entity/issues/30309

*Description of changes:*


seeker25:
Unit test WIP

auth-api/src/auth_api/services/org.py
```diff
@@ -238,6 +272,10 @@ def _create_payment_settings(
                     raise_for_status=True,
                 )
             else:
+                current_payment_account = RestService.get(
+                    endpoint=f"{pay_url}/accounts/{org_model.id}", token=RestService.get_service_account_token()
```
seeker25, line 276: moving token so it grabs it once

PR #3548, opened 2025-12-02
27014 - emails fix for inherited roles from group

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/27014

*Description of changes:*
- fix route that returns emails by role where inherited roles from groups are not returned


seeker25 (approved):
LGTM

PR #3580, opened 2026-01-15
Add FIN to email list when an account got eft enabled

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/26991

*Description of changes:*


queue_services/account-mailer/src/account_mailer/resources/worker.py
```diff
@@ -118,12 +118,14 @@ def handle_eft_available_notification(message_type, email_msg):
     template_name = TemplateType.EFT_AVAILABLE_NOTIFICATION_TEMPLATE_NAME.value
     org_id = email_msg.get("accountId")
     admin_emails = get_member_emails(org_id, (ADMIN,))
+    fin_email = "BCROSRev@gov.bc.ca"
```
seeker25: this should come from a secret, so we can configure it

PR #3605, opened 2026-02-10
Small tweak for Account-Id or business identifier undefined

seeker25:


auth-api/src/auth_api/services/authorization.py
```diff
@@ -239,6 +239,7 @@ def check_auth(**kwargs):
     else:
         business_identifier = kwargs.get("business_identifier", None)
         org_identifier = kwargs.get("org_id", None) or user_from_context.account_id
+        auth = None
         if business_identifier:
             auth = Authorization.get_user_authorizations_for_entity(business_identifier)
```
seeker25: returns {}

auth-api/src/auth_api/services/authorization.py
seeker25: auth = None possibly

seeker25: since auth is None,  the check_for_roles always fails
role = None

PR #3752, opened 2026-04-17
32405 - SAF email template update, create affiliation confirmation email

deetz99:
*Issue #:*
https://github.com/bcgov/entity/issues/32405

*Description of changes:*
- update unaffiliated email template to final design
- send confirmation email on successful affiliation creation
- update affiliation invitation model to return expires_on at the end of the day if type == unaffiliated

### DIFF
--- a/auth-api/src/auth_api/services/affiliation.py
+++ b/auth-api/src/auth_api/services/affiliation.py
@@ -202,8 +202,8 @@ class Affiliation:
             email_address = contact_link.contact.email
 
         if email_address:
-            registry_home_url = current_app.config.get("REGISTRY_HOME_URL")
-            context_url = f"{registry_home_url}login"
+            bc_registry_home_url = current_app.config.get("BC_REGISTRY_HOME_URL")
+            context_url = f"{bc_registry_home_url}login"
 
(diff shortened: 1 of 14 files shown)


auth-api/src/auth_api/services/affiliation.py
```diff
@@ -194,6 +201,27 @@ def create_affiliation(
...
+
+            mailer_data = ConfirmationEmailData(
+                business_name=entity.name,
+                email_addresses=email_address,
+                business_identifier=entity.business_identifier,
+                context_url=context_url,
+                completion_date=affiliation.created
+            )
+            publish_to_mailer(
+                notification_type=QueueMessageType.AFFILIATION_INVITATION_CONFIRMATION_EMAIL.value,
```
ochiu: I am assuming this email is okay even when it wasn't kicked off by an affiliation invitation, such as staff creates the affiliation, but in that case it wouldn't be an affiliation invitation confirmation.

If that is okay I would suggest renaming this to AFFILIATION_CONFIRMATION_EMAIL instead.

ochiu (approved):

PR #3798, opened 2026-06-03
33386: Fix StaleDataError race condition in process_name_events

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/33386

## Description of changes: ##
Fixed race condition causing StaleDataError in process_name_events.

The worker was loading the Entity from the database at the top of process_name_events, then immediately setting fields on it, and only calling save() at the very end; after a potentially multi-second PAY API call. 
This left the entity object dirty in the SQLAlchemy session for the entire duration of that call. If a concurrent auth-api request deleted the entity row during that window, the final UPDATE would match zero rows and SQLAlchemy would raise StaleDataError.

### Fix: ###
- Moved the entity load to after the PAY API call, so it stays in the session for the shortest possible time before being written back
- Moved the activity log write to after nr_entity.save() since it is a audit event and has no role on the core entity/affiliation work

### DIFF
--- a/queue_services/auth-queue/src/auth_queue/resources/worker.py
+++ b/queue_services/auth-queue/src/auth_queue/resources/worker.py
@@ -205,7 +205,7 @@ def process_name_events(event_message: SimpleCloudEvent):
             org: OrgModel = db.session.query(OrgModel).filter(OrgModel.id == int(auth_account_id or -1)).one_or_none()
             # If account is present and is not a gov account, then affiliate.
             if org and org.access_type != AccessType.GOVM.value:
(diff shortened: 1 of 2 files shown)


ochiu (approved):
Looks good, thanks

Jxio (approved):
LGTM, can PR title update like "Fix StaleDataError race condition in process_name_events"?

sumesh85:
> LGTM, can PR title update like "Fix StaleDataError race condition in process_name_events"?

Done! Didn't notice the original PR title :)

PR #3807, opened 2026-06-06
33382 - Own PAD setup-in-progress user message in auth-api

Jxio:
*Issue #:*
https://github.com/bcgov/entity/issues/33382

*Description of changes:*

when the error is only surfaced through auth-api → auth-web (e.g. account deactivate: OUTSTANDING_CREDIT). Pay-api should return type only; auth maps the code, sets HTTP status, and supplies user-facing text.

### DIFF
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.10.57",
+  "version": "2.10.56",
   "appName": "Auth Web",
   "sbcName": "SBC Common Components",
   "private": true,
--- a/auth-web/package.json
+++ b/auth-web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.10.56",
+  "version": "2.10.57",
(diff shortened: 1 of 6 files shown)


bcregistry-sre:
Temporary Url for review: https://bcregistry-account-dev--pr-3807-h23cxna6.web.app

sumesh85 (approved):

auth-web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "auth-web",
-  "version": "2.10.56",
+  "version": "2.10.57",
```
sumesh85: As discussed a while back; can we use practice of updating second digit on first change after the prod release; then increase the t3rd digit for each change ?
May be we can start with 2.11.0 and 2.11.1, 2.11.2 etc. ?

ochiu (approved):

auth-web/package.json
Jxio: that's a good idea, I agree

PR #3823, opened 2026-06-25
33557: Linking key to populate vendor account id later stage

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/33557

*Description of changes:*
- Linking Code can have PENDING status and link to the vendor later from vendor system


auth-api/src/auth_api/resources/v1/org_linking_keys.py
```diff
@@ -25,11 +25,12 @@
...
 from auth_api.utils.endpoints_enums import EndpointEnum
-from auth_api.utils.roles import ADMIN, COORDINATOR, Role
+from auth_api.utils.roles import ADMIN, COORDINATOR, USER, Role
+from auth_api.utils.user_context import UserContext, user_context
 
-bp = Blueprint("LINKING_KEYS", __name__, url_prefix=f"{EndpointEnum.API_V1.value}/orgs/<int:org_id>/linking-keys")
+bp = Blueprint("LINKING_KEYS", __name__, url_prefix=EndpointEnum.API_V1.value)
 
-_OWNER_ROLES = (COORDINATOR, ADMIN)
+_OWNER_ROLES = (USER, COORDINATOR, ADMIN)
```
Jxio: I think (USER, COORDINATOR, ADMIN) can view, but only (COORDINATOR, ADMIN) can patch/delete

auth-api/src/auth_api/services/account_linking_key.py
```diff
@@ -103,3 +113,28 @@ def validate(key: str, vendor_account_id: int) -> AccountLinkingKeyModel | None:
...
         record.save()
         return record
+
+    @staticmethod
+    def bind(key: str, vendor_account_id: int) -> AccountLinkingKeyModel | None:
+        """Bind a PENDING key to the given vendor, activating it.
+
+        Returns the updated record, or None if the key is not found, expired, or not in PENDING state.
+        """
+        record = AccountLinkingKeyModel.find_pending_by_key(key)
```
Jxio, line 103: should we also check existing_active here?

auth-api/src/auth_api/resources/v1/org_linking_keys.py
sumesh85: Thanks, updated!

auth-api/src/auth_api/services/account_linking_key.py
sumesh85: Thanks, good catch! Updated the code to revoke any existing ACTIVE for the account-vendor combo

Jxio (approved):

PR #3836, opened 2026-07-20
Changes to skip task for GOVM/N

sumesh85:
*Issue #:*

*Description of changes:*

### DIFF
--- a/auth-api/migrations/versions/2026_07_20_c4b9e7a2f1d8_add_skip_gov_review_to_product_codes.py
+++ b/auth-api/migrations/versions/2026_07_20_c4b9e7a2f1d8_add_skip_gov_review_to_product_codes.py
@@ -1,0 +1,29 @@
+"""Add skip_gov_review flag to product_codes and enable it for BCA.
+
+Revision ID: c4b9e7a2f1d8
+Revises: dfe9614f7fb0
+Create Date: 2026-07-20 00:00:00.000000
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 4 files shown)


Jxio (approved):

auth-api/migrations/versions/2026_07_20_c4b9e7a2f1d8_add_skip_gov_review_to_product_codes.py
```diff
@@ -0,0 +1,29 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table('product_codes', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('skip_gov_review', sa.Boolean(), nullable=True))
+
+    op.execute("update product_codes set skip_gov_review=false where skip_gov_review is null")
+    op.execute("update product_codes set skip_gov_review=true where code='BCA'")
```
ochiu, line 24: Should we make the column mandatory after this data has been updated here?

ochiu (approved):

auth-api/migrations/versions/2026_07_20_c4b9e7a2f1d8_add_skip_gov_review_to_product_codes.py
sumesh85: Default is set as False, do you think we still need non null constraint ?

ochiu: It is fine as is, thanks

PR #3870, opened 2026-08-21
34270 - account linking redirect flow

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/34270

*Description of changes:*
- auth-api  updates to capture redirect url that was validated against at the time of key generation
- updates to UI components to match current UX styling
- clean up existing vendor linking dialogs based on new UX designs
- vendor linking redirect flow
- filled in some previously missing tests to help with regressions


auth-web/src/components/auth/common/AccountSelectList.vue
```diff
@@ -0,0 +1,81 @@
...
+    </template>
+  </v-card>
+</template>
+
+<script lang="ts">
+import { Component, Prop, Vue } from 'vue-property-decorator'
+import { OrgWithAddress } from '@/models/Organization'
+
+@Component({})
+export default class AccountSelectList extends Vue {
```
ochiu: Extracted this piece out so it can be re-used between account creation flow and vendor linking selection. Some styling updates were made to be more inline with general styling.

auth-web/src/components/auth/vendor-linking/VendorLinkingLoginCard.vue
```diff
@@ -0,0 +1,150 @@
...
+import { Pages } from '@/util/constants'
+import VendorLinkingWarningAlert from '@/components/auth/vendor-linking/VendorLinkingWarningAlert.vue'
+import { buildSigninPath } from '@/util/vendor-connection-util'
+
+@Component({
+  components: {
+    VendorLinkingWarningAlert
+  }
+})
+export default class VendorLinkingLoginCard extends Vue {
```
ochiu: Login component to meet the needs of vendor linking since the old component styling is dated. When we move to NUXT we may update the nuxt base layer component to support this instead.

auth-web/src/components/auth/vendor-linking/VendorLinkingWarningAlert.vue
```diff
@@ -0,0 +1,46 @@
...
+      <strong>Note:</strong> {{ note }}
+    </div>
+  </v-alert>
+</template>
+
+<script lang="ts">
+import { Component, Prop, Vue } from 'vue-property-decorator'
+
+@Component
+export default class VendorLinkingWarningAlert extends Vue {
```
ochiu: Consolidated other warning dialogs into this one for re-usability. UX has changed, some of the previously defined messaging have been moved out of warning dialog boxes and just plain text.

auth-web/src/routes/router.ts
```diff
@@ -86,6 +89,13 @@ function redirectToBusinessDashboard (to) {
   return false // Stop the navigation, as we're redirecting externally
 }
 
+function requireAccountLinkingEnabled (to, from, next) {
+  if (VendorConnectionUtil.isAccountLinkingDisabled()) {
+    return next({ path: '/notfound' })
```
ochiu, line 94: Feature flagging.

auth-web/src/views/auth/create-account/AccountCreationSuccessView.vue
```diff
@@ -3,7 +3,23 @@
     class="view-container"
     data-test="div-account-setup-success-container"
   >
-    <v-row justify="center">
+    <template v-if="redirectToUrl">
```
ochiu, line 6: Added in redirect mechanism to account creation success.
Don't show existing screen and just spinner for redirect to happen.

auth-api/src/auth_api/services/account_linking_key.py
```diff
@@ -34,13 +37,24 @@ class AccountLinkingKey:
...
-        In both cases any existing PENDING key for the account is revoked first (one PENDING at a time).
+        If vendor_account_id is provided the key is ACTIVE and immediately usable, and
+        redirect_url is then required and must match one of vendor_account_id's registered
+        redirect URLs. If both are omitted a PENDING key is created — the vendor must call
+        bind() to activate it. Raises BusinessException if only one of the two is provided,
+        or if redirect_url doesn't match — checked before any other side effect, so a
+        rejected attempt doesn't revoke an existing key.
+        In all cases any existing PENDING key for the account is revoked first (one PENDING at a time).
         """
+        if bool(vendor_account_id) != bool(redirect_url):
```
sumesh85, line 53: This would evaluate to True if both are non null; also would evaulate to True if both are null. So may be consider to add a comment for anyone looking into it ?

sumesh85 (approved):
lgtm! Didn't really look deep into UI code though!

auth-web/src/views/auth/VendorLinkingView.vue
```diff
@@ -0,0 +1,328 @@
...
+              <p class="mb-0 mt-1">
+                {{ failureBody }}
+              </p>
+            </div>
+            <v-btn
+              v-if="resultErrorCode !== REDIRECT_URL_INVALID_CODE"
+              large
+              color="primary"
+              data-test="vendor-linking-result-return"
+              @click="redirectNow"
```
Jxio: what's the url in failure path? just wondering is it might redirect from BCROS page with a button to an unknown url that not validated by our backend?

ochiu: Yes I was thinking about this too. I am going to talk to Ethan about this. I don't think we should offer any return path at all if any error occurs, because we don't know if it is valid.

auth-api/src/auth_api/services/account_linking_key.py
ochiu: yea I don't think the check is fully accurate for what I want to validate and the error message I added, will update the check to be more explicit for when a vendor id is present and a redirect url is not.

auth-web/src/components/auth/vendor-linking/VendorLinkingLoginCard.vue
Jxio: can we use defineComponent here instead of class-based style?

Jxio (approved):
LGTM!

PR #3872, opened 2026-08-27
34718 - extend account linking key expiry

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/34718

*Description of changes:*
- account linking key extension auth API support
- only allow extension if currently active and within expiry reminder window
- extend 1 year from the current expiry date
- publish to activity log for extension


auth-api/src/auth_api/resources/v1/org_linking_keys.py
```diff
@@ -87,6 +89,30 @@ def delete_linking_key(org_id, key_id):
     return {}, HTTPStatus.OK
 
 
+@bp.route("/orgs/<int:org_id>/linking-keys/<int:key_id>", methods=["PATCH"])
```
Jxio, line 92: url repeat?
it will be like: /api/v1/orgs/<org_id>/linking-keys/orgs/<org_id>/linking-keys/<key_id>
hmm not sure why unit test passed..

ochiu: Blueprint is the top level for version, it shouldn't be repeating. Should work the same as DELETE

sumesh85 (approved):

Jxio (approved):

PR #3877, opened 2026-08-27
34723: Adding scope to support pay-api changes to only update cfs account

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/34723

*Description of changes:*

### DIFF
--- a/auth-api/src/auth_api/services/org.py
+++ b/auth-api/src/auth_api/services/org.py
@@ -65,7 +65,7 @@ from auth_api.utils.enums import (
 )
 from auth_api.utils.roles import ADMIN, COORDINATOR, EXCLUDED_FIELDS, STAFF, VALID_STATUSES, Role  # noqa: I001
 from auth_api.utils.user_context import UserContext, user_context
-from auth_api.utils.util import camelback2snake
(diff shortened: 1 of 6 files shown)


auth-api/src/auth_api/services/org.py
```diff
@@ -368,6 +378,24 @@ def _build_payment_request(org_model: OrgModel, payment_info: dict, payment_meth
...
+        """Return the org's stored contact in auth's mailing_address shape, or None."""
+        if not org_model.contacts or org_model.contacts[0].contact is None:
+            return None
+        return snake2camelback(ContactSchema().dump(org_model.contacts[0].contact))
+
+    @staticmethod
+    def _to_pay_contact_shape(auth_contact: dict) -> dict:
+        """Translate auth's mailing_address keys to pay-api's contact_info keys."""
+        return {
+            "addressLine1": auth_contact.get("street"),
```
ochiu, line 392: do we need street additional as well? sometimes people put unit / suite number here

ochiu:
Some tests to confirm the scope is passed to pay-api and ones related to fetching the contact info logic would help here.

ochiu (approved):

PR #3878, opened 2026-08-31
Changes to support govm account creation using service account

sumesh85:
*Issue #:*
https://github.com/bcgov/entity/issues/34759

*Description of changes:*
- Service account can create a GOVM account with manage_account role
- By default BUSINESS_SEARCH is added, everything else is paid for the account.
- No changes to existing behaviour for staff created accounts


auth-api/src/auth_api/services/org.py
```diff
@@ -606,6 +610,27 @@ def _create_payment_for_org(mailing_address, org, payment_info, is_new_org: bool
             payment_method = validator_obj.info.get("payment_type")
         return Org._create_payment_settings(org, payment_info, payment_method, mailing_address, is_new_org, scope=scope)
 
+    @staticmethod
+    def _needs_staff_review(access_type: str, user_id: int) -> bool:
```
sumesh85, line 614: Moved to helper methods to reduce complexity

auth-api/src/auth_api/services/pay.py
```diff
@@ -0,0 +1,92 @@
+# Copyright © 2026 Province of British Columbia
```
sumesh85, line 1: Created this pay service layer to move all pay-api integrations to here, not everythingn is moved yet though

ochiu (approved):

Jxio (approved):

PR #3893, opened 2026-09-04
34061 - revoke access update / expiry handling update

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/34816
https://github.com/bcgov/entity/issues/34756

*Description of changes:*
- Allow coordinator to revoke key
- Key extension buffer
- UI restructure for button states

### DIFF
--- a/auth-web/src/util/vendor-connection-util.ts
+++ b/auth-web/src/util/vendor-connection-util.ts
@@ -1,5 +1,10 @@
 
-import { AccountLinkingKey, VendorConnection, VendorConnectionStatuses } from '@/models/vendorConnection'
+import {
+  AccountLinkingKey,
+  VendorConnection,
+  VendorConnectionActions,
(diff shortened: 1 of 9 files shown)


Jxio (approved):

auth-web/src/util/vendor-connection-util.ts
```diff
@@ -29,6 +34,13 @@ function isOrgMember (membershipTypeCode: MembershipType | undefined): boolean {
   ].includes(membershipTypeCode)
 }
 
+function isAdminOrCoordinator (membershipTypeCode: MembershipType | undefined): boolean {
```
sumesh85, line 37: We used the permissions against each role from the auth-api backend, so that we don't need to check individual roles for any actions. Over the time, may be that has become obsolete, but something to keep in mind.

Image

PR #3902, opened 2026-09-14
34061 - date display updates

ochiu:
*Issue #:*
https://github.com/bcgov/entity/issues/34061

- update to latest date format standards for notifications
- update pytz version to handle permanent PDT

*Description of changes:*


auth-api/src/auth_api/services/account_linking_key.py
```diff
@@ -75,8 +75,8 @@ def generate(
         AccountLinkingKey._publish(ActivityAction.LINKING_KEY_GENERATED.value, record)
         if record.status == LinkingKeyStatus.ACTIVE.value:
             data = {
-                "linkDate": pacific_today_isoformat(),
-                "expiryDate": utc_to_pacific_isoformat(record.expires_on),
+                "linkDate": utc_to_pacific_display_date(datetime.now(UTC)),
```
sumesh85, line 78: Do we normally do this conversion in backend ? I thought it would be better to always return UTC from backend and let the display be controlled by the web.

sumesh85 (approved):

